PR #5, opened 2018-01-20
Dockerfile: switch from logutil-service to standard s6-log

jprjr:
Log-util service adds its own timestamps to logfiles - turtlecoind already has timestamps, so I changed from using logutil-service to the regular "s6-log", making for easier-to-ready logfiles.


RocksteadyTC:
I wonder how easy it'd be for our pools or other uptime-critical daemons to use a dockerized turtle-daemon with the help of something like a replication controller.

PR #9, opened 2018-01-23
Change checkpoint log message colour to yellow

ZedPea:
Fixes some users being confused by the green colour and thinking syncing is completed


RocksteadyTC:
Thank you!

PR #13, opened 2018-01-25
Rebase the 187k commit to include all new commits since v0.2.2 release

nnamon:
Time travels the 187k commit back to the beginning of v0.2.2 and then applies every other commit that has happened on top of it.


RocksteadyTC:
Hahah I still am amazed at your Git-fu

PR #23, opened 2018-02-02
Adding multi_installer.sh for multiplatform installations

michelangelo314:
The current easy_installer.sh only supports ubuntu 16+. This PR adds a multi_platform installer with support for OS X and written in an extendible format which can include other OS/distros and be run in a docker container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,40 +2,21 @@
 
 ### How To Compile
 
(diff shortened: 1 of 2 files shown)


RocksteadyTC:
Nice work :)

PR #42, opened 2018-02-06
Output private keys during new wallet creation

ereptor81:
Output private keys on wallet creation and instruct user to save them. Uses stdout so that it's not logged to file.

key_output_on_creation


RocksteadyTC:
I'm not sure if this works, all I see is `hunter2`

PR #43, opened 2018-02-06
ensure color codes show on other non-ansi platforms

ereptor81:
Because sometimes I forget Windows exists. Tested on Windows by Serroft.


RocksteadyTC:
Welcome into this world, commit f6cf18d

PR #52, opened 2018-02-09
only log levels 0-4 currently exist

blakelapierre:


RocksteadyTC:
Thank you, @blakelapierre :) Good catch

Issue #44, opened 2018-02-06
[BlockchainSynchronizer] Failed to query blocks: NodeErrorCategory:5, Internal node error


khunpoum:
ver relese v 0.3.1

after some time run simplewallet iam receiving errors but TurtleCoind is working.
and mee some time need to restart TurtleCoind and simplewallet.

how it can be fix?

simplewallet stdout:
```
2018-Feb-07 00:02:18.535571 ERROR   [BlockchainSynchronizer] Failed to query blocks: NodeErrorCategory:5, Internal node error
2018-Feb-07 00:02:33.510404 ERROR   [BlockchainSynchronizer] Failed to query blocks: NodeErrorCategory:5, Internal node error
2018-Feb-07 00:02:43.519390 ERROR   [BlockchainSynchronizer] Failed to query blocks: NodeErrorCategory:5, Internal node error
2018-Feb-07 00:02:58.523518 ERROR   [BlockchainSynchronizer] Failed to query blocks: NodeErrorCategory:5, Internal node error
```

```
$ uname -a
Linux trtl-pool-1 4.4.0-87-generic #110-Ubuntu SMP Tue Jul 18 12:55:35 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

$ uname -r
4.4.0-87-generic

$ cat /etc/*release*
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.3 LTS"
NAME="Ubuntu"
VERSION="16.04.3 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.3 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```

llawwehttam:
I believe this is a duplicate of:  https://github.com/turtlecoin/meta/issues/35

This is probably the right place for this issue now.

RocksteadyTC:
This needs to be solved. High priority. 2MM bounty for a fix in 48hrs.

MoonMoonDogo:
[...]

PR #53, opened 2018-02-09
Fix internal node error

deertacos:
The issue is a bit tough to reproduce, but locally I would usually see the error after running the wallet for an hour to an hour and a half. This code has been running on my machine for two and a half hours with no issue. I don't know that I can consider that a proper test, but I figured I should at least get the code out there so that someone who may have been having the issue more consistently can test my fix.

At any rate, I think I made the logic in the function more closely match what the comments said the function should be doing.

I also left in the logging that @MoonMoonDogo had suggested since there was a TODO for it anyway.

#44

### DIFF
--- a/src/CryptoNoteCore/BlockchainCache.cpp
+++ b/src/CryptoNoteCore/BlockchainCache.cpp
@@ -541,8 +541,12 @@ uint32_t BlockchainCache::getTimestampLowerBoundBlockIndex(uint64_t timestamp) c
     return 0;
   }
 
-  uint32_t blockIndex = parent->getTimestampLowerBoundBlockIndex(timestamp);
-  return blockIndex == INVALID_BLOCK_INDEX ? blockIndex : startIndex;
+  try {
+    return parent->getTimestampLowerBoundBlockIndex(timestamp);
+  } catch (std::runtime_error&) {
+    // parent didn't have the block, so index.front() must be the block we're looking for
+    return startIndex;
+  }
 }
 
 bool BlockchainCache::getTransactionGlobalIndexes(const Crypto::Hash& transactionHash,
--- a/src/CryptoNoteCore/Core.cpp
+++ b/src/CryptoNoteCore/Core.cpp
@@ -450,8 +450,8 @@ bool Core::queryBlocksLite(const std::vector<Crypto::Hash>& knownBlockHashes, ui
     fillQueryBlockShortInfo(fullOffset, currentIndex, BLOCKS_SYNCHRONIZING_DEFAULT_COUNT, entries);
 
     return true;
-  } catch (std::exception&) {
-    // TODO log
+  } catch (std::exception& e) {
+	logger(Logging::ERROR) << "Failed to query blocks: " << e.what();
     return false;
   }
 }


deertacos:
After running the wallet for over 3 hours, I finally saw the case which previously caused errors happen again. (I set a breakpoint to try and catch it.) After that, the wallet continued without errors. I imagine we still want other people to try it, but I am finally confident in saying that I fixed the error.

RocksteadyTC:
Thank you @deertacos for your quick contribution. I'd like to verify this code on a high volume daemon so I'll be pinging @CaptainMeatloaf for assistance if possible, and a few others just in case.

editfmah:
So, having reviewed this PR I see that recursively there is a try/catch around the subsequent calls to `getTimestampLowerBoundBlockIndex`.  But the return value, in the event that an error is not thrown will remain the same.

So is the issue that, when an error is thrown, that the wallet is sent something unpalatable and then the requests the wallet makes going forward are all invalid?

Obviously, its always good to catch errors and deal with them, but i'm struggling to understand where the difference in return value would be generated.  Because if it is sending something bad back to the wallet, we could also stop/deal with that at the entry point into the RPC service. 

I see no reason why this PR would not be included as good practice, but unless I am missing something, I can't see how it gets to the bottom of the root cause of the issue.

As a fix, I think I would be more in favour of creating a v2 API and doing what other coins do and lookup the blocks by their index and not by a tidemark timestamp which is always going to be fraught with difficulty.  Then just exclude the very top of the chain until it has slightly matured, say 60s.

RocksteadyTC:
@deertacos please post your wallet address to claim the bounty.

deertacos:
Here's my address:
TRTLuyp4NzB9PKdHiMJWgKeDpmpdgm9XHCJD1pfwNDkX6jA2cdtJxVcEQq7BFFsGuE5v191sLqmB92AnFyBp5TNYavGgRze5WUK

Thank you, and I'm glad I could help.

RocksteadyTC:
[wallet TRTLux]: transfer 5 TRTLuyp4NzB9PKdHiMJWgKeDpmpdgm9XHCJD1pfwNDkX6jA2cdtJxVcEQq7BFFsGuE5v191sLqmB92AnFyBp5TNYavGgRze5WUK 500000 -f 50            
2018-Feb-11 19:39:00.870201 INFO    Money successfully sent, transaction 6c49553c80e7530edf16dcfc5b8314f63349ae66a7d09bb2801c98038c1cfecf
[wallet TRTLux]: transfer 5 TRTLuyp4NzB9PKdHiMJWgKeDpmpdgm9XHCJD1pfwNDkX6jA2cdtJxVcEQq7BFFsGuE5v191sLqmB92AnFyBp5TNYavGgRze5WUK 500000 -f 50
2018-Feb-11 19:39:32.435163 INFO    Money successfully sent, transaction a34ee09d723e2abce5cfd00f6101e1f99c9e496b4cfcc80f2638f88f1cd04ce3
[wallet TRTLux]: transfer 5 TRTLuyp4NzB9PKdHiMJWgKeDpmpdgm9XHCJD1pfwNDkX6jA2cdtJxVcEQq7BFFsGuE5v191sLqmB92AnFyBp5TNYavGgRze5WUK 500000 -f 50
2018-Feb-11 19:39:50.990720 INFO    Money successfully sent, transaction ec3764ffa21147d50fa82963b64ed67a02474ceb9c7d551556f3302a43db5783
[wallet TRTLux]: transfer 5 TRTLuyp4NzB9PKdHiMJWgKeDpmpdgm9XHCJD1pfwNDkX6jA2cdtJxVcEQq7BFFsGuE5v191sLqmB92AnFyBp5TNYavGgRze5WUK 500000 -f 50
2018-Feb-11 19:40:13.997783 INFO    Money successfully sent, transaction aaa911f0a02d56e08e0d12a10828456c830a3d34b7911fa38b926a5b4d67a635
[wallet TRTLux]:

RocksteadyTC:
Thank you for your participation @deertacos

PR #55, opened 2018-02-11
Add view_tx_outputs command

ZedPea:
Lets you view the outputs of a specific transaction via it's hash, and will print all the outputs that are yours (if any). Useful for ensuring that a specific transaction actually did hit your wallet even if the wallet is playing up.

And here's a picture!

image


RocksteadyTC:
Good work Zed :)

PR #61, opened 2018-02-13
Added script to build releases more easily

wreiner:
Talked with @RocksteadyTC and @nnamon about a script to create releases in a more orderly fashion - this would be my proposal for such a script.


wreiner:
There is an "exit 0" for testing purposes - this should be removed if you decide to merge.

PR #64, opened 2018-02-14
Add password confirmation on new wallets

ereptor81:
Added password confirmation prompt in simplewallet when creating a new wallet. Would appreciate someone else giving these a once over w/ all of the different modes of creating/importing wallets, especially on Windows. 

wallet_changes


ereptor81:
Ah yeah didn't considering the pasting, that is kind of annoying. I'll revert that piece.

PR #70, opened 2018-02-16
Update README.md

mlametta:
Before you are able to run cmake from your turtlecoin folder you need to set the path variable.


ZedPea:
👍 

Also, might be an idea to change the default path of boost from `-DBOOST_ROOT=D:/Boost/boost_1_59_0` to `-DBOOST_ROOT=C:/local/boost_1_59_0` as I think that's the default install location (was on mine)

mlametta:
That was my default install location as well.

RocksteadyTC:
Thanks @mlametta

PR #73, opened 2018-02-16
Add outgoing transfers command

ZedPea:
We have an incoming_transfers command, and a list_transfers command which shows all transfers, why not an outgoing transfers?

Just had to copy the incoming transfers func and change like one line.


ZedPea:
The two imports? Those were what were breaking the build on windows, and were previously added erronously by me when I was messing around with the mnemonics code. It built fine on linux so I assumed I hadn't done anything wrong until then. It's been in all my commits atm because I pushed them to master so I didn't break master again depending upon the order of the PR's getting closed.

That fix already came in from the remain open on error branch, so I assume it's just a no-op in this case being applied again.

PR #74, opened 2018-02-19
added 0 to end of MONEY_SUPPLY value

blakelapierre:
untested

### DIFF
--- a/src/CryptoNoteConfig.h
+++ b/src/CryptoNoteConfig.h
@@ -34,7 +34,7 @@ const uint64_t CRYPTONOTE_BLOCK_FUTURE_TIME_LIMIT            = 60 * 60 * 2;
 const size_t   BLOCKCHAIN_TIMESTAMP_CHECK_WINDOW             = 60;
 
 // MONEY_SUPPLY - total number coins to be generated
-const uint64_t MONEY_SUPPLY                                  = UINT64_C(100000000000000);
+const uint64_t MONEY_SUPPLY                                  = UINT64_C(1000000000000000);
 const uint32_t ZAWY_DIFFICULTY_BLOCK_INDEX                 = 187000;
 const size_t ZAWY_DIFFICULTY_V2                          = 0;
 const uint8_t ZAWY_DIFFICULTY_DIFFICULTY_BLOCK_VERSION   = 3;


nnamon:
Why do we need to increase the max supply?

blakelapierre:
(using approximated numbers)

10,000,000,000 people

29,600 reward * 2 blocks/min * 60 minutes * 24 hours = 85,248,000 turtles/day

85,248,000 / 10,000,000,000 = 0.0085248 turtles/person/day

1,000 turtles/person / 0.0085248 turtles/person/day = 117,304 days for everyone to have 1,000 turtles = 321 years

Money supply of 100,000,000,000,000 shells only allows for 10,000 shells/person once fully produced.

blakelapierre:
At best, we are going to have about 10,000 shells per person on the planet (using rounded/approximated number). Is that really going to be that functional? It seems like more shells are needed.

What do you think about generating 100k shells (1000 turtles) per person per day?

blakelapierre:
Is my math wrong? Or you can only communicate through github emojis?

nnamon:
I don't think the emission schedule should be changed. Users have agreed on the initial rules by their participation and making this big of a centralised rule change while we are underway will invalidate their initial reasons for joining. 

Max supply should not be modified on the main chain.

blakelapierre:
I think that's a somewhat fair way of looking at it, but it seems like the low money supply will forever limit turtlecoin's usefulness.

What would be the damage to users if supply was increased???

blakelapierre:
Compare to Bitcoin:

16,875,100 coins with 100,000,000 satoshi per coin = 1,687,510,000,000,000 satoshis

168,751 satoshis per person (using 10,000,000,000 people)

Turtle has:

602,181,453,437 turtles with 100 shells per turtle = 60,218,145,343,700 shells

60 shells per person (using 10,000,000,000 people)

siftycat:
Bitcoin has a total supply of 21m coins, so 21,000tn satoshi.

You haven't said why you believe everyone needs 10,000 shells or what purpose this would serve. If it's egalitarian idealism then because TRTL is a freely traded coin and it costs money to mine, and if its prime purpose is to have transactional value and be a mechanism of exchange, the 10,000 shells you believe they need will not be equally distributed among persons. 

The reason there should be 10,000 TRTL for every person alive at the point where there are 10bn people on the planet and not more or less is arbitrary. It's already been explained [why TRTL has a total supply of 1tn with two decimal places](https://medium.com/@turtlecoin/one-trillion-turtles-coin-supply-and-unit-economics-5bfbea0aa1f1). The fact that supply of cryptocurrency requires work is one of its saving graces - while central banks can just "print money" in order to alleviate deeper economic issues, creating more money without any kind of work like it takes to mine natural resources (or in the case of crypto, solve mathematical problems), cryptocurrency retains value because of its strictly controlled and predictable supply and transfer mechanism. What you propose would turn TurtleCoin into the crypto equivalent of the Zimbabwean dollar, because then nothing would stop two or three more zeroes being added to the end of the money supply. Existing supplies would become increasingly worthless, people would stop mining, the blockchain would suffer from a lack of maintenance because people couldn't rely on the basic supply and algorithm - in short it would be a complete mess.

Most importantly, however, you've just added a zero without considering any of the implications of this PR. You haven't dealt with how any of the other software is going to deal with this - I assume you haven't looked at any of the other repos to see how they deal with this right now, and I assume you haven't considered the impact on coins already mined or the existing blockchain.

blakelapierre:
>You haven't said why you believe everyone needs 10,000 shells or what purpose this would serve.

wtf, I never said I believe everyone needs 10,000 shells

siftycat:
> wtf, I never said I believe everyone needs 10,000 shells

No, you're right. You said 100,000. I apologise for mixing one essentially arbitrary target spawned from a half baked idea and a second essentially arbitrary target spawned from a half baked idea.

Should I edit my original post to add a zero so it's clearer for you?

siftycat:
> nah brah, you got a copy-paste function?

Yeah I do, "brah", so here's you:

> What do you think about generating 100k shells (1000 turtles) per person per day?

This gets more redundant by the second. 100,000 shells per person per day? Is this for the 10bn people that are going to exist on this planet at some point? Or just everyone who's mining gets 100,000 shells every day? Is this a cryptonote supply algorithm or are we just handing them out? Because then we could just premine 1tn coins with 1000 units and hand them out. Cut this distributed consensus thing right out, because it gets in the way of whatever you think is happening in the world. You can't even explain why, what or how - you just farted out some half baked sums and picked a number out of the air.

Right "brah"? Tell you what. Go away and deal with the economic, social, psychological and, most importantly, engineering implications of adding a zero, then come back when you've solved those. Then you can add as many zeros as you like. "Brah".

_Don't even bother_ explaining what you mean, because that's a giant gaping rabbit hole I don't care to go down and I can't see it having any possible benefit for anyone. Not you, and _definitely_ not me.

CaptainMeatloaf:
This is getting out of hand, and frankly I believe @siftycat said everything that needed to be said. Continue this discussion elsewhere if you wish to have it.

PR #80, opened 2018-02-25
CryptoNoteCore: correctly calculate maximum transaction input size, c…

AnythingTechPro:
…heck transaction input size before placing it into the tx mempool

### DIFF
--- a/src/CryptoNoteConfig.h
+++ b/src/CryptoNoteConfig.h
@@ -52,6 +52,8 @@ const size_t   CRYPTONOTE_DISPLAY_DECIMAL_POINT              = 2;
 const uint64_t MINIMUM_FEE                                   = UINT64_C(10);
 const uint64_t DEFAULT_DUST_THRESHOLD                        = UINT64_C(10);
 
+const uint64_t MAX_TRANSACTION_SIZE_LIMIT                    = CRYPTONOTE_BLOCK_GRANTED_FULL_REWARD_ZONE_CURRENT / 4 - CRYPTONOTE_COINBASE_BLOB_RESERVED_SIZE;
+
 const uint64_t DIFFICULTY_TARGET                             = 30; // seconds
 const uint64_t EXPECTED_NUMBER_OF_BLOCKS_PER_DAY             = 24 * 60 * 60 / DIFFICULTY_TARGET;
 const size_t   DIFFICULTY_WINDOW                                = 17;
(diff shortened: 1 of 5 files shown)


ZedPea:
Doesn't this give a max transaction size of 24400 bytes?

```CRYPTONOTE_BLOCK_GRANTED_FULL_REWARD_ZONE_CURRENT / 4 - CRYPTONOTE_COINBASE_BLOB_RESERVED_SIZE;``` = ```100000 / 4 - 600 = 24400```

Yet we see transactions far larger than that go through, in fact the current max transaction size seems to be around 115k bytes. I'm not sure exactly how the max transaction size is meant to work, can you elaborate?

AnythingTechPro:
Yep your right, my bad; typo 👍

AnythingTechPro:
Adding to this, fusion transaction size should always be below (or in this case half) of the maximum allowed tx size. This is due to the fact that fusion transactions do not require fees, and they can be exploited to place transactions with higher size than maximum allowed in the tx memory pool.

ZedPea:
Maybe I'm missing something - I don't know how fusion transactions work so I'm sure you're correct on them having to be half the maximum - but this sets the maximum transaction size to ```100000 / 6 - 600 = 16066```
Yet we can see much larger transactions go through - for example here's a transaction with a size of 87874 that's happily been included in a block - https://blocks.turtle.link/?hash=396dc2191a92f1cc7dad352aff365e5c665c99549796a0e244a4393f74dbb47e#blockchain_block

Is there some byte/bits thing going on here, or am I fundamentally misunderstanding what this PR is doing?

AnythingTechPro:
Well, when sending transactions if the mixin size is lower (for some cases usually zero) and the amount of coin to be sent is about 1 million, this will result in a high transaction size (probably about 70k). But increasing mixin (ring signature) size to 3 or 4 should result in an estimated tx size of about 1k.

fexra:
@AnythingTechPro Could you elaborate why you don't have to pay any fees when using fusion transactions? I'm looking into incorporating them into an application I'm building, but don't fully grasp the concept. From what I understand you first have to estimateFusion(), before you can can use sendFusionTransaction() if you want to make sure you don't run into any errors. However, I'm not fully sure how you calculate or set the threshold. I'm sorry if this is a tad offtopic, but since you two are discussion fusions, it would be nice to here your knowledge.

PR #84, opened 2018-03-03
Add mnemonics to walletd

ZedPea:
This patch adds two things to walletd. 
Firstly, it allows you to generate a wallet from a mnemonic seed, with the `--mnemonic-seed` argument.

For example:

```
./walletd -w test.wallet -p pass -g --mnemonic-seed "beyond issued pause grunt skydive chrome mouth bays yearbook faulty voucher vessel ticket hickory behind subtly noises daily bluntly dilute return venomous madness niche noises"
```

Note that the argument has to be quoted to allow the shell to interpret it as one string.

Secondly, this patch adds an RPC method getMnemonicSeed(), which functions nearly the same as getSpendKeys(). It returns the mnemonic seed for the given address. For example:

```
curl -X POST -i -H "Accept:application/json" -d '{"params":{"address":"TRTLv2oMkQNQN4pR8JAiQr1Hdhiu4AeWicC7pg8AezbTUfeJ1HESc3n6umum2mn6U13BDpCBQsNms27svP7wKFhc48gef1gLhaG"},"jsonrpc":"2.0","id":"test","password":"test","method":"getMnemonicSeed"}' http://localhost:8070/json_rpc
```

This returns:

```
HTTP/1.1 200 OK
Content-Length: 227
Server: CryptoNote-based HTTP server

{"id":"test","jsonrpc":"2.0","result":{"mnemonicSeed":"moisture pencil usual algebra nifty evolved altitude awning sowed iguana megabyte fainted edgy solved army irony fuel upcoming border alley dwelt richly saga skater edgy"}}
```

To accomplish this, the default address is created with deterministic private keys, similarly to how the simplewallet patch worked. However, because walletd supports multiple addresses in one container, not all wallets can have a mnemonic seed, and so the RPC request will return an error notifying the user the their private keys aren't deterministic.

I'm not 100% sure how the createAddress() rpc method works with the different parameters, from my tests I haven't messed anything up but someone else checking would be good.

Finally, if this PR is merged, we will diverge from the bytecoin docs in two ways - we have a required password parameter, and we will have a new method. We should probably decide on a way
to host our own RPC documentation somewhere.


fexra:
Looking good! I'm excited about this as I been working constantly with the walletd RPC the last 5 weeks. I wouldn't mind taking the documentation as my duty.

RocksteadyTC:
@BebopTC I can merge this once we can comply with the latter requirement from @ZedPea about the rpc documentation. Who would you like to delegate the forking of RPC documentation to?

> Finally, if this PR is merged, we will diverge from the bytecoin docs in two ways - we have a required password parameter, and we will have a new method. We should probably decide on a way
> to host our own RPC documentation somewhere.

RocksteadyTC:
No response from @BebopTC, I'll be delegating the responsibility of the RPC documentation forking to a community member. I think someone has done it already, and would just need to ask.

RocksteadyTC:
Let's get this one in motion, sans RPC docs.

PR #91, opened 2018-03-11
Logging text changes

RocksteadyTC:
Give logging messages more human readable output


src/CryptoNoteCore/BlockchainCache.cpp
```diff
@@ -541,12 +541,8 @@ uint32_t BlockchainCache::getTimestampLowerBoundBlockIndex(uint64_t timestamp) c
     return 0;
   }
 
-  try {
-    return parent->getTimestampLowerBoundBlockIndex(timestamp);
-  } catch (std::runtime_error&) {
-    // parent didn't have the block, so index.front() must be the block we're looking for
-    return startIndex;
-  }
```
ZedPea: Are you sure you meant to push this change up? Might fix the syncing issue but also might break balances and reset.

Makefile
```diff
@@ -1,33 +1,734 @@
-all: all-release
```
ZedPea: Don't need these build artifacts. Does the gitignore include Makefile?

RocksteadyTC: We can add it

src/CryptoNoteCore/BlockchainCache.cpp
RocksteadyTC: True, let me know if it causes issue (it hasn't here yet) and I can roll it back. So far, I'm running this change as well as Frukstav, but I guess we're edge cases.

ZedPea: Hmm, ok. This is just a revert of the previous fix, right? Unsure why it would work now...

ereptor81: The first night I ran this it definitely hosed the sync/balance in my wallet. If it actually fixes the hanging issues in the daemon, we should look into the wallet and see why the exception is screwing up things on that side though. I'll pull this and start running it again and see what I can find.

ereptor81: Just got home and got the chance to test the latest from master. Did a full sync earlier sans bootstrap and created a new wallet from imported keys and it immediately screwed up the balance. It looks like when that exception gets thrown it interrupts the sync process and you just end up syncing from whatever point it sees as the startIndex at the time, so you just get a partial balance from that block onwards. the VM i have it on right now seems to be doing it every time, so i should be able to debug it now.

RocksteadyTC: This is great @ereptor81 have you seen anything since then?

PR #93, opened 2018-03-12
Simplewallet use walletgreen

ZedPea:
Yay, rewrite is done!

Features:
* Uses walletd file format, so you can swap between a GUI wallet and a simplewallet wallet at will
* Will open and convert legacy wallets just fine.
* Fusion transactions! This adds a command `quick_optimize` and `full_optimize` which launch fusion transactions to optimize the wallet, in preparation of sending large transactions.
* Large transaction detection (in bytes). If a transaction is > 115k bytes, we detect it, and first try to launch a single round of fusion transactions, then resend the transaction. If it's still too large, then we split it up into multiple transactions. This is implemented correctly to ensure that no single transaction is > 115k bytes. This should prevent any transactions sent from this wallet ever getting stuck.
* Much more friendly output - if something goes wrong, it tries to give you a user friendly message.
* Lots of colours - I think colours make the CLI a lot less scary, it lets you highlight the important bits.
* No annoying command lines, like `transfer 5 TRTL... 100`, just prompt the user for each single parameter, so they don't have to remember argument switches.
* Defaults and better explanations. Most users don't know what the heck mixin is, so why not tell them? It also gives them a default mixin, fee, etc, so they don't need to know what a good value is and can just hit enter.
* If we're doing an operation that can get stuck, try and detect it and let the user know.
* If we're doing an operation that takes a while, show regular updates so you know it's not stuck.
* Easy to read code! One of the main objectives was making a simple codebase, which is split into simple functions, and is easy to jump into. Obviously I'm biased, but I feel this code is pretty simple and  far easier to read than the previous simplewallet code.
* No unneeded classes. Simplewallet had a ton of class variables, when a class wasn't even needed. We have a few small structs and that's it. I like to know where my variables come from damn it!
* Default mixin of 5. Can set lower, would be cool if we made that the min though. Added this to the cryptonote config to be more configurable.
* View only wallets

Unrelated to simplewallet rewrite:

* It also changes the max transaction size in walletd to 115k, so this should prevent walletd tx's getting stuck as well.
* Removes the miner executable and connectivity tool. No code is removed, they are just removed from the cmakefiles to speed up builds. Most people won't need them, and it's simple to re-enable if you do.
* Slightly simplified src/CmakeLists, and added a few comments.

I don't think gives a great picture of the differences between the current simplewallet and my rewrite. I would you suggest you try it out, and hopefully the difference in usability will be obvious. Here's a screenshot.

Obviously simplewallet has existed for a while, and so there might have been some issues I have overlooked. Let me know if there's anything missing that you'd like, or a bug that needs fixing.


georgyous:
Hi ZedPea,

I had a open issue for a feature( that would be satisfied by your pull request) for forknote. Would you be willing to open a pull request for that project also?

Very nice work by the way

Thanks

mtl1979:
I think removing miner and connectivity_tool from build files should be separated as another pull request and rest of the changes squashed as one single commit. All the profanity in the commit messages makes the patch set look unprofessional which I'm sure was not the intent.

CaptainMeatloaf:
Agree with the commit squashing. Also, have you tested that this does not break compatibility with the pool software? The pools use simplewallet.

mtl1979:
Squashing the commits also make it easier for other people to test the code, only require one cherry-pick instead of full branch change/rebase... RPC compatibility with pool software should indeed be tested. Also read forknote/forknote#23... If everything works, maybe should be sent upstream so could be reused for all forknote coins.

hensleyrob:
Can you keep the ability to put all of the transfer parms in one command as well? I much prefer transfer 5 XXX 123.45 instead of being prompted for everything. Maybe only prompt if no arguments are given?

mtl1979:
@hensleyrob +1 for asking parameters only if not given after the command.

ZedPea:
Connectivity tool and miner have been re-added. I attempted to do a rebase/squash but it looks like I messed it up. If someone can explain how to do it for a brainlet that would be good.

This will break compatibility with the pool software. I had been talking with the other guys about the implications of this, and it seemed like in the future, the best way to move forward would be to move the pool software to work with walletd rather than the simplewallet RPC, as it supports the same functionality as simplewallet.

For now, I was going to suggest retaining the current simplewallet, and simply renaming it pool-wallet or something similar, whilst removing the cli interface. I'm not sure how easy it would be to make this patch support the pool workflow, but I wanted to drop the amounts of arguments used heavily.

I'll add the transfer cli command shortly.

I'm not sure exactly how integrating with forknote would work... don't they build their source by applying the forknote config to bytecoin? So to implement it wouldn't it need to be added as an extension?

mtl1979:
@ZedPea forknote is patched Bytecoin where all hard-coded coin-specific data is moved to config file, so you can use same executable with many CryptoNight coins. A lot of coins are actually based on Forknote code (There is only two choices: forknote or Monero), but the config file is integrated back to the source code as a header file.

Using walletd for pools might warrant separate pull request as exchanges usually also use walletd instead of simplewallet when available.

ZedPea:
@mtl1979 I'll happily make a PR to the main forknote repo then, last time I did one it was to the config generator so I wasn't sure what the difference was.

mtl1979:
Squashing commits using command-line git client only requires running:

```git rebase -i HEAD~N``` where N is number of commits to be squashed + 1

Change all commands (first word on each line) starting from line that contains second of your commits to ```fixup```.

When pushing changes to GitHub, you need to use ```git push --force```

**Comment by ZedPea** [comment]
@mtl1979 I'll give that a go, thanks. I didn't force push last time, maybe that was it. I also changed them to squash instead of fixup. There are a few of someone else's commits last time I rebased. Do I just leave them as they are?

**Comment by mtl1979** [comment]
@ZedPea On the list of commits, there should always be at least one commit made by someone else, because that commit is used as anchor point of the rebase, so git knows what commit will be "parent" of the squashed commit. You should always change only commits that you have made yourself. The number used for ```N``` means rebasing current commit, and ```N```-1 commits before current commit, on top of the anchor commit, which is ```N``` commits before latest commit.

**Comment by mtl1979** [comment]
I think we need to fix the pool code first and then apply this patch after that...

**Comment by ZedPea** [comment]
I'm not sure how hard it is to make the pool use walletd... I see this patch (quite outdated) https://github.com/turtlecoin/turtle-pool/commit/72f09886b3f879b0de3f13e3b1d58b4b2809bb05

For reference, the walletd api is here: https://wiki.bytecoin.org/wiki/Bytecoin_RPC_Wallet_API
And the simplewallet api is here: https://wiki.bytecoin.org/wiki/Simplewallet_JSON_RPC_API

Kinda hard to test without running a pool...

**Comment by CaptainMeatloaf** [comment]
If the RPC interfaces remain the same on simplewallet with this patch, it should remain the same.

You can easily set up a pool with the provided instructions in the turtle-pool repo on the testnet which can be found on rocks account.

**Comment by ZedPea** [comment]
@CaptainMeatloaf I don't run an RPC, I just talk to turtlecoind. I think it would be pretty easy to add compatibility with the current RPC, but I don't want to add all the overhead of all the commands that current simplewallet has. It makes the code a lot more ugly and complex when it should just be user facing IMO. It seems more sensible to transfer to walletd for a programmatic interface in the future, and sunset current simplewallet.

In regards to setting up a pool, from people I had seen in help/mining, I thought the instructions were a bit lacking. Maybe I'm wrong.

**Comment by CaptainMeatloaf** [comment]
In that case I would suggest holding off merging this request until a viable patch can be organised for the pool software, otherwise you end up with pools just running outdated software.

The instructions are perfectly understandable if you follow them exactly, and use the exact versions it specifies. Most of the issues people have are from not using the correct versions and just getting the latest from apt or whatever.

**Comment by ZedPea** [comment]
OK, I agree. I will try messing around with the pool at some point unless someone else wants to take it on.

**Comment by ZedPea** [comment]
There is a patch for pools available here: https://github.com/ZedPea/turtle-pool/commit/aee42ced4d6269e7eb52a155c4e81cf3bc9b2bbf

I have yet to try it out. I will attempt setting up a pool on testnet soon.

**Comment by RocksteadyTC** [comment]
@ZedPea I think we can maybe find a pool admin to run this and give it a shot. Maybe Cod or Tim or Cision.

**Comment by CaptainMeatloaf** [comment]
I wouldn't be prepared to run this on prod. I think ZedPea running it on testnet first is a good idea.

File: src/CMakeLists.txt [hunk 1]
```diff
@@ -12,29 +12,32 @@ file(GLOB_RECURSE Daemon Daemon/*)
...
-file(GLOB_RECURSE JsonRpcServer JsonRpcServer/*)
-
-file(GLOB_RECURSE PaymentGate PaymentGate/*)
-file(GLOB_RECURSE PaymentGateService PaymentGateService/*)
-file(GLOB_RECURSE Miner Miner/*)
-
-source_group("" FILES $${Common} ${ConnectivityTool} ${Crypto} ${CryptoNoteCore} ${CryptoNoteProtocol} ${Daemon} ${JsonRpcServer} ${Http} ${Logging} ${NodeRpcProxy} ${P2p} ${Rpc} ${Serialization} ${SimpleWallet} ${System} ${Transfers} ${Wallet} ${WalletLegacy})
+# This appears to be an IDE thing, to group files together. 
+# https://cmake.org/cmake/help/v3.0/command/source_group.html
+# Probably not what you need to be looking at if somethings not building
```
mtl1979: "something is not building" or "something isn't building".

src/SimpleWallet/SimpleWallet.cpp
```diff
@@ -1,752 +1,575 @@
...
-  } else {
-    addrEnd = url.find('/');
-    port = 80;
-  }
-
-  address = url.substr(addrStart, addrEnd - addrStart);
-  return true;
-
-}
+#include <SimpleWallet/SimpleWallet.h>
```
mtl1979: Removing copyright block might be a bad idea even if the file is completely rewritten... Not having a copyright block in a source file means it can't be merged back in.

src/SimpleWallet/SimpleWallet.cpp
```diff
@@ -757,990 +580,756 @@ bool simple_wallet::is_valid_mnemonic(std::string &mnemonic_phrase, Crypto::Secr
...
 
-  if (command_line::has_arg(vm, arg_log_level)) {
-    logLevel = static_cast<Level>(command_line::get_arg(vm, arg_log_level));
-  }
+    /* Now we rescan the chain to re-discover our balance and transactions */
+    findNewTransactions(node, walletInfo->wallet);
 
-  logManager.configure(buildLoggerConfiguration(logLevel, Common::ReplaceExtenstion(argv[0], ".log")));
+    /* Then we save the wallet again, with the hopefully fixed balance. This
+       step is not strictly neccessary, but it's nice to do in case the user
```
mtl1979: "necessary"

src/SimpleWallet/SimpleWallet.cpp
```diff
@@ -757,990 +580,756 @@ bool simple_wallet::is_valid_mnemonic(std::string &mnemonic_phrase, Crypto::Secr
...
+                if (t.totalAmount < 0)
+                {
+                    printOutgoingTransfer(t);
+                }
+                /* Don't display fusion transactions */
+                else if (t.totalAmount > 0)
+                {
+                    printIncomingTransfer(t);
+                }
+            }
```
mtl1979: This if block would be clearer if it would be inside the previous if block... Then the comment about fusion transactions wouldn't need to be duplicated.

src/SimpleWallet/SimpleWallet.cpp
```diff
@@ -757,990 +580,756 @@ bool simple_wallet::is_valid_mnemonic(std::string &mnemonic_phrase, Crypto::Secr
...
+
+                if (t.totalAmount < 0)
+                {
+                    printOutgoingTransfer(t);
+                }
+                /* Don't display fusion transactions */
+                else if (t.totalAmount > 0)
+                {
+                    printIncomingTransfer(t);
+                }
```
mtl1979: This if block would be clearer if it would be inside the previous if block... Then the comment about fusion transactions wouldn't need to be duplicated.

src/SimpleWallet/SimpleWallet.cpp
```diff
@@ -757,990 +580,756 @@ bool simple_wallet::is_valid_mnemonic(std::string &mnemonic_phrase, Crypto::Secr
...
+
+    return InformationMsg("[TRTL " + shortName + "]: ");
+}
+
+void connectingMsg()
+{
+    std::cout << std::endl << "Making initial contact with TurtleCoind."
+              << std::endl
+              << "Please note, this sometimes can take a long time."
+              << std::endl << "Please wait..." << std::endl
```
mtl1979: Instead of saying "Please note" on second line, just say "Please wait, this sometimes can take a long time..." as "note" literally means that user is expected to write something down...

src/SimpleWallet/SimpleWallet.cpp
```diff
@@ -757,990 +580,756 @@ bool simple_wallet::is_valid_mnemonic(std::string &mnemonic_phrase, Crypto::Secr
...
+              << std::endl;
 }
 
+void viewWalletMsg()
+{
+    std::cout << InformationMsg("Please note that when using a view wallet "
```
mtl1979: "Please remember" is better wording than "Please note"...

src/SimpleWallet/SimpleWallet.h
```diff
@@ -1,195 +1,117 @@
...
-#include "TransferCommand.h"
-
-#include "Common/ConsoleHandler.h"
-#include "CryptoNoteCore/CryptoNoteBasicImpl.h"
-#include "CryptoNoteCore/Currency.h"
-#include "NodeRpcProxy/NodeRpcProxy.h"
-#include "WalletLegacy/WalletHelper.h"
-
-#include <Logging/LoggerRef.h>
+#include <cctype>
```
mtl1979: Missing copyright block again...

src/SimpleWallet/Tools.cpp
```diff
@@ -0,0 +1,65 @@
+#include <SimpleWallet/Tools.h>
```
mtl1979: Missing copyright block...

src/SimpleWallet/Tools.h
```diff
@@ -0,0 +1,103 @@
+#include <cctype>
```
mtl1979: Like a parrot, we want copyright block at the top of each file...

src/SimpleWallet/Transfer.cpp
```diff
@@ -0,0 +1,967 @@
+
```
mtl1979: Missing copyright block...

src/SimpleWallet/Transfer.cpp
```diff
@@ -0,0 +1,967 @@
...
+                      << "this is because some of your balance is used when "
+                      << "sending another transaction to help hide the size "
+                      << "of your transaction, and is locked for a short "
+                      << "time. It will return shortly." << std::endl
+                      << "Needed balance: " << formatAmount(neededBalance) 
+                      << std::endl << "Available balance: " 
+                      << formatAmount(wallet.getActualBalance())
+                      << std::endl << "Locked balance: "
+                      << formatAmount(wallet.getPendingBalance())
+                      << std::endl << "Will try again in 5 seconds..."
```
mtl1979: ̀5 seconds is too short time to wait for balance to unlock... Usually it takes at least as long as one new block is mined and there is no guarantee it will be in the very next block. Too short time might cause unnecessary spam in console output.

src/SimpleWallet/Transfer.cpp
```diff
@@ -0,0 +1,967 @@
...
+           start is attempting to split them into chunks of 55k bytes or so.
+           We then check at the end that each transaction is small enough, and
+           if not, we up the numTxMultiplier and try again with more
+           transactions. */
+        int numTransactions = numTxMultiplier * 
+                              (std::ceil(double(txSize) / double(maxSize)));
+
+        /* Split the requested fee over each transaction, i.e. if a fee of 200
+           TRTL was requested and we split it into 4 transactions each one will
+           have a fee of 5TRTL. If the fee per transaction is less than the min
```
mtl1979: Missing space between "5" and "TRTL".

src/SimpleWallet/Transfer.cpp
```diff
@@ -0,0 +1,967 @@
...
+
+        uint64_t totalCost = p.destinations[0].amount + totalFee;
+        
+        /* If we have to use the minimum fee instead of splitting the total fee,
+           then it is possible the user no longer has the balance to cover this
+           transaction. So, we slightly lower the amount they are sending. */
+        if (totalCost > wallet.getActualBalance())
+        {
+            p.destinations[0].amount = wallet.getActualBalance() - totalFee;
+        }
```
mtl1979: This is bad...

src/SimpleWallet/Transfer.cpp
```diff
@@ -0,0 +1,967 @@
...
+                                        "confirmed.");
+            }
+            else
+            {
+                std::cout << WarningMsg(" fusion transactions still to be "
+                                        "confirmed.");
+            }
+            
+            std::cout << std::endl
+                      << SuccessMsg("Will try again in 5 seconds...")
```
mtl1979: 5 seconds is likely to be way too short period before retrying...

src/SimpleWallet/Transfer.cpp
```diff
@@ -0,0 +1,967 @@
...
+                      << "than your full balance to get the transfer to "
+                      << "succeed." << std::endl << "This is possibly due to "
+                      << "dust in your wallet that is unable to be sent."
+                      << std::endl;
+        }
+        /* The internal node error I believe is caused by the same issue as
+           the mixin error. Rocksteady explained this as not enough traffic
+           having occured on the network to allow your to mixin with.
+           Hopefully, this will only occur on the testnet and not the main
+           network. It seems sending multiple smaller transactiosn will
```
mtl1979: "smaller transactions"

src/SimpleWallet/Transfer.cpp
```diff
@@ -0,0 +1,967 @@
...
+           provide the network with more change to allow tx's to go through.
+           However, in some wallets that have only recieved one big single
```
mtl1979: "only received"

src/SimpleWallet/Transfer.h
```diff
@@ -0,0 +1,82 @@
+#include <string>
```
mtl1979: Missing copyright block...

ZedPea: Thanks for the review @mtl1979 , I've fixed most of the issues pointed out.

In regards to the 5 second refresh time, I wanted to make the UI appear responsive and confirm that it is still doing work, as the current simplewallet quite often you'll do something and it will just hang, and it's unclear if it's frozen or not. 
5 seconds is usually too short for a block to be found, but if we up it to 30 seconds or so and the block is found very early on whilst the thread is sleeping, we'll be waiting for an unneeded 25 extra seconds or so, which is quite a bit of wasted time from a user's point of view, in my opinion. I think it's a decent tradeoff.

In regards to the copyright notices, I'll consult with rocksteady and see what he wants me to put up there if anything.

Finally, regarding the total amount being sent slightly smaller due to the fee increasing when splitting a transaction, this is not exactly ideal but before a transaction splitting is attempted, the user is warned of this and that the total amount sent or total fee sent might decrease/increase, and they have the option of cancelling if this is not satisfactory.

mtl1979: In regards to 5 second refresh time, I think it should only print remaining transaction count when it changes... This reduced "spam" when the confirmation takes more than few iterations.

My personal opinion about sending less than user requests is definitely unacceptable and it should definitely abort instead of trying to reduce the amount.

ZedPea:
License headers have been added.

mtl1979:
To know how many times you can split a transaction, divide the requested fee by minimum fee and take the integer part. Then you can simply fail with inadequate fee if it is required to split more times. This allows the logic to be as simple as possible, but also accurate in terms of requested transfer amount and required fee.

RocksteadyTC:
@mtl1979 Thanks for your help with this
@ZedPea great work as always

SoreGums:
might want to publish the renamed previous simplewallet or implement get balance / transfer api endpoints :)

SoreGums:
I agree with everything you've said. Pool ops are the ones going to cry foul. "I run pool mining for n coins, why should I have to change just because trtl made improvements and chucked out the kitchen sink that worked fine?"

So however this is going to be handled, we'll need to have an action plan so we're singing off the same song sheet etc

mtl1979:
Like I said earlier, it would be nice if all ForkNote coins could implement the patch... Then TurtleCoin would not be the only coin that could use walletd for both pools and exchanges. Using walletgreen for simplewallet makes it possible to use same wallet files on simplewallet and GUI wallets without intermediate converting. Not all users have high-end enough hardware to use GUI wallets but they would like to have as many features GUI wallets have as possible.

bobbieltd:
Just for info, on one another forknote coin, by using walletd, I avoid simplewallet memory leak (or heap corruption, I don’t know what it is, so it’s my guess).
I think this sh*t memory bug in simplewallet caused my turtle daemon freeze and turtle pool forking before.
Most pool owners will object this move for sure.

If simplewallet is the cause of pool forking (and alot of alternative blockchains in the turtle network), switching to walletd is a must. I’m not sure about this.

I changed my mind and now support this move to walletd.

PR #99, opened 2018-03-17
fix seed node missing port to avoid warning on node start

alanwilhelm:
trival fix for warning: `2018-Mar-17 23:01:05.795769 ERROR   Failed to parse seed address from string: '174.138.68.141'`


RocksteadyTC:
good catch thanks :)

PR #102, opened 2018-03-20
Minor README edit

Swardu:
ubuntu => Ubuntu


bobbieltd:
Sorry to hijack your post. 
Weird enough. The Makefile has wrong paths more than a week but no one says anything or makes PR.

ZedPea:
The makefile shouldn't be up there at all, I think rock accidentally added it.

Issue #85, opened 2018-03-05
print_bc = no output


SoreGums:
windows 10 x64
TurtleCoin v0.3.2.1253 (0.1-g7fcc6ad)

```
print_bc
need block index parameter
print_bc  238560
print_bc  238561
print_bc  238550 238561
print_bc  238550 238565
end block index parameter shouldn't be greater than 238562
print_bc  238562 238561
end block index should be greater than starter block index
print_bc  238563
end block index should be greater than starter block index
print_bc  238564
end block index should be greater than starter block index
print_bc  238561
:(
```

![image](https://user-images.githubusercontent.com/317/36956107-16487bc6-205f-11e8-88d2-4ece565a63c5.png)

everything else about print_cn seems to work with regards to validating the params of the command, once you get the params correct it seems to just bail.

yamiM0NSTER:
As of right now print_bc is only placeholder and has no functionality implemented for printing out blockchan info in given blocks.

SoreGums:
this appears to be the answer ;)

created the issue as RockSteady thought it should be working.
what isn't tracked, isn't worked on etc.

SoreGums:
I've been playing with this in v0.4.2
Works pretty good, thanks @yamiMonster

PR #107, opened 2018-03-22
print_bc functionality

yamiM0NSTER:
1. commit to test if output is correct

Any feedback, tips on improving and suggestions much appreciated.

In final version should solve #85


SoreGums:
maybe could make sure major version is ascii like the rest?

image

yamiM0NSTER:
Improved print_bc function:

Can be called to print info about 1 block header as following:
`print_bc 295000`
Can be called to print info about block headers in given range:
`print_bc 295000 296000`

Fixed issue where last block could not be printed.
Fixed issue where block 0 could not be printed.
Function should be ready to be used in JsonRpc if there is need for it in the future.

RocksteadyTC:
@ZedPea I see the thumbs up, is this ready to go?

ZedPea:
I've not tested I'm afraid, I just thumbs up everything I see that looks cool :heart: @RocksteadyTC

PR #121, opened 2018-03-25
Format money with thousand separators

ZedPea:
Makes large balances a bit easier to read.

Before: `Available balance: 3139485.50 TRTL`
After: `Available balance: 3,139,485.50 TRTL`

Unfortunately this doesn't have locale support. Read my long essay in Tools.cpp I wrote for justification on that.

Commas are removed when people are inputting, so you can have a transfer input of `123,456.78` and it will parse fine, for example if the user copy pastes their balance.

### DIFF
--- a/src/SimpleWallet/Tools.cpp
+++ b/src/SimpleWallet/Tools.cpp
@@ -32,20 +32,66 @@ void confirmPassword(std::string walletPass)
 
 std::string formatAmount(uint64_t amount)
 {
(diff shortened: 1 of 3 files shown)


RocksteadyTC:
Is this ready to go? Tested Windows, Mac, Linux?

ZedPea:
Works fine on linux, I haven't tested on Windows/Mac but I assume it will be working. I can test on Windows tomorrow.

PR #124, opened 2018-03-28
Default simplewallet mixin, simplewallet bug fixes

ZedPea:
Adds a default mixin of 3 to simplewallet. This is waived if the transaction is dust and cannot be sent without lowering it. This can of course be bypassed by compiling without the restrictions but is a good start to introducing a minimum.

Also fixes issues with being unable to send on testnet.

I couldn't find a way to easily go from a prefix to a string to print out the expected string, so for now it just prints out `TRTL`.

Also fixes windows crashing with "simplewallet.exe has stopped working" when you ctrl+c - it's doing this when exit(0) is called, not sure if this is intended behaviour or possibly something being cleared up calling terminate(), for now it is hacked around :disappointed:  

Also fixes being unable to connect to daemon sometimes on windows - was able to reproduce 100% in windows 7 VM but worked fine for soregums on I think windows 10... was just using 0.0.0.0 instead of 127.0.0.1.

### DIFF
--- a/src/CryptoNoteConfig.h
+++ b/src/CryptoNoteConfig.h
@@ -53,6 +53,7 @@ const size_t   CRYPTONOTE_COINBASE_BLOB_RESERVED_SIZE        = 600;
 const size_t   CRYPTONOTE_DISPLAY_DECIMAL_POINT              = 2;
 const uint64_t MINIMUM_FEE                                   = UINT64_C(10);
 const uint16_t DEFAULT_MIXIN                                 = 5;
+const uint16_t MINIMUM_MIXIN                                 = 3;
 const uint64_t DEFAULT_DUST_THRESHOLD                        = UINT64_C(10);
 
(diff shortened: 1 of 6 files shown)


yamiM0NSTER:
Weird that address length is hardcoded 99 O_o

bobbieltd:
I checked testnet mine2gether. Why aren’t there any payments on testnet pool ? It’s great to have a testnet to avoid unexpected issues.

sumlnoether:
The Mine2Gether testnet pool is just to test the algorithm for the miner devs.

ZedPea:
This wouldn't effect pool payments btw, because they are either using old simplewallet or walletd which doesn't have this issue. This is just for sending testnet TRTL to people without having to recompile.

ZedPea:
Idk why the build is failing on some random files I haven't touched... it worked fine on both linux and windows when I built

PR #139, opened 2018-04-03
if wallet sync is stuck, save wallet to cause sync process to attempt…

ereptor81:
… a restart before warning user. Also updated the stuck messaging to be a little more succinct, but we may want to think about that a little more and decide if we want more troubleshooting info in the message or not.

Example output when save happens:
screen shot 2018-04-03 at 11 37 36 am


ZedPea:
Wow, this is awesome. Nice find!

Issue #126, opened 2018-03-29
Daemon failing to parse data-dir from config file


netmonk:
After compiling master branch on my local server. i wanted to test the use of config file: `--config-file`. 

As i can run multiple daemon instance on same server, i use to have a dedicated `data-dir` for each of them. 

I tried to start the daemon with this config file : 
```
plonky@nexus:~/work$ cat TurtleCoin.conf 
data-dir=/data/turtle/.TurtleCoin
rpc-bind-port=15000
p2p-bind-port=15001
rpc-bind-ip=0.0.0.0
```

And i got this result : 
```
plonky@nexus:~/work$ ./turtlecoin/build/src/TurtleCoind --config-file /home/plonky/work/TurtleCoin.conf 
Failed to parse arguments: unrecognised option 'data-dir'
Allowed options:
```
When i comment the data-dir line, the server is starting very fine, and i need to add `--data-dir /data/turtle/.Turtlecoin` in the command line next to the `--config-file TurtleCoin.conf`

There is some kind of issue here ! :)

netmonk:
Just to add a little more info, i got a strange issue with `--data-dir` and docker : 
i was unable to successfully start container with  `ENTRYPOINT "/home/turtled/bin/TurtleCoind" "--data-dir /turtlecoin/"` , but i successfully started it with `ENTRYPOINT "/home/turtled/bin/TurtleCoind" "--data-dir /turtlecoin/"`

I was quite disappointed cause, i never user a `=` when using standard bash command line to run it.

ereptor81:
@netmonk I think you need to separate each arg when using entrypoint like that. This should probably work:

ENTRYPOINT "/home/turtled/bin/TurtleCoind" "--data-dir" "/turtlecoin/"

netmonk:
Well for docker i can understand, and i find a similar workaround. 
Still for direct run, in the first post, the issue still remains from config file.

ereptor81:
That issue is fixed in the above referenced PR, not yet merged into master.

PR #143, opened 2018-04-04
allow data-dir to be set in turtlecoind config file

ereptor81:
Fixes https://github.com/turtlecoin/turtlecoin/issues/126


ZedPea:
While you're at it when I tested quickly I think rpc-password doesn't work either. Neither does mnemonic-seed either I imagine.

ereptor81:
Ok I'll take a look at those.

ereptor81:
Ok I've got the walletd ones working as well, but I'm going to add those in a separate PR b/c it needs a good bit more testing still.

PR #144, opened 2018-04-04
Pool wallet re-upload

ZedPea:
Allows easier migration for pools who don't want to upgrade to walletd yet, are using incompatible pools like snipa's, etc.

Adds a deprecation message on startup warning users that this is no longer supported.


ZedPea:
Ah darn this includes the other changes I made earlier in another PR

PR #136, opened 2018-04-02
Building Tests is now optional

SoreGums:
simply add -DDO_TESTS=OFF to cmake.
also updated release script to skip.

cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS="-fassociative-math" -DCMAKE_CXX_FLAGS="-fassociative-math" ../..; time make -j4

real    8m12.852s
user    29m10.744s
sys     2m23.076s

>>>

cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS="-fassociative-math" -DCMAKE_CXX_FLAGS="-fassociative-math" -DDO_TESTS=OFF ../..; time make -j4
real    3m21.756s
user    11m41.880s
sys     1m3.640s

That is a lot faster. Digital Ocean Droplet 4vcpu-8gb.

Also removed connectivity tool from release script

PR #149, opened 2018-04-05
Additional tests can be skipped on build

brandonlehmann:
This supplements #136 by @SoreGums as we don't need to build gtests either in the event DO_TESTS is set to off.


RocksteadyTC:
Thanks @brandonlehmann

PR #159, opened 2018-04-09
Cleanup some build issues and compiler warnings

brandonlehmann:
- Moves the RocksDB build process into the build tree instead of building it in the source tree. This keeps the source clean and prevents someone from committing something they didn't mean to.
- Resolves all of the unused variable warnings at compile time. (Not proud of how)
- Resolves a singular compiler warning regarding comparisons between signed and unsigned integers
- Resolved a compiler warning indicating that a function is completely unused
- Renames the project from ByteCoin to TurtleCoin and updated the README and AppVeyor build directions to reflect such

Although I've tested this on *nix and Windows. Please perform additional testing before merging.


SoreGums:
not much difference for Windows
https://gist.github.com/SoreGums/2e6da6cc1c7121f2703972fc760e9584
got 3x extra warnings though

brandonlehmann:
The large majority of the warnings are aliasing issues. Those will take a bit more work to sift through as they may or may not be superficial. I'll see what I can do with some of the Windows ones as well as I did see a few pop up I didn't see before that I believe are linked to the boost version.

brandonlehmann:
Pushed more updates that resolve more of the compiler warnings in Windows builds.

PR #161, opened 2018-04-10
Add timestamps to simplewallet list_transfers

ZedPea:
Caveats:
* It's slower to list all transactions because it has to scour the blockchain
* I had to comment out the serialization of block.penalty due to binary serialization not supporting doubles.

Screenshot:
selection_001

It should use the users' timezone. It appears to work on GMT, but I haven't tested other time zones.
Both linux and windows work fine - though if TurtleCoind is being slow to respond it will hold up the process quite a bit.


ZedPea:
Hold off on merging. Just found a rare segfault.

PR #162, opened 2018-04-11
Optional checkpoints

ereptor81:
Moved checkpoints out of CryptoNoteConfig.h into their own file.

Added parameter --load-checkpoints <default|filename> to optionally use checkpoints compiled in (default) or from a csv file in the format height,hash

Example checkpoints csv available here: https://s3.amazonaws.com/turtle-blockchain/checkpoints.csv

Not yet tested on Windows, but nothing included that should cause issues. I'd like to get some other peoples test for speeds of syncing the default vs full checkpoint csv as well.


SoreGums:
I'll be home in around 2hrs or so, will pull and run on windows then ;)

RocksteadyTC:
Checkpoints are successfully activated on Ubuntu LTS. However, I'd like to provide the rest of the checkpoints that are not present to provide the fastest sync possible for those that opt for the checkpointed daemon.

RocksteadyTC:
Assuming this is a GO on Windows, we should go ahead and merge it in for the 0.4.3-4 pre-release and push it to release following that.

SoreGums:
here you go windows 10 x64 results 
u: & x: are ramdrives.

--load-checkpoint default
```
time ./TurtleCoind.exe --load-checkpoints default --rpc-bind-ip=127.0.0.1 --rpc-bind-port=11901 --p2p-bind-port=11900 --data-dir u:/ --log-file=x:/log11900.log
real    66m44.954s
user    0m0.000s
sys     0m0.000s

height @ 359260
```

--load-checkpoint checkpoints.csv
```
time ./TurtleCoind.exe --load-checkpoints u:/checkpoints.csv --rpc-bind-ip=127.0.0.1 --rpc-bind-port=11901 --p2p-bind-port=11900 --data-dir u:/ --log-file=x:/log11900.log --allow-local-ip --add-exclusive-node 127.0.0.1:11800
real    33m40.567s
user    0m0.000s
sys     0m0.000s

height @ 359332
```

ereptor81:
@RocksteadyTC The csv I linked to on s3 has almost all of the checkpoints already, this script can be used to top it off: https://gist.github.com/ereptor81/71912f41cfd84d5c39f6ec67aadbe1a7 . Just enable block explorer and pass the minimum block to stop at to the script. 

It's a little slow using that to generate the full list b/c it uses the block explorer calls and only gets 30 blocks at a time, but it's not bad since the bulk of them already exist and you can just generate the last bit and append. Probably need to make a script to call the bulk binary block endpoints and decode them so checkpoints can be generated quickly.

PR #168, opened 2018-04-14
Better error message on opening a non wallet file

ZedPea:
This occurs if you open a corrupted wallet, or just some random file.


SoreGums:
works well - any chance it can go back to the start instead of exiting?
start menu has the ctrl+c to quit

ZedPea:
Welp, not very easily. It would be so much easier if C++ had nice optional types...

I'll look into it, might try and make my own Nothing/Just because I hate the boost equivalent.

It would be very simple to go back to the Open dialog again, but it's hard to say what the preferred behaviour is - did you accidentally open the wrong file, or is your wallet corrupted? With the former you want to reopen, with the later you probably want to reimport via keys.

SoreGums:
yeah, I opened the address file to test - then was like hmm, comment :)

Issue #169, opened 2018-04-15
simplewallet - maybe show the payment id/mixin for confirmation?


SoreGums:
![image](https://user-images.githubusercontent.com/317/38777597-d7681f92-40d4-11e8-99af-5e9671b2da7e.png)

might as well show everything :)

RocksteadyTC:
PENDING RELEASE, CLOSING THIS LOLZ LMAO YOLOOO

PR #170, opened 2018-04-15
Add Payment ID and MixIn in transaction confirmation

yamiM0NSTER:
Solves #169


SoreGums:
works great - cheers
image

why all the commits?

RocksteadyTC:
Because it's Yami. Just doing the work required without a fancy Irish jig at the end would be unusual.

Issue #171, opened 2018-04-15
v0.4.3 crashing on launch for some people


ZedPea:
As the title says, v0.4.3 has been crashing on launch for some users, whilst v0.4.2 is fine. 

It's reported to crash with TurtleCoind, and possibly walletd, but nothing else.

Some people appear to get a "TurtleCoind has stopped working" whilst for others it just instantly closes. The obvious difference between the two releases is the RocksDB upgrade. If someone who could reproduce this could do a `git bisect`, that would be great to confirm if it is due to RocksDB or not.

Nothing is seen in command prompt if the process is launched from there. I'm not sure if anyone has tried running with `--log-level 4` yet.

People who have this issue have also reported that compiling doesn't fix the issue. I have seen reports of this on both Windows 10 and Windows 7, but not anyone having it on GNU/Linux yet.

One person who experienced this suggested that the error message they were receiving indicated the crash was due to an "Illegal Instruction", which is commonly seen when software is ran on hardware it's not compiled for - This is not the case however if people can reproduce it when compiling themselves presumably.

ZedPea:
Turtle? gave this error information, running walletd on Windows 7.

```
Problem signature:
  Problem Event Name:    APPCRASH
  Application Name:    walletd.exe
  Application Version:    0.0.0.0
  Application Timestamp:    5acea4cc
  Fault Module Name:    walletd.exe
  Fault Module Version:    0.0.0.0
  Fault Module Timestamp:    5acea4cc
  Exception Code:    c000001d
  Exception Offset:    00000000002e3d8b
  OS Version:    6.1.7601.2.1.0.256.1
  Locale ID:    2057
  Additional Information 1:    670f
  Additional Information 2:    670f4e457958ecb5a65cbae63a61a4ad
  Additional Information 3:    5498
  Additional Information 4:    549806130bd7a661512d53f284373a32
```

brandonlehmann:
Does this occur if you nuke the DB folder and p2pstate.bin?

ZedPea:
I'm fairly sure someone tried this yesterday and it still occurred. Clarification would be good though.

ZedPea:
Just got Turtle? to test this, still occurs when no TurtleCoin folder is present

brandonlehmann:
Can he confirm the file version when right clicking on the program, properties, details? Is he launching it with or without arguments?

brandonlehmann:
And is he using precompiled or compile himself?

[...]

PR #172, opened 2018-04-16
Detect support for AVX/AVX2 in rocksdb

yamiM0NSTER:
Solves  #171. People with old hardware (pre-Haswell and old AMD CPU's) will have to compile source themselves or download special binaries for older architecture.

### DIFF
--- a/external/rocksdb/CMakeLists.txt
+++ b/external/rocksdb/CMakeLists.txt
@@ -1017,3 +1017,61 @@ option(WITH_TOOLS "build with tools" ON)
 if(WITH_TOOLS)
   add_subdirectory(tools)
 endif()
+
+
+if(MSVC)
+# Check for the presence of AVX and figure out the flags to use for it.
+	string (REPLACE "/arch:AVX2" ""     CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+	string (REPLACE "/arch:AVX" ""     CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+
+    include(CheckCXXSourceRuns)
+    
+    # Check AVX
+    check_cxx_source_runs("
+        #include <immintrin.h>
+        int main()
+        {
+          __m256 a, b, c;
+          const float src[8] = { 1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 6.0f, 7.0f, 8.0f };
+          float dst[8];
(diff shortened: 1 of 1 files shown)


yamiM0NSTER:
Please someone with old hardware test before PR is merged.

external/rocksdb/CMakeLists.txt
```diff
@@ -1017,3 +1017,61 @@ option(WITH_TOOLS "build with tools" ON)
...
+          return -1;
+        }
+      }
+      return 0;
+    }"
+    HAVE_AVX2_EXTENSIONS)
+
+  # Set Flags
+  if(HAVE_AVX2_EXTENSIONS AND NOT MSVC_VERSION LESS 1800)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /arch:AVX2")
```
brandonlehmann: Can we added a message during the cmake via message(STATUS "RocksDB: AVX2 Enabled") / Disabled / Etc?

yamiM0NSTER: it shows HAVE_AVX_EXTENSIONS test and result of it, same for avx2 one

brandonlehmann: Yes, yes, see it was unclear in the test was for RocksDB or for TurtleCoin. If not, I'll deal.

ZedPea:
How much speed do these instructions add?

I think it's pretty anti user to have them downloading a binary that doesn't work, and then they have to go into discord and ask what's up to find they have to download another version. Furthermore, if you're using a GUI wallet, you'll have to download a different version of walletd and drop it in, which is the opposite of easy.

If possible we should do run time checks for the instructions (is that possible? I think you were talking about it before) or just disable the AVX/AVX2 support. It's also not an easy one to download, most users know what version of windows they're using, but do they know what CPU features they have? Unlikely.

yamiM0NSTER:
Uh as you might have seen the gains were noticable. The problem is with people running 6 generation old CPU's (pls bury the old hardware).
About removing AVX/AVX2 I proposed PORTABLE build for releases but was met with "make it smart".

brandonlehmann:
@ZedPea  long term goal is to detect at runtime. The quick fix is to update the release files with ones built without AVX2 or AVX (I have a build right now that's okay). Further release builds will either (a) build w/o AVX2/AVX or; (b) have code that detects it at runtime.

PR #181, opened 2018-04-17
fix `multi_installer.sh` path in README.md

pypper:
it moved to /scripts, update readme to reflect that.


RocksteadyTC:
Thanks @pypper Good eye!

PR #184, opened 2018-04-18
Fix not connecting to remote node initially when all args specified

ZedPea:
It seems if you specified all args needed on the command line, 50% of the time the wallet wouldn't connect to turtlecoind quickly enough, and so would report it is closed. We run the update function in init() once before returning so it is fully inited when it returns.

Also save periodically when scanning chain initially so progress isn't lost if a user exits before it's complete.

### DIFF
--- a/src/NodeRpcProxy/NodeRpcProxy.cpp
+++ b/src/NodeRpcProxy/NodeRpcProxy.cpp
@@ -160,6 +160,8 @@ void NodeRpcProxy::workerThread(const INode::Callback& initialized_callback) {
       m_cv_initialized.notify_all();
     }
 
+    updateNodeStatus();
+
     initialized_callback(std::error_code());
 
     contextGroup.spawn([this]() {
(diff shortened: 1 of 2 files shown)


hensleyrob:
I've seen this occur when not providing the password as an argument and typing the password interactively. Perhaps this should be done anytime when using a remote node, not just when both wallet file and password are provided as arguments?

ZedPea:
@hensleyrob Thanks for the info, done.

elite5472:
I'm in Brazil and it takes more than 2 seconds for it to start accepting the request. I think its best to just wait until the httprequest is performed?

ZedPea:
@elite5472 I'm not really sure what's going on under the hood - it should wait to the request to complete, but we have no way of controlling this really. The call that this makes is returning a local variable rather than making a new request iirc, however I would think that after node.init() returns, everything should be fully initiated rather than having to add random timeouts...

I need to look more into what's actually going on in init().

elite5472:
I'd help if I knew C++, sorry man xD

ZedPea:
Ok, it looks like it returns the callback before the update loop starts. If we just run this update once before returning, it should be updated after the init() method returns.

PR #185, opened 2018-04-19
AVX/AVX2 Now Optional via CMAKE

brandonlehmann:
* Added CMAKE flag ENABLE_AVX (default: OFF)
* Only builds RocksDB with AVX/AVX2 detection if ENABLE_AVX=ON


CMakeLists.txt
```diff
@@ -8,7 +8,8 @@ set(CMAKE_CONFIGURATION_TYPES Debug RelWithDebInfo Release CACHE TYPE INTERNAL)
 set(CMAKE_SKIP_INSTALL_RULES OFF)
 set(CMAKE_SKIP_PACKAGE_ALL_DEPENDENCY ON)
 set(CMAKE_SUPPRESS_REGENERATION ON)
-SET(DO_TESTS ON CACHE STRING "Spend time making tests? Defaults to ON")
+set(DO_TESTS ON CACHE STRING "Spend time making tests? Defaults to ON")
```
SoreGums: i noticed this while going through the version stuff, was like "what?" nice that you spotted it and corrected 👍

yamiM0NSTER:
this won't work

CMakeLists.txt
brandonlehmann: It happens :)

brandonlehmann:
Would you please be more descriptive? I've confirmed that it DOES work on MSVC without issue.

brandonlehmann:
@yamiM0NSTER  Perhaps instead of saying "it's just done in a crap way" you'd like to suggest a better method? Constructive criticism is far more valuable than that negative comments without support.

yamiM0NSTER:
Yikes, my bad. Had a moment off from my activities so took quick look. From github changes it looked like some crappy if statement usage but now that I had time I checked the resulting file and it looks alright. Sorry for the statements, github's representation and indentation style didn't help, but in the end I made wrong statements. Once again my bad. Not sure why @RocksteadyTC merged previous PR before final decision was set but it's up to him.

PR #202, opened 2018-05-01
Add ascii art on daemon startup and sync

ZedPea:
Ascii art is A E S TH E T I C

image


ZedPea:
Looks nicer in windows imo

image

SoreGums:
adjust the settings?
different font

RocksteadyTC:
Any thoughts on doing an ascii art contest?

ZedPea:
Yeah I like the idea, yami and turtle? already put some nice ones in #dev_general

RocksteadyTC:
Merging for now, then we can replace when we do the contest. My logic is that people will see it, make note of it in the chat, then we can hit them with the `!tag contest` and have it tell them about the ascii contest and give the constraints/proportions.

holytastyguacamole:
image

PR #208, opened 2018-05-03
Updated TurtleCoin version

andrewnk:


RocksteadyTC:
That was a temporary release to check the CI toolchain, and it got deleted, BUT, the next release will have the same tag, so I'm leaving it because I'll forget if I remove it. Thanks Andrew!

PR #210, opened 2018-05-05
Remove simplewallet file logger

ZedPea:
Don't think we need to log to a file anymore, haven't found any issues that have needed the log for. Gets kinda big too.

### DIFF
--- a/src/SimpleWallet/SimpleWallet.cpp
+++ b/src/SimpleWallet/SimpleWallet.cpp
@@ -34,17 +34,8 @@ int main(int argc, char **argv)
         return 0;
     }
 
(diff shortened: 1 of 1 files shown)


CaptainMeatloaf:
Perhaps it would be useful if someone changes the log level from the normal one for debugging purposes?

ZedPea:
Not 100% what you mean. This has always been set to a level of debug, and is independent to the turtlecoind logging level.

I don't think it will be helpful in any cases, because it's just the output from walletd really, which is very sparsely logged, and never seems to go wrong. You end up with a 30MB file after only a few days of running simplewallet.

It seems basically every single line is just the blockchain synchronizer printing out what it's doing.

PR #221, opened 2018-05-08
Verify mixin of block transactions

ZedPea:
To follow up from the previous patch, @SoreGums pointed out that that patch only prevented transactions with mixin out of bounds being included in the pool, but an un-updated/malicious daemon could still include the transaction into a block, which would then cause daemons to crash regardless.

This has tentatively been put at an upgrade height of 450,000. We cannot instantly apply this, as it would prevent new users syncing any previous blocks which fell out of the current mixin bounds. However, this is not hard fork required. Provided no-one submits a block with out of bounds mixin nothing will happen, and if one is, if enough pools are updated then they will just ignore that block.

These bounds have been renamed to be versioned, when mixin bounds change in the future, we can check if the height is above the upgrade point, and apply the mixin bounds for when the block was submitted.

If that's a bit confusing, what I mean is if someone submitted a block today with 1 mixin, that would be valid. If we later update the bounds to ban mixin below 3 for example, that block is still valid at the time it was created. So, version our bounds.

### DIFF
--- a/src/CryptoNoteConfig.h
+++ b/src/CryptoNoteConfig.h
@@ -57,8 +57,9 @@ const uint16_t DEFAULT_MIXIN                                 = 5;
    minimum_mixin_no_dust = enforced for simplewallet, when dust is not present
    if dust is present, 0 mixin allowed. Possibly later disabled, or relegated to sweep_unmixable */
 const uint16_t MINIMUM_MIXIN_NO_DUST                         = 3;
-const uint16_t MINIMUM_MIXIN                                 = 0;
-const uint16_t MAXIMUM_MIXIN                                 = 100;
+const uint16_t MINIMUM_MIXIN_V1                              = 0;
+const uint16_t MAXIMUM_MIXIN_V1                              = 100;
+const uint32_t MIXIN_LIMITS_V1_HEIGHT                        = 450000;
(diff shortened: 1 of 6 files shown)


RocksteadyTC:
Good work. I like the idea of progressively updating the bounds. We should think about when the next will be, and by how much, judging by the response we get from this.

SoreGums:
Well re-read it all - and groked the code a bit - looks sweet _(I'm not an expert...)_

As you've commented in Discord, the miners really determine blocks, as long as the pools adopt the master with this commit, everything should be fine as they have the most hash and will reject blocks that don't follow the mixin boundaries (0-100). Then solo miners will get orphans complain and update, network is sane/protected!

Hopefully, there isn't an off by 1 issue regarding the calculation. As you said, if you specify mixin 0 - it isn't really zero because of the need to have at least one input otherwise where are the outputs going to be generated from!? 

Still though, did your 2,000 mixin tx actually have 2,001 inputs. I don't get this still when looking at the block explorer. High mixins rarely show the equiv number of inputs. Something I hope to uncover as I write docs about low-level functionality of the project.

ZedPea:
From my experience they do show +1 on the block explorer when you look up the transactions, aye. Haven't got a tx hash to demonstrate, but when I was sending a few with 90 mixin, they all showed up as 91 mixin. I actually opened an issue on the block explorer, as I though it was an off by one, but it's correct.

PR #222, opened 2018-05-08
Make Release Builds Great Again

brandonlehmann:
* Hides RocksDB CPU feature detection unless enabled manually to prevent building CPU specific binaries (unless you really want to)
  * ENABLE_LEAF_FRAME
  * ENABLE_SSE42
  * ENABLE_THREAD_LOCAL
  * ENABLE_SYNC_FILE_RANGE_WRITE
  * ENABLE_PTHREAD_MUTEX_ADAPTIVE_NP
  * ENABLE_MALLOC_USABLE_SIZE
  * ENABLE_SCHED_GETCPU
* Changes the default build architecture from **native** to **default** to prevent building CPU specific binaries (unless you really want to)
* Changes the default build to **STATIC** to prevent building system specific binaries (unless you really want to)
* Changes the default build type to **Release** to prevent building release binaries with Debug Info (unless you really want to)


RocksteadyTC:
Good job @brandonlehmann 
Do you think we have enough of a userbase that needs the CPU specific binaries that we should include those instructions in the README for those that need them?

RocksteadyTC:
If it is just to help a few of our friends out, I guess no worries, but if a lot of people have old hardware, it might be good to include the compile instructions for the relevant CPUs

brandonlehmann:
Definitely not a bad idea. I'll update the the readme in another PR to include some information on how to enable those flags.

brandonlehmann:
To clarify, then additional CPU and other flags are not required... but they do enable faster instruction sets for users with newer hardware. 

Compile your own instructions will be updated with how to turn them all on (so each users build works great on their machine again).

PR #232, opened 2018-05-15
Remove f string syntax for backwards compatibility

ZedPea:


SoreGums:
works now on python 3.5.2
image

thanks

SoreGums:
i guess i could press "merge"... feel like that is not what was agreed when I was given contribs to the repo though :P

PR #236, opened 2018-05-20
Fix windows ascii art

ZedPea:
windows is C H A L L E N G E D

Uses shitty ASCII art which windows can render, tested on win 7.


ZedPea:
FUCK

PR #238, opened 2018-05-20
Fix appveyor build

ZedPea:
Appveyor removed support for boost 1.59 recently causing the automated builds to fail.

https://www.appveyor.com/docs/build-environment/#boost

Bump the boost version to the latest supported to prevent this happening for a while.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,8 +1,8 @@
 os: Visual Studio 2015
 
 environment:
-    BOOST_ROOT: C:\Libraries\boost_1_59_0
-    BOOST_LIBRARYDIR: C:\Libraries\boost_1_59_0\lib64-msvc-14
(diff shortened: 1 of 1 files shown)


SoreGums:
are the boost libs supposed to be backwards compatible?

guess I could pull it into these 16.04 Ubuntu boxes and see what happens

ZedPea:
@SoreGums I am currently using 1.66 on my linux box which works fine. The appveyor build at least completed, so I see no reason why it wouldn't work.

On a semi related note, it seems like the travis build are not firing for some reason though, not sure on that one.

RocksteadyTC:
@SoreGums since the beginning I've been leaning toward only supporting the latest Ubuntu LTS, to not keep us in this dependency spiral for legacy users.

PR #241, opened 2018-05-22
Fix always printing of the version

hensleyrob:


RocksteadyTC:
Is this in response to a conversation with some context, or is this your personal preference?

PR #245, opened 2018-05-27
Add status command to TurtleCoind

thinkpol2:
Adds status command to TurtleCoind. In addition, Core now tracks time started.


SoreGums:
nice one buddy :)

PR #257, opened 2018-06-04
Increase mixin limits and disable dust at next fork

ZedPea:
Fork height is currently set at 600,000, this probably needs to be changed to something to fit with our desired release schedule. 

This fork will set the minimum mixin to 7, and disable dust outputs. This should mean 0 mixin is no longer needed to send dust. The code has been improved to make it easier to implement new mixin forks.

One thing that needs to be tested is what happens to transactions that are currently in the transaction pool when the fork hits, if they have a valid mixin before the fork, but invalid mixin after the fork. ETN experienced this with their recent fork, so blocks were created with these invalid transactions and then rejected.

I would like to test this out on testnet to make sure it's been implemented correctly.

Edit: Currently being tested out on lwma-2 testnet

Not really worth mentioning section:

Changed simplewallet help message to only mention "optimize" instead of quick/full, in my experience full is generally the same as quick. Regardless, full and quick are still commands, just undocumented.


RocksteadyTC:
@ZedPea ready?

ZedPea:
@RocksteadyTC The fork on the testnet went well, however I was asleep so wasn't able to spam the mempool with low mixin transactions to see if blocks could still be produced.

RocksteadyTC:
Reviewed.

PR #263, opened 2018-06-05
Start of transition away from binary handlers

brandonlehmann:
* Adds JSON handlers that mimic the binary handlers in ```RpcServer```
* Changes ```NodeRpcProxy``` to use the new JSON handlers
* Fixes a few typos


brandonlehmann:
All public Turtlenode.io have been upgraded to run both this PR and #262 for additional verification.

PR #278, opened 2018-06-09
Remove mixin options

ZedPea:
Sets default mixin to 7, simplewallet users no longer have an option to choose their mixin. (GUI wallet users could, but it would be rejected by the network)


SoreGums:
how does the maxMixin++ play into this?
is it all good?
just checking ;)

ZedPea:
Yep, this just changes the previous limits from 7 to 100, to 7 to 7. I managed to send a transaction fine, but to be fair it does only come into place at 600k. I'll alter the fork height and try sending a transaction at some point to make sure.

PR #290, opened 2018-06-14
[BUGFIX] Resolves Walletd Interaction with node via --local flag

### LINKED ISSUE #286: Walletd In-Process Node Hangs on sendTransaction
@woodyjon and I tested this a bunch today. This issue occurs back to v0.5.0 and possibly earlier.

When running walletd with the --local option (in-process node) and then sending a transaction via the RPC once fully synced, walletd hangs and appears to deadlock.

Jon was able to trace it as best as possible to https://github.com/turtlecoin/turtlecoin/blob/7af7f150977f947d3704e5e6b06316d8a8f441ca/src/InProcessNode/InProcessNode.cpp#L533 and I can confirm that's where it appears to hang up. This does *not* occur when using a remote node with walletd.

I think we can get this resolved pretty quickly if we spin up the RpcServer for the daemon in conjunction with the InProcessNode and then use the NodeRpcProxy to connect to the in-process node.

Thoughts? Or would someone like to go down this rabbit hole and try to hammer this bad boy out?

brandonlehmann:
Instead of using InProcessNode for walletd to talk to the daemon that it's running, this instead launches the RpcServer the same way that the normal daemon does and then uses the NodeRpcProxy to talk the the daemon that is running in its process. While this isn't the most elegant solution, it does appear to resolve #286 

NodeRpcProxy is also uses with the remote daemon when specified -- or when --local is not supplied. So it's basically doing the same thing in 3 out of 3 cases now where it was doing something special for --local that really wasn't the best idea IMO.

***Please have MULTIPLE people test this before it makes it towards release.***

### DIFF
--- a/src/PaymentGateService/PaymentGateService.cpp
+++ b/src/PaymentGateService/PaymentGateService.cpp
@@ -34,6 +34,9 @@
 #include "CryptoNoteCore/RocksDBWrapper.h"
 #include "CryptoNoteProtocol/CryptoNoteProtocolHandler.h"
 #include "P2p/NetNode.h"
+#include "Rpc/RpcServer.h"
+#include "Rpc/RpcServerConfig.h"
+#include "NodeRpcProxy/NodeRpcProxy.h"
 #include <System/Context.h>
 #include "Wallet/WalletGreen.h"
 
@@ -211,7 +214,9 @@ void PaymentGateService::runInProcess(Logging::LoggerRef& log) {
 
   CryptoNote::CryptoNoteProtocolHandler protocol(currency, *dispatcher, core, nullptr, logger);
   CryptoNote::NodeServer p2pNode(*dispatcher, protocol, logger);
-
+  CryptoNote::RpcServerConfig rpcConfig;
+  CryptoNote::RpcServer rpcServer(*dispatcher, logger, core, p2pNode, protocol);
+  
   protocol.set_p2p_endpoint(&p2pNode);
 
   log(Logging::INFO) << "initializing p2pNode";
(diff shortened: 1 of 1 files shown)


woodyjon:
I can confirm that sendTransaction now works on local.
But there is a serious issue: everytime I launch walletd in local with my wallet, it deletes the transactions and resyncs the wallet from scratch.

woodyjon:
I want to add that your fix is a very good idea. I don't find it not elegant, as having similar things done 2 different ways did not seem optimal. Supposedly, we can now remove hundreds of lines of code.

woodyjon:
Has anyone been able to reproduce my issue? I reproduce it every time, on every platforms, with different wallet files. To reproduce:

Build latest turtlecoin (latest from dev branch, after this PR), open a wallet with walletd in --local, sync a little bit, save the wallet. Then restart walletd in --local with the same wallet, see if it re-syncs from start or from last saved point.

To launch walletd:
```
./walletd -w yourWallet.wallet -p yourPassword --local --rpc-password test
```

to save:
```
curl -d '{"params":{},"jsonrpc":"2.0","id":"test","method":"save", "password":"test"}' http://localhost:8070/json_rpc
```

to get status:
```
curl -d '{"params":{},"jsonrpc":"2.0","id":"test","method":"getStatus", "password":"test"}' http://localhost:8070/json_rpc
```

ZedPea:
To clarify, do you experience blockCount being set to zero when you reopen? I opened my main wallet, which should have been synced and it started rescanning, but upon saving and reopening the wallet, it restored where it was in the walletd sync progress.

woodyjon:
Yes, blockCount starts ok, then few seconds later goes back to zero. Did you try it 2 times in a row in --local?

ZedPea:
Ok, I didn't realise there was a pause. Now I can reproduce this. When I turn up logging, I see this:

```
11936 2018-Jun-16 21:09:55.895299 DEBUG   Shortest chain size 50392                                                                                                                                                
11937 2018-Jun-16 21:09:55.895334 DEBUG   Common history: start block index 0, sparse chain size 1
11938 2018-Jun-16 21:09:56.128007 DEBUG   Blocks received, start index 0, count 100                                                                                                                                
11939 2018-Jun-16 21:09:56.128086 DEBUG   Process blocks, start index 0, count 100                                                                                                                                 
11940 2018-Jun-16 21:09:56.129315 DEBUG   Detach consumer, consumer 0x55db0a012ee0, block index 1
11941 2018-Jun-16 21:09:56.129398 DEBUG   Adding blocks to consumer, consumer 0x55db0a012ee0, start index 0, count 100                                                                                             
11942 2018-Jun-16 21:09:56.152836 DEBUG   Last block hash 3dd8ee5b4f9a95ff989f75bf3434683afb10c8396efb0d8e668938a1155f730a, index 99                                                                               
11943 2018-Jun-16 21:09:56.152911 DEBUG   Blocks added to consumers                                                                                                                                                
11944 2018-Jun-16 21:09:56.153014 DEBUG   Starting blockchain synchronization...                                                                                                                                   
11945 2018-Jun-16 21:09:56.153039 DEBUG   Shortest chain size 101
```

So it had the correct shortest chain size from my previous partial sync, then for some reason threw it away and restarted the sync...

woodyjon:
I created an issue here #303 to have a little bit more visibility.

PR #292, opened 2018-06-16
destoy method

brandonlehmann:
This has been driving me crazy.


ZedPea:
Lolling at that branch name. There are so damn many of these spelling errors, and sometimes they use the same variable spelt right and then the variable again spelt wrong.

brandonlehmann:
There should be a CI plugin for GH that runs spellcheck on repos.

RocksteadyTC:
DESTOY THE HUMANS

PR #293, opened 2018-06-16
Hard fork at 600k to use LWMA-2 difficulty algorithm

ZedPea:
After weeks of testing and tweaking zawy has arrived on the perfect difficulty algorithm... for now!

This also applies the jagerman MTP patch to make sure we always produce a valid block template. Longer explanation in the code for the effects of this.


RocksteadyTC:
it still says don't merge. Im apprehensive. https://www.youtube.com/watch?v=4F4qzPbcFiA

PR #297, opened 2018-06-17
Update CMakeLists.txt

crappyrules:
On the seventh day, @ZedPea created the Zedwallet.  And let it be so!


RocksteadyTC:
image
congrats young man :dancing_women: Ill update your roles

PR #311, opened 2018-06-22
Prevent divide by zero with 50x attacker (thanks fireice)

ZedPea:
Don't merge yet, had to add a few extra casts to get it to compile and just need to check with zawy it doesn't change the semantics.


RocksteadyTC:
if this kills us, zed did it

ZedPea:
:cold_sweat:

PR #314, opened 2018-06-23
Readline support for ZedWallet

mrrovot:
One of the features requested in https://medium.com/@turtlecoin/zedwallet-a-new-simplewallet-for-cryptonote-currencies-2c74c5fc1302


ZedPea:
Huh, dunno why I didn't see this PR pop up. Mostly looks good, but I think this bit:

```
#ifdef HAVE_READLINE 
+        //do nothing
+        #else
+        //pring yellow prompt
+        std::string str = CYELLOW + getPrompt(walletInfo) + "" + RESET;
+        std::cout << str;
+        #endif
```

Should instead by doing `std::cout << InformationMsg(getPrompt(walletInfo));`, as I imagine the CYELLOW and RESET terminal escape code don't work on Windows.

PR #315, opened 2018-06-24
Add optional cors header to walletd

ZedPea:
Has the same syntax as turtlecoind, e.g. --enable-cors "\*"
(Make sure to quote your "*" otherwise your shell expands it!)


RocksteadyTC:
that's a big warning. I wonder if there's a way we can emphasize that

PR #319, opened 2018-06-28
Fix mnemonic seed info not being output to walletd log file

ZedPea:
We were just using std::cout to write stuff to the console, but this also should be written to the log file. The logger doesn't inherit from ostream so we're using a template instead, which means the implementation needs to be moved to the header file.

Thanks to @woodyjon for finding this issue.


woodyjon:
Tested and approved! Works as expected.
Thank you @ZedPea

PR #325, opened 2018-06-30
Update Dockerfile.test

crappyrules:
not sure if that was intentional or not.


andrewnk:
IIRC, it was intentional. This is part of @funkypenguin testnet docker swarm and was changed to ensure you wouldn't be mining on the main chain.

funkypenguin:
Yup, intentional, see the top of Dockerfile.test for an illustration of what’s changed for a testnet :)

crappyrules:
My bad.

PR #337, opened 2018-07-05
Add integrated addresses to walletd

ZedPea:
Adds integrated address support to walletd. Adds a new RPC call, createIntegratedAddress(), that takes a parameter `address`, and a parameter `paymentId`, which returns an integrated address created from the two inputs.

This rpc call will need to be added to the relevant RPC libraries and the api-docs updated.

sendTransaction() will now support integrated addresses as arguments, and will correctly convert them to a payment ID and address pair internally.

Suggested usage is sending one transfer, with one integrated address, but if for some reason you want to also specify the payment ID in the json, that will work. If the payment ID in the json and extracted from the integrated address don't match, it will error out. Similarly, if you specify multiple integrated addresses in the transfers array, it will be allowed if all of the integrated addresses contain the same payment ID.

Again, not sure why you'd want this, but hey, support is there.

Exchanges or other services that require payment ID's can instead call createIntegratedAddress() with the desired address+payment ID, and supply this to users. Users need not worry about forgetting the payment ID, and the service can continue decoding their payment ID's on the receiving end as before.

### DIFF
--- a/src/PaymentGate/PaymentServiceJsonRpcMessages.cpp
+++ b/src/PaymentGate/PaymentServiceJsonRpcMessages.cpp
@@ -353,4 +353,18 @@ void EstimateFusion::Response::serialize(CryptoNote::ISerializer& serializer) {
   serializer(totalOutputCount, "totalOutputCount");
 }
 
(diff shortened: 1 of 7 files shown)


RocksteadyTC:
I can't wait till pools adopt this!

PR #349, opened 2018-07-11
Port over karbo crypto, and some of aeon/monero's crypto (mainly slow hash)

ZedPea:
As suggested by aiwe, I ported over their crypto. I decided to port over aeon's slow hash as that will be more portable and easier to alter with future modifications.

The chain appears to sync from both fresh and from the top of the chain, though a full sync hasn't been done. I would assume as long as it can validate blocks from both algos we're good.

~~I also need to test mnemonic seeds as that changed a tiny bit, and that wallet passwords are still working.~~

Update: mnemonics work fine, wallet passwords work fine.

Regarding licenses - not sure what license header to use, some bits are karbo (but ported from monero), and some parts are aeon (but ported from monero).

I haven't actually tested if this helps rpi compilation as I don't have one.


brandonlehmann:
License should attribute both group(s)

ZedPea:
Ok, think I did the licenses correctly. Tests are broken, I don't get why they build on travis but not on my machine, maybe I am stupid :thinking:

ZedPea:
Ok, it builds now :+1:

RocksteadyTC:
.....no way

RocksteadyTC:
is this what I think it is?

crappyrules:
I'll give this a spin let me load up a pi

PR #359, opened 2018-07-14
Fix proof of work too weak on v0.6.4

ZedPea:
This patch is *untested*. My node is still syncing to confirm it works. **Please don't merge yet.**

There were some slight LWMA-2 changes which I erroneously believed were included in v0.6.3, which were not. This caused the users using 0.6.4 to not validate the proof of work supplied at block 602020 by the v0.6.3 daemons.

This patch simply sets those LWMA-2 changes to be activated at the next slated hard fork of 800,000, and uses the same code as 0.6.3 until then.

### DIFF
--- a/src/CryptoNoteConfig.h
+++ b/src/CryptoNoteConfig.h
@@ -35,6 +35,8 @@ const size_t ZAWY_DIFFICULTY_V2                              = 0;
 const uint8_t ZAWY_DIFFICULTY_DIFFICULTY_BLOCK_VERSION       = 3;
 
 const uint64_t LWMA_2_DIFFICULTY_BLOCK_INDEX                 = 620000;
+const uint64_t LWMA_2_DIFFICULTY_BLOCK_INDEX_V2              = 800000;
+
 const uint64_t DIFFICULTY_WINDOW_V3                          = 60;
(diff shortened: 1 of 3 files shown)


src/CryptoNoteConfig.h
```diff
@@ -89,6 +91,7 @@ const uint64_t DEFAULT_DUST_THRESHOLD                        = UINT64_C(10);
 const uint64_t DEFAULT_DUST_THRESHOLD_V2                     = UINT64_C(0);
 
 const uint32_t DUST_THRESHOLD_V2_HEIGHT                      = MIXIN_LIMITS_V2_HEIGHT;
```
iamamyth: It seems a bit confusing to have `DUST_THRESHOLD_V2_HEIGHT` as well as `FUSION_DUST_THRESHOLD_HEIGHT`, which, if I understand correctly, were originally intended to target the same feature.

ZedPea: Do you think it should be FUSION_DUST_THRESHOLD_V2_HEIGHT instead?

iamamyth: That option would make these changes in CryptoNoteCore/Currency.h a bit more readable to me: https://github.com/turtlecoin/turtlecoin/pull/359/files#diff-eb8416efd4da61774083345a51ef2af7R71.

ZedPea: Ok, done

PR #363, opened 2018-07-17
Removing the option to run the InProcessNode with PaymentGateService

brandonlehmann:
This removes the ability to run (walletd/service) with the --local option.

Even if you try, it won't.

👋👋👋👋👋👋👋👋👋👋👋


ZedPea:
>doesn't even remove the code
:thinking:

brandonlehmann:
*sigh* @ZedPea do you have any kind of idea what kind of harm this is going to cause my well being?

*grabs pitchfork and torch*

brandonlehmann:
There -- it's gone @ZedPea

ZedPea:
:heart_eyes: :eggplant: :sweat_drops:

PR #364, opened 2018-07-17
Full Rename of walletd to service

brandonlehmann:


brandonlehmann:
Hmmmm... oh dur.... i changed other stuff.

PR #369, opened 2018-07-19
Fixing: Fatal error: boost::bad_any_cast: failed conversion using boost::any_cast

brandonlehmann:


rashedmyt:
And the patch is working great on ubuntu 18.04.. just compiled and tested

Issue #228, opened 2018-05-13
Illegal instruction with release binaries


ZedPea:
So we're still seeing this on some CPU's on 0.5.0. Good to have somewhere to track it.
Compiling from source always fixes the issue.

From zoidan:

> I get the illegal instruction on 0.5.0 on an iMac and a MacBook Air bother with intel core 2 duos.
> Simplewallet works under 0.5.0, it didn’t under previous releases. TurtleCoind is still broke.

From Rossobimbo:

(Also 0.5.0)
> linux mint 18.3 cinamon
> cinnamon version 3.6.7
> kernal:4.10.0-38 generic
> core 2 duo cpu E6850 3GHz

I'm not sure if 0.4.2 was working for everyone, but Rossobimbo says it was for him. So probably the rocksDB upgrade causing issues again.

RocksteadyTC:
If we're thinking about yet another RocksDB adjustment, should we also think about what our experience might be like on another DB- SQLite, LMDB, Influx, etc.?

brandonlehmann:
It may not be a bad idea.

RocksteadyTC:
Assuming this is the route we could take, changing databases, let's:

1. list all options for DBs, 
2. discard the ones that can't be implemented *easily* and quickly 
3. discard from those the ones who aren't actively maintained or don't have good documentation
4. have a critical discussion about the remaining options. 

I think RocksDB has served us well, however, we may have grown out of its utility for our specific application. Even if not moving to a new DB right away, it would still help to keep our finger on a few DB projects for the sake of familiarity.

RocksteadyTC:
Is this still an issue?

RocksteadyTC:
Please ping me to reopen this if it needs to be reopened.

ZedPea:
@RocksteadyTC Occurring for a user with a Xeon E5-2670 and Ubuntu 16.04.1

thinkpol2:
Can confirm on Centos 7.5.1804 with X5650:

SIGILL: [https://pastebin.mozilla.org/9088547](https://pastebin.mozilla.org/9088547)

```
$ cat /proc/cpuinfo 
processor	: 0
vendor_id	: GenuineIntel
cpu family	: 6
model		: 44
model name	: Intel(R) Xeon(R) CPU           X5650  @ 2.67GHz
stepping	: 2
microcode	: 0x1a
cpu MHz		: 1596.000
cache size	: 12288 KB
physical id	: 1
siblings	: 12
[...]

PR #376, opened 2018-07-25
Update CMakeLists.txt To Address Illegal Instruction

brandonlehmann:
Setting a few more flags here that may help with issue #228 

Please let this pass CI tests as this is a web edit and I have not tested on all platforms yet.


brandonlehmann:
It passed. That’s good news.

PR #379, opened 2018-08-06
Rework integrated addresses to be more compact

ZedPea:
~~Don't merge yet, I don't have a synced wallet to test yet.~~

As katz pointed out, we do some weird encoding where we take the address, add a payment ID to that, and encode it with base58, along with another prefix and checksum which we've already got in the standard address.

Instead, we should just encode the paymentID + two keys, and then add the prefix and checksum to that. This brings integrated address lengths down from 231 chars to 187 chars.


ZedPea:
Ok, I got around to testing this. It works fine with both zedwallet and walletd/service - feel free to merge whenever you're ready.

PR #380, opened 2018-08-08
Improve readline cmake

ZedPea:
This displays a message when readline is not found, with how to install it.

Futhermore, Apple has some sort of readline library which gets picked up, but it is not GNU Readline, so it fails during compilation. This patch detects that, and tells the user how to fix it.

Big thanks to tmac for testing.


RocksteadyTC:
Thanks TMac :D

PR #381, opened 2018-08-09
[FEATURE] Implementing the use of the node fees in zedwallet

brandonlehmann:
@ZedPea give this a gander, I think I covered it all.

No matter how much I wanted to charge it against fusion transactions -- I didn't. Everything else I think I caught though.

### DIFF
--- a/src/zedwallet/Commands.cpp
+++ b/src/zedwallet/Commands.cpp
@@ -128,7 +128,7 @@ bool dispatchCommand(std::shared_ptr<WalletInfo> &walletInfo,
     }
     else if (command == "transfer")
     {
-        transfer(walletInfo, node.getLastKnownBlockHeight());
+        transfer(walletInfo, node.getLastKnownBlockHeight(), false, node.feeAddress(), node.feeAmount());
     }
     else if (command == "optimize")
     {
@@ -148,7 +148,7 @@ bool dispatchCommand(std::shared_ptr<WalletInfo> &walletInfo,
     }
     else if (command == "ab_send")
     {
-        sendFromAddressBook(walletInfo, node.getLastKnownBlockHeight());
+        sendFromAddressBook(walletInfo, node.getLastKnownBlockHeight(), node.feeAddress(), node.feeAmount());
     }
     else if (command == "change_password")
     {
@@ -160,7 +160,7 @@ bool dispatchCommand(std::shared_ptr<WalletInfo> &walletInfo,
     }
     else if (command == "send_all")
     {
-        transfer(walletInfo, node.getLastKnownBlockHeight(), true);
+        transfer(walletInfo, node.getLastKnownBlockHeight(), true, node.feeAddress(), node.feeAmount());
     }
     /* This should never happen */
     else
--- a/src/zedwallet/Transfer.cpp
+++ b/src/zedwallet/Transfer.cpp
@@ -147,8 +147,11 @@ bool confirmTransaction(CryptoNote::TransactionParameters t,
     return false;
 }
 
+/* Note that the originalTXParams, and thus the splitTXParams already has the
+   node transfer added */
 void splitTX(CryptoNote::WalletGreen &wallet, 
-             const CryptoNote::TransactionParameters p, uint32_t nodeFee)
+             const CryptoNote::TransactionParameters originalTXParams,
+             uint32_t nodeFee)
 {
     std::cout << "Transaction is still too large to send, splitting into "
               << "multiple chunks." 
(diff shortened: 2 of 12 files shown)


src/zedwallet/Commands.cpp
```diff
@@ -128,7 +128,7 @@ bool dispatchCommand(std::shared_ptr<WalletInfo> &walletInfo,
     }
     else if (command == "transfer")
     {
-        transfer(walletInfo, node.getLastKnownBlockHeight());
+        transfer(walletInfo, node.getLastKnownBlockHeight(), false, node.feeAddress(), node.feeAmount());
```
ZedPea: From looking at the above code, I think this is done correctly, but I want to confirm. `.node.feeAddress()` just returns a variable, rather than actually contacting the daemon, the same as `node.getLastKnownBlockHeight()`?

If so, that's fine, otherwise we could hang here if the daemon is hanging.

src/zedwallet/Transfer.cpp
```diff
@@ -98,14 +98,15 @@ bool parseAmount(std::string strAmount, uint64_t &amount)
...
+                        bool integratedAddress, uint32_t nodeFee)
 {
     std::cout << std::endl
               << InformationMsg("Confirm Transaction?") << std::endl;
 
     std::cout << "You are sending "
               << SuccessMsg(formatAmount(t.destinations[0].amount))
-              << ", with a fee of " << SuccessMsg(formatAmount(t.fee));
+              << ", with a network fee of " << SuccessMsg(formatAmount(t.fee))
+              << " and a node fee of " << SuccessMsg(formatAmount(nodeFee));
```
ZedPea: If possible, I'd prefer to not show all the node fee messages if the node fee is 0, for simplicity. Less people asking WTF is this???

src/zedwallet/Transfer.h
```diff
@@ -14,12 +14,13 @@ enum AddressType {NotAnAddress, IntegratedAddress, StandardAddress};
...
+              bool sendAll = false, std::string nodeAddress = std::string(), uint32_t nodeFee = 0);
 
 void doTransfer(std::string address, uint64_t amount, uint64_t fee,
                 std::string extra, std::shared_ptr<WalletInfo> walletInfo,
                 uint32_t height, bool integratedAddress,
-                uint64_t mixin = WalletConfig::defaultMixin);
+                uint64_t mixin = WalletConfig::defaultMixin, 
+                std::string nodeAddress = std::string(), uint32_t nodeFee = 0);
 
 void splitTX(CryptoNote::WalletGreen &wallet,
```
ZedPea: Have you tested how this works with splitTX?

src/zedwallet/Commands.cpp
```diff
@@ -160,7 +160,7 @@ bool dispatchCommand(std::shared_ptr<WalletInfo> &walletInfo,
     }
     else if (command == "send_all")
     {
-        transfer(walletInfo, node.getLastKnownBlockHeight(), true);
+        transfer(walletInfo, node.getLastKnownBlockHeight(), true, node.feeAddress(), node.feeAmount());
```
ZedPea: Is the node fee correctly removed when calculating the amount to send in send_all?

src/zedwallet/ZedWallet.cpp
```diff
@@ -127,6 +127,27 @@ int main(int argc, char **argv)
                       << std::endl << std::endl;
         }
     }
+    
+    /*
+      This will check to see if the node responded to /feeinfo and actually
+      returned something that it expects us to use for convenience charges
+      for using that node to send transactions.
+    */
+    if (node->feeAmount() != 0 && !node->feeAddress().empty()) {
```
ZedPea: Have you tested how this works with slow responding nodes? Does it hang if the node never responds?

ZedPea:
Not passing tests :grimacing:

src/zedwallet/Commands.cpp
brandonlehmann: Correct. Tries it once. Stores it and returns it. If it fails at start, oh well, we tried.

src/zedwallet/Commands.cpp
brandonlehmann: Check below. I think I pulled it out right.

src/zedwallet/Transfer.cpp
brandonlehmann: I think that adds a lot of “if” logic for something with a bunch of branches just testing for 0.

I’m on the fence 50/50

brandonlehmann:
Will check tests. I think I forgot to update the stubs in tests.

src/zedwallet/ZedWallet.cpp
ZedPea: I guess the above clarification applies here.

brandonlehmann: Yes. It timesout. Then continues as if nothing ever happened. :)

src/zedwallet/Transfer.h
ZedPea: With a very dusty wallet the user could get pretty pwned by node fees. Might need a separate disclaimer, or a suggestion to run full optimize first.

brandonlehmann: It looks like the wallet attempts to do fusion transactions before it will force a split... doesn't it? or am i misreading the code?

ZedPea: That is correct - however it only does one round of optimization, then again this is normally all you need. Probably not worth mentioning.

src/zedwallet/Transfer.cpp
brandonlehmann: I think showing them that there is a 0 node fee when not using a node with a fee (aka, probably their own) helps to reinforce the idea of running your own node. Which we definitely need more of.

src/zedwallet/Transfer.h
brandonlehmann: I've push up code to grab the node fee on each split transaction. That being said I'm interested to see how this plays out.

src/zedwallet/Commands.cpp
brandonlehmann:
```c
    if (sendAll)
    {
        if (WalletConfig::defaultMixin != 0 && balance != balanceNoDust)
        {
            uint64_t unsendable = balance - balanceNoDust;

            amount = balanceNoDust - fee - nodeFee;

            std::cout << WarningMsg("Due to dust inputs, we are unable to ")
                      << WarningMsg("send ")
                      << InformationMsg(formatAmount(unsendable))
                      << WarningMsg("of your balance.") << std::endl;

            if (!WalletConfig::mixinZeroDisabled ||
                height < WalletConfig::mixinZeroDisabledHeight)
            {
                std::cout << "Alternatively, you can set the mixin count to "
                          << "zero to send it all." << std::endl;

                if (confirm("Set mixin to 0 so we can send your whole balance? "
                            "This will compromise privacy."))
                {
                    mixin = 0;
                    amount = balance - fee - nodeFee;
                }
            }
            else
            {
                std::cout << "Sorry." << std::endl;
            }
        }
        else
        {
            amount = balance - fee - nodeFee;
        }
    }
```

Yeah, I think it's okay.

ZedPea: Looks good.

src/zedwallet/Transfer.cpp
ZedPea: Good point.

brandonlehmann:
Passing tests.... I think I've answered @ZedPea 's concerns well -- I'm gonna go get my popcorn.

brandonlehmann:
Hold this a little bit. I may have to touch walletgreen for get this to work with the walletd implementation

brandonlehmann:
Go ahead. Whatever I'm doing I'll clean up this afterwards.

PR #386, opened 2018-08-10
After second thought, let's skip 0.6.4 and go straight to 0.7.0

brandonlehmann:
Afterall, we are breaking a few things.


RocksteadyTC:
good idea :)

PR #407, opened 2018-08-16
Compile readline (+ ncurses) statically into zedwallet

ZedPea:
* Needs testing on windows (Shouldn't have readline support, but should build (Possibly can add readline support later but probably will be hard))
* Needs testing on mac (Should have readline support)
* Need to confirm that binaries build on one PC work on another, though ldd suggests they will

This adds readline and ncurses to /external/ and compiles them as part of the build process. This allows us to statically compile them - it seems both readline and ncurses do not have .a files distributed on some distros, so we cannot statically compile without building ourselves.

This adds quite a lot of code - I'm not sure how much we can remove - ncurses seems pretty big. Readline failed to configure when I tried removing the docs directory.

### Downsides
* More code in the repo
* Slightly slower build process (+ 1-2 mins on my machine)
* Will need to manually update code if we want a newer version of readline or ncurses

### Upsides
* Don't need to install readline or ncurses before compiling
* Won't have issues when distributing as dependencies are compiled in


RocksteadyTC:
```Time Elapsed 00:15:12.54
cd src\Release
mkdir turtlecoin-%APPVEYOR_REPO_TAG_NAME%
copy *.exe turtlecoin-%APPVEYOR_REPO_TAG_NAME%
miner.exe
turtle-service.exe
TurtleCoind.exe
zedwallet.exe
        4 file(s) copied.
copy ..\..\LICENSE turtlecoin-%APPVEYOR_REPO_TAG_NAME%
The system cannot find the file specified.
```

ZedPea:
yo this nigga finna get dabbed on /o/

RocksteadyTC:
hahaha

thinkpol2:
I see you are linking against the GNU readline library. Be aware that it is GPL and not LGPL. Including it makes Turtlecoin itself a GPL project. 

[en.wikipedia.org/wiki/GNU_Readline#Choice_of_the_GPL_as_GNU_Readline's_license](https://en.wikipedia.org/wiki/GNU_Readline#Choice_of_the_GPL_as_GNU_Readline's_license)

PR #418, opened 2018-08-18
Add linenoise as smaller alternative to readline

rixombea:
- add linenoise as lighter readline alternative for line editing support (completion, hints, history)
- linenose library taken from this maintained fork: https://github.com/rain-1/linenoise-mob
- add ENABLE_LINENOISE cmake switch, readline can still be enable with ENABLE_READLINE
- completion + history works as it was with readline
- only tested on linux (gcc 8.2.0)

Signed-off-by: rixombea <42405930+rixombea@users.noreply.github.com>


brandonlehmann:
Please see review note.

src/CMakeLists.txt
```diff
@@ -22,7 +23,13 @@ file(GLOB_RECURSE Wallet Wallet/*)
 file(GLOB_RECURSE WalletLegacy WalletLegacy/*)
 file(GLOB_RECURSE JsonRpcServer JsonRpcServer/*)
 file(GLOB_RECURSE service WalletService/*)
+
+if(ENABLE_LINENOISE)
+message(STATUS "linenoise library is enabled")
+file(GLOB_RECURSE zedwallet zedwallet/* ${CMAKE_SOURCE_DIR}/external/linenoise/*)
```
brandonlehmann: Can you link this via the library directives instead? Rather keep these areas cleaner IMO.

CMakeLists.txt
```diff
@@ -14,7 +14,8 @@ set(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build, options are
 set(CMAKE_SKIP_INSTALL_RULES OFF)
 set(CMAKE_SKIP_PACKAGE_ALL_DEPENDENCY ON)
 set(CMAKE_SUPPRESS_REGENERATION ON)
-set(ENABLE_READLINE ON CACHE BOOL "Enable readline support for zedwallet on Linux and Mac? Defaults to ON")
+set(ENABLE_LINENOISE ON CACHE BOOL "Enable line editing support (completion & history) support for zedwallet? Defaults to ON")
+set(ENABLE_READLINE OFF CACHE BOOL "Enable readline support for zedwallet on Linux and Mac? Defaults to OFF")
```
brandonlehmann: This is going to create a merge conflict if #416 is accepted... just FYI

brandonlehmann:
@rixombea this is currently failing CI tests

src/CMakeLists.txt
rixombea: updated

rixombea:
@brandonlehmann 

> Given the option of readline or this I’m not sure which one I prefer but ultimately, I think it’s better to settle on one

That's fine if you prefer to stick with libreadline+ncurses

I'm proposing the switch to lighter libs is to see if we can get required functionality with less dependencies, and if so, maybe we can drop libreadline (unless there are plans to use full-fledged readline+ncurses's features, other than to provide completion/hints & history)

brandonlehmann:
@ZedPea @RocksteadyTC 

What do you guys think? Swap to this instead of readline? Z does this solve your concerns about static/shared readline and what not?

ZedPea:
@brandonlehmann Not fully looked over it but if all the code is included inline it's preferable, especially if there's a lot less of it. It sounds like this gives all the features of readline with less code? If so it's a +1 for me.

My autism requests that the lib code be moved out of the zedwallet folder, I don't want that ugly code in my folder :angry: - Furthermore, it would allow it to be used in other executables, potentially. Possibly in Common/

ZedPea:
Only bit I'm not sure about is the history file. Is this required, or can we just use the previous commands only from this launch? I'm not a big fan of making more files if it's not really needed.

If it's required, or you prefer it this way, I would suggest adding a '.' to the start of the name so it's hidden, and define the filename in WalletConfig.h

Finally, I agree with @brandonlehmann - If this pull request goes ahead, I think we should rip out readline along with it.

rixombea:
history file is only not needed, we can remove it.
command history will still works using non-persistent/in-memory history

rixombea:
@ZedPea persistent history file removed:  17f45b07892cba0308f74856cb635a2d2908cd83

brandonlehmann:
I vote to use this as a full replacement for readline. 

If @rixombea if you wanna rip readline our completely with this PR that’d be great.

ZedPea:
+1

rixombea:
@brandonlehmann, @ZedPea readline removed in 38c3d93b86112e5c79c1505003007e5b77483415, let me know if I'm not doing it correctly.

brandonlehmann:
Will review

ZedPea:
Ok, first things I noticed. When hitting tab with nothing typed, it will start scrolling through commands. I would usually expect to have to double hit tab and then it display all the available commands, as a shell would do.

Secondly, If you type a command, and then hit `ctrl+c` to exit, it will crash the wallet:

```
[TRTL .turtle]: address
TRTLv2Fyavy8CXG8BPEbNeCHFZ1fuDCYCZ3vW5H5LXN4K2M2MHUpTENip9bbavpHvvPwb4NDkBWrNgURAd5DB38FHXWZyoBh4wW
[TRTL .turtle]: 
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
Aborted (core dumped)
```

PR #420, opened 2018-08-19
[ADMINISTRATIVE] Adding jet fuel

brandonlehmann:
Travis docs say that the build machine has 2 cores — might as well use that to our advantage.


SoreGums:
looks like it saves about 8mins, that's pretty sweet for a 40minute job ;)
Nice find

PR #443, opened 2018-08-25
[ENHANCEMENT] Allow scanning from a specific height when importing a wallet

ZedPea:
Adds support to both turtle-service and zedwallet for starting the wallet scanning at a specific height when importing.

If a wallet is created with no parameters, i.e. it is a completely fresh wallet, not imported, it will always be scanned from the timestamp when it was created. There is no way for it to have any transactions previously, so it is pointless to scan those blocks.

When a wallet is created with parameters, such as a mnemonic seed, a view key, or a view and spend key, an optional scan height can be given. This causes the wallet to ignore any blocks before this point, massively speeding up sync times.

It is worth noting that the block hashes still need to be received from the node. However, this process is far faster than the standard sync process. From my tests I found it took roughly 7 minutes and 30 seconds to sync 500,000 blocks from a public node. I assume a local node would be faster.

On the turtle-service side, the createAddress() and createAddressList() calls now have two new *optional* parameters.

These are newAddress (bool), and scanHeight (uint64_t). You cannot specify both at once.

If newAddress is set to true, the wallet will be assumed to be scanned from the current timestamp, with no transactions in the past.

If scanHeight is set to > 0, the wallet will skip blocks before this height when scanning.

If you wish to apply a scanHeight whilst creating a wallet with turtle-service, the config parameter `--scanHeight` can be used for this.

You can also reset from a given height, both in turtle-service, and zedwallet. In turtle-service, this adds the same two optional parameters, to the reset() method, newAddress (bool, and scanHeight (uint64_t).

You cannot specify both at once.
You cannot specify newAddress unless the optional viewSecretKey is specified. (It makes no sense to perform a reset with a timestamp of now, that is not a reset.)

### DIFF
--- a/include/IWallet.h
+++ b/include/IWallet.h
@@ -133,13 +133,14 @@ public:
   virtual ~IWallet() {}
 
   virtual void initialize(const std::string& path, const std::string& password) = 0;
-  virtual void initializeWithViewKey(const std::string& path, const std::string& password, const Crypto::SecretKey& viewSecretKey) = 0;
+  virtual void initializeWithViewKey(const std::string& path, const std::string& password, const Crypto::SecretKey& viewSecretKey, const uint64_t scanHeight, const bool newAddress) = 0;
   virtual void load(const std::string& path, const std::string& password, std::string& extra) = 0;
   virtual void load(const std::string& path, const std::string& password) = 0;
   virtual void shutdown() = 0;
 
   virtual void changePassword(const std::string& oldPassword, const std::string& newPassword) = 0;
   virtual void save(WalletSaveLevel saveLevel = WalletSaveLevel::SAVE_ALL, const std::string& extra = "") = 0;
+  virtual void reset(const uint64_t scanHeight) = 0;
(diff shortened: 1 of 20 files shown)


brandonlehmann:
Will check.

brandonlehmann:
I gave this a quick scan through and it looks good.

One question though, If I go into zedwallet, and type in `reset` will it scan from 0, the known scanHeight start, or ?

Perhaps, in zedwallet, `reset` should take an optional parameter to specify the reset from height? Or does that make no sense?

ZedPea:
Yes, you are right. I forgot to add a section on this. I *think* it will reset from the known scanHeight. Needs to take an optional parameter.

brandonlehmann:
Same would apply for reset in turtle-service

ZedPea:
Updated with reset() method

RocksteadyTC:
Would be nice to follow this up with a sweep function you can run periodically to move all inputs to a new address and associate the keys with the old wallet file to not lose transaction history etc. I mean, it's pretty much a manual stick shift version of mobius at that point but it'd be handy to make that all wrapped up in one function.

RocksteadyTC:
Good job, I'm sure a lot of people will be thankful to not have to sync a bunch of BS

PR #450, opened 2018-08-28
[BUGFIX] Correctly set default parameters

ZedPea:


ZedPea:
What's going on with travis, it's showing as pending to me, but all three builds have completed...

brandonlehmann:
Web hook fail.

PR #451, opened 2018-08-28
Revert Linenoise in Daemon

brandonlehmann:
I've reverted the changes that added linenoise to the daemon. I know it seemed like a good idea but it seems to be causing more trouble than what it's worth.

I'll leave it to others to decide if we want to merge this or not.


rixombea:
@brandonlehmann Sure, I'm fine with it if you think it's unnecessary feature.

btw, you labeled this PR as bugfix, can you tell me what kind of bug causes by those commits? I ask because I want to know and try to reproduce & fix it, at least for my own use.

brandonlehmann:
@rixombea I'll defer to @RocksteadyTC  on that one as I know i haven't run into it myself yet.

RocksteadyTC:
I was noticing what I suspected to be random SIGINT being triggered on daemons left running without HA. Reverting the commit solved that for some odd reason. I might be wrong, and would be happy to add it back if that's the case.

rixombea:
So, causing daemon to randomly exit?

I'm not able to reproduce it, and re-reading the codes, still can't see how is it possible that SIGINT  randomly triggered by this.  A call to raise SIGINT is inside completion callback, which means won't get excecuted without user input. The SIGINT will get triggered only by specific user input (ctrl+c) + some conditions (e.g. only when command line empty) to initiate a graceful exit (In my understanding, SIGINT, SIGTERM & SIGPIPE will get caught by SignalHandler)

Anyway, I agree that it's not essential for the daemon, so +1 to nuke it.

PR #452, opened 2018-08-28
[ENHANCEMENT] Update mixin to static 3, make anonymity parameter optional in sendTransaction() and friends

zpalmtree:
Sets the mixin to a static 3 at block 800,000.

Makes anonymity an optional parameter in turtle-service. If it is not specified, the DEFAULT_MIXIN_V? will be used where ? is based on the current height.

This means only the turtle-service executable will need updating, and not GUI wallets such as nest, to check for a before/after height to set the anonymity param.

I had to modify the serialization a tad to pass through WalletService, and it does appear to serialize correctly, but I wouldn't mind some extra eyes on this.

TODO: api-docs need updating with both reset from height and this new sendTransaction() modifications. Also might want to modify the pool, and nest.

[I'll fix my fucked up git repo after these PR's are done]

### DIFF
--- a/src/CryptoNoteCore/Core.cpp
+++ b/src/CryptoNoteCore/Core.cpp
@@ -958,6 +958,11 @@ bool Core::validateMixin(const std::vector<CachedTransaction> transactions,
     /* We also need to ensure that the mixin enforced is for the limit that
      was correct when the block was formed - i.e. if 0 mixin was allowed at
      block 100, but is no longer allowed - we should still validate block 100 */
+    if (height >= CryptoNote::parameters::MIXIN_LIMITS_V3_HEIGHT)
+    {
+        minMixin = CryptoNote::parameters::MINIMUM_MIXIN_V3;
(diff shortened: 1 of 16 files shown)


src/CryptoNoteCore/Core.cpp
```diff
@@ -958,6 +958,11 @@ bool Core::validateMixin(const std::vector<CachedTransaction> transactions,
     /* We also need to ensure that the mixin enforced is for the limit that
      was correct when the block was formed - i.e. if 0 mixin was allowed at
      block 100, but is no longer allowed - we should still validate block 100 */
+    if (height >= CryptoNote::parameters::MIXIN_LIMITS_V3_HEIGHT)
+    {
+        minMixin = CryptoNote::parameters::MINIMUM_MIXIN_V3;
+        maxMixin = CryptoNote::parameters::MAXIMUM_MIXIN_V3;
+    }
     if (height >= CryptoNote::parameters::MIXIN_LIMITS_V2_HEIGHT)
```
brandonlehmann: Shouldn't this be `else if` ?

PR #453, opened 2018-08-29
[ENHANCEMENT] Adding additional turtle-service method getFeeInfo

brandonlehmann:
See https://github.com/turtlecoin/turtlecoin-walletd-rpc-php/pull/3


zpalmtree:
y u no replace old handler

brandonlehmann:
That would change the API in a non-backwards compatible way and thus require a major version change. This way simply extends it such that it's backwards compatible.

PR #462, opened 2018-09-02
Update encode script

Watt3r:


RocksteadyTC:
shouldn't anonymity come out and autodetect it from the network?

brandonlehmann:
It only becomes an *optional* parameter with 0.8.3-RC if I recall -- until then it must be specified.

zpalmtree:
True, but we've got to update it at some point, and this seems like a good time - the people running this script are probably also running dev

brandonlehmann:
I came the the conclusion that we were discussing this script a lot more than we should be. lol

zpalmtree:
touché

PR #469, opened 2018-09-04
Remove any direct use of boost from zedwallet

zpalmtree:
Also made a few functions const because I can't help myself


SoreGums:
Seeing the tedium required for the basics, has me appreciating the higher level languages, (ruby, php, elixir) even if they are slower. Fast enough is fast enough if I don't have to deal with all this nonsense ;)

Nice work Z

brandonlehmann:
Very nice!

PR #477, opened 2018-09-07
[BUGFIX] Incoming Transactions Block The Event Loop

brandonlehmann:
In looking into creating a different Transaction Mempool to solve some *observed* issues with the handling of a large influx of Transactions, I found this little tidbit.

As we know, it's possible to send a bunch of transactions (fusion for example) to a node without issue; however, as those transactions are relayed around the network it appears that when our local node is notified of new transactions and they are valid that it will wait to continue processing our event loop until it has notified all of our peers of the transaction. This essentially results in blocking the event loop.

What this PR does is attempts to use the same Transaction relay calls (with a little bit of lipstick) as the daemon RPC sendrawtransaction calls. Thereby pushing the sending of the transactions to our peers into the dispatcher... which *should* keep the event loop from blocking.

Please test accordingly... I seriously hope that this brings some much needed additional stability to the daemon across the platforms.

### DIFF
--- a/src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp
+++ b/src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp
@@ -33,7 +33,7 @@ bool post_notify(IP2pEndpoint& p2p, typename t_parametr::request& arg, const Cry
 
 template<class t_parametr>
 void relay_post_notify(IP2pEndpoint& p2p, typename t_parametr::request& arg, const net_connection_id* excludeConnection = nullptr) {
-  p2p.relay_notify_to_all(t_parametr::ID, LevinProtocol::encode(arg), excludeConnection);
+  p2p.externalRelayNotifyToAll(t_parametr::ID, LevinProtocol::encode(arg), excludeConnection);
 }
 
 std::vector<RawBlockLegacy> convertRawBlocksToRawBlocksLegacy(const std::vector<RawBlock>& rawBlocks) {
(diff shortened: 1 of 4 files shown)


src/P2p/NetNode.cpp
```diff
@@ -318,9 +318,9 @@ std::string print_peerlist_to_string(const std::list<PeerlistEntry>& pl) {
   }
 
   //-----------------------------------------------------------------------------------
-  void NodeServer::externalRelayNotifyToAll(int command, const BinaryArray& data_buff) {
-    m_dispatcher.remoteSpawn([this, command, data_buff] {
-      relay_notify_to_all(command, data_buff, nullptr);
+  void NodeServer::externalRelayNotifyToAll(int command, const BinaryArray& data_buff, const net_connection_id* excludeConnection) {
```
zpalmtree: How come we're adding an extra parameter here which appears to be only called with nullptr?

src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp
```diff
@@ -33,7 +33,7 @@ bool post_notify(IP2pEndpoint& p2p, typename t_parametr::request& arg, const Cry
 
 template<class t_parametr>
 void relay_post_notify(IP2pEndpoint& p2p, typename t_parametr::request& arg, const net_connection_id* excludeConnection = nullptr) {
-  p2p.relay_notify_to_all(t_parametr::ID, LevinProtocol::encode(arg), excludeConnection);
+  p2p.externalRelayNotifyToAll(t_parametr::ID, LevinProtocol::encode(arg), excludeConnection);
```
brandonlehmann: @zpalmtree  this... this is calling it with something other than nullptr because we don't actually relay transactions heard via p2p to the same node that we heard them from

Versus the RPC calls, which require that we don't exclude any peers.

To avoid confusion, I decided to make both paths into relaying transactions the same thing to you know... provided a *common* entry point.

src/P2p/NetNode.cpp
brandonlehmann: See https://github.com/turtlecoin/turtlecoin/pull/477#pullrequestreview-153501591

src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp
zpalmtree: Ok, I misread the default argument

zpalmtree:
Happy to merge?

brandonlehmann:
Sure, why not?

PR #484, opened 2018-09-11
Tidying up the RPC

brandonlehmann:
Quick clean up in the daemon RPC server and fixing an implicit conversion warning of uint64_t to uint32_t.


brandonlehmann:
So weird. Travis has the builds for the PRs... but they aren't showing here :(

PR #486, opened 2018-09-11
[ENHANCEMENT] P2P Version Grace Period Configurable

brandonlehmann:
Some people *cough* @zpalmtree make good arguments to support this.

### DIFF
--- a/src/P2p/NetNode.cpp
+++ b/src/P2p/NetNode.cpp
@@ -582,7 +582,7 @@ std::string print_peerlist_to_string(const std::list<PeerlistEntry>& pl) {
     if (rsp.node_data.version < CryptoNote::P2P_MINIMUM_VERSION) {
       logger(Logging::DEBUGGING) << context << "COMMAND_HANDSHAKE Failed, peer is wrong version! (" << std::to_string(rsp.node_data.version) << "), closing connection.";
       return false;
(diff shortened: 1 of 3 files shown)


src/P2p/NetNode.cpp
```diff
@@ -582,7 +582,7 @@ std::string print_peerlist_to_string(const std::list<PeerlistEntry>& pl) {
     if (rsp.node_data.version < CryptoNote::P2P_MINIMUM_VERSION) {
       logger(Logging::DEBUGGING) << context << "COMMAND_HANDSHAKE Failed, peer is wrong version! (" << std::to_string(rsp.node_data.version) << "), closing connection.";
       return false;
-    } else if (rsp.node_data.version > CryptoNote::P2P_CURRENT_VERSION) {
+    } else if ((rsp.node_data.version - CryptoNote::P2P_CURRENT_VERSION) > CryptoNote::P2P_UPGRADE_WINDOW) {
```
zpalmtree: Shouldn't this be a `>=`?

brandonlehmann: Yes, yes it should. as we actually need to warn inclusive of 2

brandonlehmann: Pushed update.

PR #487, opened 2018-09-12
[ENHANCEMENT] Add daemon local height to wallet getStatus()

zpalmtree:
I've tested with a synced node, but when I tried to test with an unsynced node, it was just returning 1. I think it was the daemon playing up though, since zedwallet did the same.


woodyjon:
Works well! Thanks Z

woodyjon:
Same as you, it's working well with a synced node (or even when almost synced), but it returns 1 when I tried to resync from scratch. And everytime I relaunch the sync, localDaemonBlockCount is still 1.

PR #488, opened 2018-09-12
[ENHANCEMENT] Makes NodeRpcProxy actually shutdown when you tell it to

zpalmtree:


src/NodeRpcProxy/NodeRpcProxy.cpp
```diff
@@ -104,29 +104,11 @@ void NodeRpcProxy::init(const INode::Callback& callback) {
...
-
-  m_dispatcher->remoteSpawn([this]() {
-    m_stop = true;
-    // Run all spawned contexts
-    m_dispatcher->yield();
-  });
-
   if (m_workerThread.joinable()) {
-    m_workerThread.join();
+    /* FUCK OFF */
```
brandonlehmann: That needs to go.

zpalmtree: Fixed

PR #490, opened 2018-09-15
RingCT

Amity-Core:


Amity-Core:
Honestly thought the RingCT folder uploaded with a commit but it looks like it never though, I'll close this and PR when I've actually implemented it

Issue #491, opened 2018-09-15
Mixin Soft Fork Parameters


brandonlehmann:
In the future, we need to make sure that the addBlock() method in Core.cpp allows OLD mixin values for the mined money unlock window. However, the addTransactionToPool must remain at the NEW level.

PR #493, opened 2018-09-15
[BUGFIX] Core.addBlock() to accept transactions with older mixin values for a period of time.

brandonlehmann:
Hopefully addresses #491

### DIFF
--- a/src/CryptoNoteCore/Core.cpp
+++ b/src/CryptoNoteCore/Core.cpp
@@ -594,7 +594,7 @@ std::error_code Core::addBlock(const CachedBlock& cachedBlock, RawBlock&& rawBlo
   uint32_t mixinChangeWindow = blockIndex;
   if (mixinChangeWindow > CryptoNote::parameters::CRYPTONOTE_MINED_MONEY_UNLOCK_WINDOW)
   {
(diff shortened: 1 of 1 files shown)


src/CryptoNoteCore/Core.cpp
```diff
@@ -588,7 +588,11 @@ std::error_code Core::addBlock(const CachedBlock& cachedBlock, RawBlock&& rawBlo
     return error::BlockValidationError::DIFFICULTY_OVERHEAD;
   }
 
-  if (!validateMixin(transactions, blockIndex))
+  // This allows us to accept blocks with transaction mixins for the mined money unlock window
+  // that may be using older mixin rules on the network. This helps to clear out the transaction
+  // pool during a network soft fork that requires a mixin lower or upper bound change
+  uint32_t mixinChangeWindow = blockIndex - CryptoNote::parameters::CRYPTONOTE_MINED_MONEY_UNLOCK_WINDOW;
```
zpalmtree: This could be negative if we're at the start of a chain.

brandonlehmann: *sigh*

brandonlehmann: That work?

zpalmtree: Yep

brandonlehmann:
Forgot a `;`

Issue #492, opened 2018-09-15
Transaction Pool Cleaner


brandonlehmann:
https://github.com/turtlecoin/turtlecoin/blob/8c8641c2c249120563c06dc5b646a6e3f9beeed1/src/CryptoNoteCore/TransactionPoolCleaner.cpp#L84

This needs turned up a notch to validate the mixin of transactions and clear out those that don't match the current network state.

RocksteadyTC:
![image](https://user-images.githubusercontent.com/34389545/45589001-b3e4bf80-b8e3-11e8-8e83-ad9abc6b83b9.png)

zpalmtree:
haha, what a funny image. Can I save it?

brandonlehmann:
Look. It’s @fexra

brandonlehmann:
PR is ready for this as #495 -- awaiting CI test completion.

PR #495, opened 2018-09-15
[ENHANCEMENT] Mixin Validation Enhancement & Transaction Pool Cleaner Update

brandonlehmann:
Working towards fixing #492 

Yes... I know this doesn't build as is. And yes, it's because of what I added in `src/CryptoNoteCore/TransactionPoolCleaner.cpp` Line 92.

### DIFF
--- a/src/CryptoNoteCore/Core.cpp
+++ b/src/CryptoNoteCore/Core.cpp
@@ -1,7 +1,7 @@
 // Copyright (c) 2012-2017, The CryptoNote developers, The Bytecoin developers
 // Copyright (c) 2014-2018, The Monero Project
 // Copyright (c) 2018, The TurtleCoin Developers
-// 
+//
 // Please see the included LICENSE file for more information.
 
 #include <algorithm>
@@ -25,6 +25,7 @@
 #include "CryptoNoteCore/TransactionPool.h"
 #include "CryptoNoteCore/TransactionPoolCleaner.h"
 #include "CryptoNoteCore/UpgradeManager.h"
+#include "CryptoNoteCore/Mixins.h"
 #include "CryptoNoteProtocol/CryptoNoteProtocolHandlerCommon.h"
 
 #include <System/Timer.h>
@@ -442,7 +443,7 @@ bool Core::queryBlocksLite(const std::vector<Crypto::Hash>& knownBlockHashes, ui
 
     return true;
   } catch (std::exception& e) {
-	logger(Logging::ERROR) << "Failed to query blocks: " << e.what();
+    logger(Logging::ERROR) << "Failed to query blocks: " << e.what();
     return false;
   }
 }
@@ -597,9 +598,19 @@ std::error_code Core::addBlock(const CachedBlock& cachedBlock, RawBlock&& rawBlo
     mixinChangeWindow = mixinChangeWindow - CryptoNote::parameters::CRYPTONOTE_MINED_MONEY_UNLOCK_WINDOW;
   }
 
-  if (!validateMixin(transactions, blockIndex) && !validateMixin(transactions, mixinChangeWindow))
+  bool success;
+  std::string error;
+  std::tie(success, error) = Mixins::validate(transactions, blockIndex);
+
+  if (!success)
   {
+    std::tie(success, error) = Mixins::validate(transactions, mixinChangeWindow);
+
+    if (!success)
+    {
+      logger(Logging::DEBUGGING) << error;
       return error::TransactionValidationError::INVALID_MIXIN;
+    }
   }
 
   uint64_t cumulativeFee = 0;
@@ -688,7 +699,7 @@ std::error_code Core::addBlock(const CachedBlock& cachedBlock, RawBlock&& rawBlo
           ret = error::AddBlockErrorCode::ADDED_TO_ALTERNATIVE_AND_SWITCHED;
 
           logger(Logging::INFO) << "Resolved: " << blockStr
-                                << ", Previous: " << chainsLeaves[endpointIndex]->getTopBlockIndex() << " (" 
+                                << ", Previous: " << chainsLeaves[endpointIndex]->getTopBlockIndex() << " ("
                                 << chainsLeaves[endpointIndex]->getTopBlockHash() << ")";
         }
       }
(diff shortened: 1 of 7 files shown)


src/CryptoNoteCore/TransactionPoolCleaner.cpp
```diff
@@ -95,6 +84,18 @@ std::vector<Crypto::Hash> TransactionPoolCleanWrapper::clean() {
...
         deletedTransactions.emplace_back(std::move(hash));
       }
+
+      CachedTransaction transaction = transactionPool->getTransaction(hash);
+      std::vector<CachedTransaction> transactions;
+      transactions.emplace_back(transaction);
+
+      if (!Core::validateMixin(transactions, height))
+      {
+        logger(Logging::DEBUGGING) << "Deleting invlid transaction " << Common::podToHex(hash) << " from pool";
```
zpalmtree: typo ;o

src/CryptoNoteCore/TransactionPoolCleaner.cpp
```diff
@@ -95,6 +84,18 @@ std::vector<Crypto::Hash> TransactionPoolCleanWrapper::clean() {
         transactionPool->removeTransaction(hash);
         deletedTransactions.emplace_back(std::move(hash));
       }
+
+      CachedTransaction transaction = transactionPool->getTransaction(hash);
+      std::vector<CachedTransaction> transactions;
+      transactions.emplace_back(transaction);
```
zpalmtree: Why are we adding the transactions to a vector here? Part of the WIP?

brandonlehmann: Because validatemixin wants a vector ?

brandonlehmann: And this pulls them out one at a time to check them...

src/CryptoNoteCore/TransactionPoolCleaner.cpp
brandonlehmann: Did you fix it ? Lol

src/CryptoNoteCore/TransactionPoolCleaner.cpp
zpalmtree: Sorry, I'm dumb

src/CryptoNoteCore/TransactionPoolCleaner.cpp
zpalmtree: >invlid

brandonlehmann: Well they are invalid. :)

zpalmtree: :woman_facepalming: ok i laughed

brandonlehmann:
I may pull validate mixins our into a static method and throw logging around those attempts to use it. That logger isn’t the only thing though that makes it not static I think.

brandonlehmann:
Maybe also pull it own on its own entirely.

zpalmtree:
You can just pass in the logger object, I remember doing this for some hacky mnemonic code

brandonlehmann:
Did you say hacky? How is this a good thing?

zpalmtree:
just saying, it works

brandonlehmann:
Will investigate.

src/CryptoNoteCore/TransactionPoolCleaner.cpp
brandonlehmann: Fixed

brandonlehmann:
Rebased on top of the current development branch.

Pulled validateMixin out on it's own, made it static, now returning a tuple <bool, std::string> let's see if that does the trick.

src/CryptoNoteCore/Mixins.h
```diff
@@ -0,0 +1,102 @@
...
+// Please see the included LICENSE file for more information.
+
+#pragma once
+#include "CachedTransaction.h"
+#include "TransactionApi.h"
+#include <config/CryptoNoteConfig.h>
+
+namespace CryptoNote
+{
+  class Mixins
```
zpalmtree: I don't see much point in making this a static class instead of just free functions in a namespace - You can make the namespace `Mixins` instead if you want to make it obvious where the functions come from, and then use `Mixins::validate()` instead of `Mixins.validate()`.

brandonlehmann:
Fixed the error I caused in resolving merge conflicts.

src/CryptoNoteCore/Core.cpp
```diff
@@ -597,9 +598,18 @@ std::error_code Core::addBlock(const CachedBlock& cachedBlock, RawBlock&& rawBlo
...
+  std::tie(resa, erra) = Mixins.validate(transactions, blockIndex);
+
+  bool resb;
+  std::string errb;
+  std::tie(resb, errb) = Mixins.validate(transactions, mixinChangeWindow);
+
+  if (!resa && !resb)
   {
-      return error::TransactionValidationError::INVALID_MIXIN;
+    logger(Logging::DEBUGGING) << err;
```
zpalmtree: Where does `err` come from, shouldn't it be erra or errb? (Not a big fan of those variables names ;x) - Why not do something like

```
bool success;
std::string error;

std::tie(success, error) = Mixins.validate(transactions, blockIndex);

if (!success)
{
    std::tie(success, error) = Mixins.validate(transactions, mixinChangeWindow);

    if (!success)
    {
        logging(Logging::DEBUGGING) << error;
        return error::TransactionValidationError::INVALID_MIXIN;
    }
}

File: src/CryptoNoteCore/Mixins.h [hunk 5]
```diff
@@ -0,0 +1,102 @@
...
+            std::string err;
+
+            std::tie(res, err) = validate(transaction, minMixin, maxMixin);
+            if (!res)
+            {
+                return std::make_tuple(res, err);
+            }
+        }
+
+        return std::make_tuple(true, nullptr);
```
zpalmtree: Returning nullptr as a string is undefined behaviour. Use `std::string()` instead.

src/CryptoNoteCore/Mixins.h
brandonlehmann: Personally, I like to know where stuff came from as it avoids confusion later and stops namespace pollution.

src/CryptoNoteCore/Mixins.h
brandonlehmann: Fixing...

src/CryptoNoteCore/Core.cpp
brandonlehmann: That makes more sense. Updating.

brandonlehmann:
There we go, I think that'll do it. Do I get bonus points for making it header only? 😄

src/CryptoNoteCore/Mixins.h
```diff
@@ -0,0 +1,145 @@
...
+
+        for (const auto& transaction : transactions)
+        {
+            bool success;
+            std::string error;
+
+            std::tie(success, error) = validate(transaction, minMixin, maxMixin);
+            if (!success)
+            {
+                return std::make_tuple(success, error);
```
zpalmtree: Might be better to use `false` here - returning success is a tad confusing at a quick glance

brandonlehmann: Everyone's a critic

brandonlehmann:
#yolo

PR #521, opened 2018-09-28
[BUGFIX] Resolving Wallet Coin Burn Bug

brandonlehmann:
Borrowed from https://github.com/seredat/karbowanec/commit/137f3d9d5693c00e5045a6cd2a25cd42f804c715

Working on testing -- initiating pull for others to test as well.

### DIFF
--- a/src/Transfers/TransfersConsumer.cpp
+++ b/src/Transfers/TransfersConsumer.cpp
@@ -472,11 +472,6 @@ std::error_code createTransfers(
           throw std::runtime_error("duplicate transaction output key is found");
         }
 
-        if (std::find(temp_keys.begin(), temp_keys.end(), out.key) != temp_keys.end())
-        {
-          throw std::runtime_error("duplicate transaction output key is found");
-        }
-
         temp_keys.push_back(out.key);
       }
 
@@ -487,6 +482,7 @@ std::error_code createTransfers(
     transfers.push_back(info);
   }
 
+  transactions_hash_seen.insert(tx.getTransactionHash());
   std::copy(temp_keys.begin(), temp_keys.end(), std::inserter(public_keys_seen, public_keys_seen.end()));
   temp_keys.clear();
   temp_keys.shrink_to_fit();
--- a/src/Transfers/TransfersConsumer.cpp
+++ b/src/Transfers/TransfersConsumer.cpp
@@ -493,9 +493,6 @@ std::error_code createTransfers(
 
   transactions_hash_seen.insert(tx.getTransactionHash());
   std::copy(temp_keys.begin(), temp_keys.end(), std::inserter(public_keys_seen, public_keys_seen.end()));
-  temp_keys.clear();
-  temp_keys.shrink_to_fit();
-
   return std::error_code();
 }
 
--- a/src/Transfers/TransfersConsumer.cpp
+++ b/src/Transfers/TransfersConsumer.cpp
@@ -420,7 +420,8 @@ std::error_code createTransfers(
   const ITransactionReader& tx,
   const std::vector<uint32_t>& outputs,
   const std::vector<uint32_t>& globalIdxs,
-  std::vector<TransactionOutputInformationIn>& transfers)
+  std::vector<TransactionOutputInformationIn>& transfers,
+  Logging::LoggerRef& m_logger)
 {
 
   auto txPubKey = tx.getTransactionPublicKey();
@@ -429,6 +430,8 @@ std::error_code createTransfers(
 
   for (auto idx : outputs)
   {
+    bool isDuplicate = false;
+
     if (idx >= tx.getOutputCount())
     {
       return std::make_error_code(std::errc::argument_out_of_domain);
@@ -469,17 +472,23 @@ std::error_code createTransfers(
       {
         if (public_keys_seen.find(out.key) != public_keys_seen.end())
         {
-          throw std::runtime_error("duplicate transaction output key is found");
+          m_logger(WARNING, BRIGHT_RED) << "A duplicate public key was found in " << Common::podToHex(tx.getTransactionHash());
+          isDuplicate = true;
         }
-
-        temp_keys.push_back(out.key);
+        else
+        {
+          temp_keys.push_back(out.key);
+        }
       }
 
       info.amount = amount;
       info.outputKey = out.key;
     }
 
-    transfers.push_back(info);
+    if (!isDuplicate)
+    {
+      transfers.push_back(info);
+    }
   }
 
   transactions_hash_seen.insert(tx.getTransactionHash());
@@ -522,18 +531,10 @@ std::error_code TransfersConsumer::preprocessOutputs(const TransactionBlockInfo&
     auto it = m_subscriptions.find(kv.first);
     if (it != m_subscriptions.end()) {
       auto& transfers = info.outputs[kv.first];
-      try
+      errorCode = createTransfers(it->second->getKeys(), blockInfo, tx, kv.second, info.globalIdxs, transfers, m_logger);
+      if (errorCode)
       {
-          errorCode = createTransfers(it->second->getKeys(), blockInfo, tx, kv.second, info.globalIdxs, transfers);
-          if (errorCode)
-          {
-            return errorCode;
-          }
-      }
-      catch (const std::exception& e)
-      {
-        m_logger(WARNING, BRIGHT_RED) << "Failed to process transaction: " << e.what() << ", transaction hash " << Common::podToHex(tx.getTransactionHash());
-        return std::error_code();
+        return errorCode;
       }
     }
   }
--- a/src/Transfers/TransfersConsumer.cpp
+++ b/src/Transfers/TransfersConsumer.cpp
@@ -1,23 +1,14 @@
 // Copyright (c) 2012-2017, The CryptoNote developers, The Bytecoin developers
+// Copyright (c) 2018, The BBSCoin Developers
+// Copyright (c) 2018, The Karbo Developers
+// Copyright (c) 2018, The TurtleCoin Developers
 //
-// This file is part of Bytecoin.
-//
-// Bytecoin is free software: you can redistribute it and/or modify
-// it under the terms of the GNU Lesser General Public License as published by
-// the Free Software Foundation, either version 3 of the License, or
-// (at your option) any later version.
-//
-// Bytecoin is distributed in the hope that it will be useful,
-// but WITHOUT ANY WARRANTY; without even the implied warranty of
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-// GNU Lesser General Public License for more details.
-//
-// You should have received a copy of the GNU Lesser General Public License
-// along with Bytecoin.  If not, see <http://www.gnu.org/licenses/>.
+// Please see the included LICENSE file for more information.
 
 #include "TransfersConsumer.h"
 
 #include <numeric>
+#include <future>
 
 #include "CommonTypes.h"
 #include "Common/BlockingQueue.h"
@@ -31,6 +22,12 @@
 
 using namespace Crypto;
 using namespace Logging;
+using namespace Common;
+
+
+std::unordered_set<Crypto::Hash> transactions_hash_seen;
+std::unordered_set<Crypto::PublicKey> public_keys_seen;
+std::mutex seen_mutex;
 
 namespace {
 
@@ -360,7 +357,7 @@ uint32_t TransfersConsumer::onNewBlocks(const CompleteBlock* blocks, uint32_t st
 
 std::error_code TransfersConsumer::onPoolUpdated(const std::vector<std::unique_ptr<ITransactionReader>>& addedTransactions, const std::vector<Hash>& deletedTransactions) {
   TransactionBlockInfo unconfirmedBlockInfo;
-  unconfirmedBlockInfo.timestamp = 0; 
+  unconfirmedBlockInfo.timestamp = 0;
   unconfirmedBlockInfo.height = WALLET_UNCONFIRMED_TRANSACTION_HEIGHT;
 
   std::error_code processingError;
@@ -375,7 +372,7 @@ std::error_code TransfersConsumer::onPoolUpdated(const std::vector<std::unique_p
       return processingError;
     }
   }
-  
+
   for (auto& deletedTxHash : deletedTransactions) {
     m_poolTxs.erase(deletedTxHash);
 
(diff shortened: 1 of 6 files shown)


brandonlehmann:
Hold

brandonlehmann:
Holding. New code coming. Do not merge this yet.

brandonlehmann:
REBASE to HEAD. Working on finishing the work that needs done here now.

brandonlehmann:
Borrowed from:

bbscoin/bbscoin@15a39df
seredat/karbowanec@24f9ad9

Credit to @hatoyakohane and @aivve for their hard work on this

brandonlehmann:
@RocksteadyTC and @zpalmtree  I know you guys approved this when it was a WIP but I've pulled in additional changes to make it more complete. Please review again and thumbs up if you're good :) I'd like to make sure we have enough eyes on this since you know... wallets

src/Transfers/TransfersConsumer.cpp
```diff
@@ -457,6 +463,15 @@ std::error_code createTransfers(
 
       assert(out.key == reinterpret_cast<const PublicKey&>(in_ephemeral.publicKey));
 
+      std::unordered_set<Crypto::Hash>::iterator it = transactions_hash_seen.find(tx.getTransactionHash());
+	  if (it == transactions_hash_seen.end()) {
+        std::unordered_set<Crypto::PublicKey>::iterator key_it = public_keys_seen.find(out.key);
+        if (key_it != public_keys_seen.end()) {
+          throw std::runtime_error("duplicate transaction output key is found");
+          return std::error_code();
```
zpalmtree: We won't ever get to the line I don't think.

src/Transfers/TransfersConsumer.cpp
```diff
@@ -457,6 +463,15 @@ std::error_code createTransfers(
 
       assert(out.key == reinterpret_cast<const PublicKey&>(in_ephemeral.publicKey));
 
+      std::unordered_set<Crypto::Hash>::iterator it = transactions_hash_seen.find(tx.getTransactionHash());
+	  if (it == transactions_hash_seen.end()) {
+        std::unordered_set<Crypto::PublicKey>::iterator key_it = public_keys_seen.find(out.key);
```
zpalmtree: Will we not miss a duplicate key which is in the same transaction here, since we only check public_keys_seen and not temp_keys? I expect that will be rejected by the daemon though.

src/Transfers/TransfersConsumer.cpp
```diff
@@ -465,12 +480,26 @@ std::error_code createTransfers(
     transfers.push_back(info);
   }
 
+  transactions_hash_seen.insert(tx.getTransactionHash());
+  for (std::vector<PublicKey>::iterator it = temp_keys.begin(); it != temp_keys.end(); it++) {
```
zpalmtree: We could just do `std::copy(temp_keys.begin(), temp_keys.end(), std::inserter(public_keys_seen, public_keys_seen.end())` here I think.

src/Transfers/TransfersConsumer.cpp
```diff
@@ -457,6 +463,15 @@ std::error_code createTransfers(
 
       assert(out.key == reinterpret_cast<const PublicKey&>(in_ephemeral.publicKey));
 
+      std::unordered_set<Crypto::Hash>::iterator it = transactions_hash_seen.find(tx.getTransactionHash());
```
zpalmtree: I personally would use `const auto it` here, and the other places we use iterators, but that's just style preference.

src/Transfers/TransfersConsumer.cpp
```diff
@@ -489,10 +518,16 @@ std::error_code TransfersConsumer::preprocessOutputs(const TransactionBlockInfo&
...
-      }
+      try {
+          errorCode = createTransfers(it->second->getKeys(), blockInfo, tx, kv.second, info.globalIdxs, transfers);
+          if (errorCode) {
+            return errorCode;
+          }
+	  }
+	  catch (const std::exception& e) {
+		  m_logger(ERROR, BRIGHT_RED) << "Failed to process transaction: " << e.what() << ", transaction hash " << Common::podToHex(tx.getTransactionHash());
+		  return std::error_code();
```
zpalmtree: Should we be returning an error code here? Or does that stop the wallet syncing?

src/Transfers/TransfersConsumer.cpp
brandonlehmann: Agreed. Updating.

src/Transfers/TransfersConsumer.cpp
brandonlehmann: Agreed. Replacing with

```c
std::copy(temp_keys.begin(), temp_keys.end(), std::inserter(public_keys_seen, public_keys_seen.end()));
```

src/Transfers/TransfersConsumer.cpp
brandonlehmann: It's my understanding that returning an actual error will result in the wallet not syncing. Which makes sense -- as we actually need it to keep processing and keep going despite the event occurring otherwise you'd have to burn an entire wallet if someone sent this against your account. I'm not sure that it should be an ERROR though. Maybe a warning? Thoughts?

src/Transfers/TransfersConsumer.cpp
brandonlehmann: I think you're right.

brandonlehmann: New block

```c
const auto it = transactions_hash_seen.find(tx.getTransactionHash());
      if (it == transactions_hash_seen.end())
      {
        const auto key_it = public_keys_seen.find(out.key);
        if (key_it != public_keys_seen.end())
        {
          throw std::runtime_error("duplicate transaction output key is found");
          return std::error_code();
        }
        
        const auto temp_key_it = temp_keys.find(out.key);
        if (temp_key_it != temp_keys.end())
        {
          throw std::runtime_error("duplicate transaction output key is found");
          return std::error_code();
        }

        temp_keys.push_back(out.key);
      }
```

zpalmtree: You could inline this - `if (public_keys_seen.find(out.key) != public_keys_seen.end()) { }`

src/Transfers/TransfersConsumer.cpp
brandonlehmann: Please elaborate.

zpalmtree: We throw an exception on the line before it.

src/Transfers/TransfersConsumer.cpp
brandonlehmann: Okay then... 

```c
if (transactions_hash_seen.find(tx.getTransactionHash()) == transactions_hash_seen.end())
      {
        if (public_keys_seen.find(out.key) != public_keys_seen.end())
        {
          throw std::runtime_error("duplicate transaction output key is found");
          return std::error_code();
        }
        
        if (temp_keys.find(out.key) != temp_keys.end())
        {
          throw std::runtime_error("duplicate transaction output key is found");
          return std::error_code();
        }

        temp_keys.push_back(out.key);
      }
```

src/Transfers/TransfersConsumer.cpp
brandonlehmann: You mean the assert?

```c
assert(out.key == reinterpret_cast<const PublicKey&>(in_ephemeral.publicKey));
```

??

zpalmtree: Are we seeing different bits of code?

```
throw std::runtime_error("duplicate transaction output key is found"); // throws an exception
return std::error_code(); // never get to this line
```

brandonlehmann: 🤦‍♂️

brandonlehmann: Well... the throw occurs and the rest of the code is written for that so....

```c
if (transactions_hash_seen.find(tx.getTransactionHash()) == transactions_hash_seen.end())
      {
        if (public_keys_seen.find(out.key) != public_keys_seen.end())
        {
          throw std::runtime_error("duplicate transaction output key is found");
        }

        if (std::find(temp_keys.begin(), temp_keys.end(), out.key) != temp_keys.end())
        {
          throw std::runtime_error("duplicate transaction output key is found");
        }

        temp_keys.push_back(out.key);
      }
```

brandonlehmann:
Pushed new commit https://github.com/turtlecoin/turtlecoin/pull/521/commits/a2f1273a726a5f62dd4ffdf8e5ed722afa596267

src/Transfers/TransfersConsumer.cpp
hatoyakohane: There is a output_index in the output key when generate it. So the public key in same transaction can't be same

src/Transfers/TransfersConsumer.cpp
```diff
@@ -457,30 +465,59 @@ std::error_code createTransfers(
...
+
       info.amount = amount;
       info.outputKey = out.key;
-
     }
 
     transfers.push_back(info);
   }
 
+  std::copy(temp_keys.begin(), temp_keys.end(), std::inserter(public_keys_seen, public_keys_seen.end()));
```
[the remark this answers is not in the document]
aivve: In this place we have `transactions_hash_seen.emplace(tx.getTransactionHash());`

aivve:
We `return std::error_code();` in `createTransfers` function so transaction with duped output key is not processed and credited to balance.

src/Transfers/TransfersConsumer.cpp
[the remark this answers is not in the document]
brandonlehmann: Did I chop too many lines? I think so. Sorry. Good catch.

src/Transfers/TransfersConsumer.cpp
brandonlehmann: I've taken that small block back out. Thanks @hatoyakohane

brandonlehmann:
Added additional labels. 

I made some slight changes to the control flow that I *think* accomplishes the same thing.

src/Transfers/TransfersConsumer.cpp
```diff
@@ -457,30 +468,61 @@ std::error_code createTransfers(
...
+    if (!isDuplicate)
+    {
+      transfers.push_back(info);
+    }
   }
 
+  transactions_hash_seen.insert(tx.getTransactionHash());
+  std::copy(temp_keys.begin(), temp_keys.end(), std::inserter(public_keys_seen, public_keys_seen.end()));
+  temp_keys.clear();
+  temp_keys.shrink_to_fit();
```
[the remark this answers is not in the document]
aivve: This will allow to check for double counting, i.e. duplicate output key within one transaction. I doubt it's possible but can be left just in case.

brandonlehmann:
> > > We `return std::error_code();` in `createTransfers` function so transaction with duped output key is not processed and credited to balance.
> > 
> > 
> > @aivve but you `throw` just before that.... `throw` passes control to the exception handler (outside of `createTransfers`) and the code never makes it to the `return`... or is there some voodoo magic here we're missing?
> > Or, is the intent to `return std::error_code()` and not `throw`?
> > That's the confusion here.
> 
> You are right, I made throw because it was inline function with no logger access. It's possible to just log error message and return error_code without throw

Thanks! The last commit I pushed passes in m_logger to allow logging in the method.

Thoughts on not returning right away? I’m concerned that it’s possible to have a situation like valid, dupe, valid that would result in missing the third that is “okay” if we return early. I *think* simply logging a dupe and ignoring it is all we need to do.

src/Transfers/TransfersConsumer.cpp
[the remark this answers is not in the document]
brandonlehmann: @aivve I believe @zpalmtree is referring to the .clear() and .shrink_to_fit() as temp_keys is destroyed when the method returns one line below.

brandonlehmann:
> > Thoughts on not returning right away? I’m concerned that it’s possible to have a situation like valid, dupe, valid that would result in missing the third that is “okay” if we return early. I _think_ simply logging a dupe and ignoring it is all we need to do.
> 
> We discard the entire transaction if there are duped outputs, it's not listed among incoming transfers, its amount is not added to the wallet balance. Maybe I will come up with a better solution, we couldn't linger anymore.

Understood. I *think* what I did in the last commit still allows the transaction but blocks the duplicate.

What’s the best way to reproduce the error in the first place?

src/Transfers/TransfersConsumer.cpp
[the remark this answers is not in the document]
brandonlehmann: Updated.

src/Transfers/TransfersConsumer.cpp
```diff
@@ -457,30 +468,61 @@ std::error_code createTransfers(
 
       assert(out.key == reinterpret_cast<const PublicKey&>(in_ephemeral.publicKey));
 
+      if (transactions_hash_seen.find(tx.getTransactionHash()) == transactions_hash_seen.end())
```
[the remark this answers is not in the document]
brandonlehmann: The network will reject a transaction hash that matches another transaction hash so isn't it plausible that you'd likely never scan a transaction that you've already scanned anyways? Which begs the question... how is this handled during a reset of a wallet? I think we need to make sure that a reset clears those lists as well.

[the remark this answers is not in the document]
zpalmtree: Hmm... is it because you might have already scanned a transaction, then the mempool updates or something, and you see it again? And, since it has the same hash, it's the same transaction, so it's not a dupe?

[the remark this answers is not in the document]
brandonlehmann: Plausible as we know the wallet scans the mempool as well.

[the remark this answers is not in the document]
zpalmtree: That makes sense then. However, it would potentially mean there are multiple of the same transaction in the transfers vector? But, I assume that's not worth mentioning since this is a new addition, and syncing works fine as is.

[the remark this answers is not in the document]
brandonlehmann: Agreed. I'd say resolve this for now, concur?

brandonlehmann:
I've been unable to replicate the bug under ANY circumstances even with the advice that @aivve provided to do so.

I'm at a loss on what to do with this ☹️

brandonlehmann:
To be very clear, that's attempting to replicate the bug on the (a) current development build, (b) with the initial changes suggested, and (c) with the edits between @zpalmtree and myself.

brandonlehmann:
#yolo

Issue #496, opened 2018-09-16
0.8.3 Windows Daemon Crashes on Ivy Bridge CPU


dschroth:
### System Information

**Operating System**: Windows
**OS Version**: 7
**CPU Make, Model, & Speed**: VM running on E5-2695 v2
**Amount of RAM**: 16GB
**Hard Drive Type**: Drive Storage is on platter based RAID 10
**Hard Drive Size**: 250GB
**Volume Type**: NTFS

### TurtleCoin Build Information

**Version:** v0.8.3.1275
**Binary Source:** Used both 0.8.3 windows binaries AND compiled from source
**If self compiled, please provide all the build commands you used**
``` Followed 0.8.3 build instructions for Windows 10
Please insert your build commands here. :

Use the start menu or similar to open 'x64 Native Tools Command Prompt for vs2017' command prompt.
cd <your_turtlecoin_directory>
mkdir build
cd build
cmake -G "Visual Studio 14 Win64" .. -DBOOST_ROOT=D:/Boost/boost_1_59_0 (Or your boost installed dir.)
MSBuild TurtleCoin.sln /p:Configuration=Release
```

**If self compiled, please provide of your TurtleCoin CMake output at the default verbosity**:
```

Please insert your default verbosity CMake output here
```
[CMakeOutput.log](https://github.com/turtlecoin/turtlecoin/files/2386306/CMakeOutput.log)
[CMakeCache.txt](https://github.com/turtlecoin/turtlecoin/files/2386307/CMakeCache.txt)
### Issue Report

**Binary Reported:** TurtleCoind

#### Expected Behavior

Daemon is expected to start and run.

#### Current Behavior

Daemon crashes opening the DB within the RocksDBWrapper. Tried deleting existing DB with no change.

##### Screen Shots

![image](https://user-images.githubusercontent.com/14900108/45593006-f04b0680-b949-11e8-9b12-86c1429a5b71.png)

#### Steps to Reproduce

1. Execute Turtlecoind on an IvyBridge generation Intel CPU running Windows.
2. ...
3. No profit.

#### Possible Solution

CPU instructions present in Haswell (daemon works) but missing from Ivy Bridge (doesn't work) - VT-x, AVX2, FMA3

Based on Discord conversation - 

GNU/iburnmycd™Today at 12:20 AM
Thanks.
Are you sure you built from release source ? And totally blew away any previous builds in the folder ?
I’m seeing stuff in there that doesn’t match the current default build Params

[...]

PR #526, opened 2018-09-29
[ADMINISTRATIVE] Upgrading to RocksDB 5.15.10

brandonlehmann:
Upgrading RocksDB to the latest in hopes of addressing #496 as well as picking up any other improvements.


brandonlehmann:
According to the user in #496 this resolved that error.

PR #528, opened 2018-09-30
fixed pronoun

andrewnk:


brandonlehmann:
Daemon is self-aware. It shares the experience with each user. “Our” is correct. :)

PR #541, opened 2018-10-10
[ENHANCEMENT] Crypto Performance Tests

brandonlehmann:


src/CryptoTest/main.cpp
```diff
@@ -28,22 +31,40 @@ int main(int argc, char** argv) {
     Hash hash = Hash();
 
     cn_fast_hash(rawData.data(), rawData.size(), hash);
```
zpalmtree: I don't see much point in doing cn_fast_hash if we're not going to do groestl, blake, etc, since it's just keccak. Other than that, this all looks fine.

brandonlehmann: That section remains unchanged other than the logging output. It’s being done for transparency.

PR #542, opened 2018-10-11
[BUGFIX] Memory Allocation in slow-hash.c

brandonlehmann:


zpalmtree:
Have you performance tested this? I think it's intentional, allocate the memory the very first time you use it, and then for each subsequent slow hash iteration you don't need to allocate it. If you have your iteration count set correctly, then the aes encryption will fill the scratchpad completely, so there shouldn't be rubbish left over from last time.

If you want to use different iterations which don't cover the whole scratchpad, perhaps you could call memzero? Even so, I'm not sure how slow_hash_allocate_state() works, does it even zero the memory on first allocation? Might be undefined behaviour.

zpalmtree:
I agree this will cause issues with soft shell though, since you're allocating a scratchpad once, maybe it's 1MB, and then you reach another iteration and need an allocation of 2MB, which isn't there. If the performance loss is significant, you could just allocate for the peak amount.

brandonlehmann:
I have tested it. Multiple times across a few different platforms. It produces no difference in performance.

zpalmtree:
OK.

brandonlehmann:
You're welcome to test it yourself for good measure as well :)

PR #545, opened 2018-10-12
[BUGFIX] Fixing regression of node fee info during rename

brandonlehmann:


zpalmtree:
Is it worth leaving in the old handler so people who don't update to 0.8.4 (and nest users) don't get a free ride?

brandonlehmann:
Probably not a bad idea.

brandonlehmann:
Updated

PR #552, opened 2018-10-16
Settle on one graey for gray_peers variable naming

fexra:
Noticed that some variables where labeled ``gray`` while others where labeled ``grey`` for the peer lis. For convenience and sanity I changed ``grey`` to ``gray`` based on the least mentions.

Nothing personal kid.


src/ConnectivityTool/ConnectivityTool.cpp
```diff
@@ -215,7 +215,7 @@ bool handle_get_daemon_info(po::variables_map& vm) {
       << "outgoing_connections_count: " << res.outgoing_connections_count << ENDL
       << "incoming_connections_count: " << res.incoming_connections_count << ENDL
       << "white_peerlist_size: " << res.white_peerlist_size << ENDL
-      << "grey_peerlist_size: " << res.grey_peerlist_size << ENDL;
```
brandonlehmann: Didn't we delete the connectivity tool?

PR #577, opened 2018-10-28
[ENHANCEMENT] Added Additional Performance Benchmarks to CryptoTest

brandonlehmann:


src/CryptoTest/main.cpp
```diff
@@ -180,13 +180,45 @@ int main(int argc, char** argv) {
     auto elapsedTime = std::chrono::high_resolution_clock::now() - startTimer;
     std::cout << "cn_slow_hash_v0: " << (PERFORMANCE_ITERATIONS / std::chrono::duration_cast<std::chrono::seconds>(elapsedTime).count()) << " H/s\n";
 
+    startTimer = std::chrono::high_resolution_clock::now();
+    for (auto i = 0; i < PERFORMANCE_ITERATIONS; i++)
+    {
+      cn_slow_hash_v1(rawData.data(), rawData.size(), hash);
```
zpalmtree: These tests don't fall under the `if (input.size() > 43)` guard above.

brandonlehmann: Good catch.

brandonlehmann: Now they do :)

PR #578, opened 2018-10-28
[ENHANCEMENT] Replace Boost Program Options In Binaries

brandonlehmann:
The goal of this PR is to remove the boost requirement for simple CLI argument processing as well as generating the loading configuration files (JSON) for TurtleCoind and turtle-service.

As a result of some of the changes, I've grabbed the compiler updates for travis from the C++17 branch but have NOT updated the code to C++17. 

Definitely need a full test of all binaries to make sure I didn't mangle any options in the process.


src/Daemon/DaemonConfiguration.h
```diff
@@ -0,0 +1,520 @@
...
+    {
+      std::cout << "Error: Unable to parse command line argument options: " << e.what() << std::endl << std::endl
+        << options.help({}) << std::endl;
+      exit(1);
+    }
+  };
+
+  void handleSettings(const std::string configFile, DaemonConfiguration& config)
+  {
+    std::ifstream data(configFile.c_str());
```
zpalmtree: There's no need to use .c_str() for this, ifstream's take a string argument just fine.

I *think* this will create the file if it doesn't exist already, leaving a blank file hanging around.

In C++17, we can check for file existence with `std::filesystem::exists(configFile);` which doesn't leave a file hanging around.
Furthermore if using C++17 we can use `std::filesystem::remove(configFile)` to clean up this file, whilst in C++11 we can use `std::remove(configFile.c_str());`

It's probably better to use the bool operator of std::ifstream to see if a file exists if we're using C++11:
```
if (std::ifstream(configFile))
{
    // file exists
}
```

Or, if you want to bind it to a var:

```
std::ifstream data(configFile);
if (data)
{
    // file exists
}
```

src/config/CryptoNoteConfig.h
```diff
@@ -244,7 +244,13 @@ const uint64_t P2P_DEFAULT_INVOKE_TIMEOUT                    = 60 * 2 * 1000; //
 const size_t   P2P_DEFAULT_HANDSHAKE_INVOKE_TIMEOUT          = 5000;          // 5 seconds
 const char     P2P_STAT_TRUSTED_PUB_KEY[]                    = "";
 
+const uint64_t DATABASE_WRITE_BUFFER_MB_DEFAULT_SIZE         = 256;
+const uint64_t DATABASE_READ_BUFFER_MB_DEFAULT_SIZE          = 10;
+const uint32_t DATABASE_DEFAULT_MAX_OPEN_FILES               = 100;
+const uint16_t DATABASE_DEFAULT_BACKGROUND_THREADS_COUNT     = 2;
+
 const char     LATEST_VERSION_URL[]                          = "http://latest.turtlecoin.lol";
+const char     LICENSE_URL[]                                 = "https://github.com/turtlecoin/turtlecoin/blob/master/LICENSE";
```
zpalmtree: I'd prefer to make any new constants we add a `const std::string` if possible.

src/zedwallet/ParseArguments.cpp
```diff
@@ -1,254 +1,122 @@
...
 #include <config/WalletConfig.h>
-
 #include <zedwallet/Tools.h>
+#include "version.h"
 
-/* Thanks to https://stackoverflow.com/users/85381/iain for this small command
-   line parsing snippet! https://stackoverflow.com/a/868894/8737306 */
-char* getCmdOption(char ** begin, char ** end, const std::string & option)
+template <class Container>
+void split(const std::string& str, Container& cont, char delim = ' ')
```
zpalmtree: I'd much prefer if this function returned the container, rather than void. I also don't see much need for it to be a template, especially as not all containers support push_back anyway. Finally, it probably belongs in Tools.h/cpp.

src/zedwallet/ParseArguments.cpp
```diff
@@ -1,254 +1,122 @@
...
-            config.exit = true;
-            return config;
-        }
-
-        config.walletFile = std::string(wallet);
-        config.walletGiven = true;
-    }
-
-    if (cmdOptionExists(argv, argv+argc, "--password"))
+  std::vector<std::string> parts;
```
zpalmtree: Can we get 4 space tabs in this file? :)

src/zedwallet/ParseArguments.cpp
```diff
@@ -1,254 +1,122 @@
...
-
-        config.walletPass = std::string(password);
-        config.passGiven = true;
+      host = parts.at(0);
+      port = std::stoi(parts.at(1));
+      return true;
     }
-
-    if (cmdOptionExists(argv, argv+argc, "--remote-daemon"))
+    catch (std::exception& e)
```
zpalmtree: Are we trying to catch `std::stoi` throwing here? We should use `const std::invalid_argument &` instead. (Not using the exception value, so don't name it, so it doesn't go unused)

src/zedwallet/ParseArguments.cpp
```diff
@@ -1,254 +1,122 @@
...
+  cxxopts::Options options(argv[0], CryptoNote::getProjectCLIHeader());
+
+  bool help, version;
+  std::string remoteDaemon;
+
+  options.add_options("Core")
+    ("h,help", "Display this help message", cxxopts::value<bool>(help)->implicit_value("true"))
+    ("v,version", "Output software version information", cxxopts::value<bool>(version)->default_value("false")->implicit_value("true"))
+    ("debug", "Enable " + WalletConfig::walletdName + " debugging to "+ WalletConfig::walletName + ".log",
+      cxxopts::value<bool>(config.debug)->default_value("false")->implicit_value("true"));
```
zpalmtree: Out of interest, what does `default_value("false")->implicit_value("true")` do?

src/zedwallet/ZedWallet.cpp
```diff
@@ -34,12 +33,8 @@ int main(int argc, char **argv)
     #endif
 
     Config config = parseArguments(argc, argv);
-
-    /* User requested --help or --version, or invalid arguments */
-    if (config.exit)
```
zpalmtree: Looks like config.exit isn't being used anymore, should probably remove from the Config definition in Types.h

src/Daemon/Daemon.cpp
```diff
@@ -331,25 +294,27 @@ int main(int argc, char* argv[])
...
     }
 
-    logger(INFO) << "Starting core rpc server on address " << rpcConfig.getBindAddress();
-    rpcServer.start(rpcConfig.bindIp, rpcConfig.bindPort);
-    rpcServer.setFeeAddress(command_line::get_arg(vm, arg_set_fee_address));
-    rpcServer.setFeeAmount(command_line::get_arg(vm, arg_set_fee_amount));
-    rpcServer.enableCors(command_line::get_arg(vm, arg_enable_cors));
+    // Fire up the RPC Server
+    logger(INFO) << "Starting core rpc server on address " << config.rpcInterface << ":" << config.rpcPort;
+    rpcServer.start(config.rpcInterface, config.rpcPort);
```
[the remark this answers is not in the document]
brandonlehmann: That may have been an accident. Will double check.

src/Daemon/Daemon.cpp
```diff
@@ -150,150 +134,129 @@ int main(int argc, char* argv[])
...
       }
     }
 
-    Level cfgLogLevel = static_cast<Level>(static_cast<int>(Logging::ERROR) + command_line::get_arg(vm, arg_log_level));
+    Level cfgLogLevel = static_cast<Level>(static_cast<int>(Logging::ERROR) + config.logLevel);
 
     // configure logging
     logManager.configure(buildLoggerConfiguration(cfgLogLevel, cfgLogFile));
 
-    /* Yay, ascii art ^__^ */
```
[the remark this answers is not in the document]
brandonlehmann: Correct.

src/zedwallet/ParseArguments.cpp
brandonlehmann: No, you heathen.

src/zedwallet/ParseArguments.cpp
brandonlehmann: Unspecified parameter uses the default value, false. 

If specify the argument flag it uses the implicit value (true) no matter the following value after it. Otherwise something like `—help me` would error. This way it’s just ignored and set to true because you specified the flag.

src/Daemon/DaemonConfiguration.h
brandonlehmann: Wouldn’t the file only get created with ofstream?

zpalmtree: Yes, you are correct. I was remembering some code where I used both ifstream and ofstream, to check if it existed, and that it was writable, sorry.

src/Daemon/Daemon.cpp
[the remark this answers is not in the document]
brandonlehmann: Yes, I think this is related to this. I've updated in upcoming commit such that the values are set before the RPC server is started.

src/config/CryptoNoteConfig.h
brandonlehmann: Adjusted

src/zedwallet/ParseArguments.cpp
brandonlehmann: fine

zpalmtree: :smile:

src/zedwallet/ParseArguments.cpp
brandonlehmann: I've moved `parseDaemonAddressFromString` over to `Tools.h` as well

src/zedwallet/ParseArguments.cpp
brandonlehmann: Updated

brandonlehmann:
No more conflicts :)

brandonlehmann:
@zpalmtree

PR #581, opened 2018-10-30
turtle-service: support old flat ini config for backward compatibility

rixombea:
When reading config file as new json format failed, try to read it as old ini format.

keep `--dump-config` and `--save-config` to only use new json format.

#### Testing turtle-service with ini config (all valid options):
```
; file: valid.conf
; this is a valid ini file comment
; this config only includes valid config key
rpc-password=523d4557fe5f9824ae71eaa026c618f41d2cdfd24688978585739e5f337c80ea
container-file=/home/mywallet.twl
container-password=secr3t=-passwUrtz
enable-cors=*
daemon-address=public.turtlenode.io
daemon-port=11898
# this is also valid comment
log-level=0
```

command: `./turtle-service -c valid.conf --dump-config`

output:
```
{
  "bind-address": "127.0.0.1",
  "bind-port": 8070,
  "container-file": "/home/mywallet.twl",
  "container-password": "secr3t=-passwUrtz",
  "daemon-address": "public.turtlenode.io",
  "daemon-port": 11898,
  "enable-cors": "*",
  "log-file": "service.log",
  "log-level": 0,
  "rpc-legacy-security": false,
  "rpc-password": "523d4557fe5f9824ae71eaa026c618f41d2cdfd24688978585739e5f337c80ea",
  "server-root": ""
}
```

#### Testing turtle-service with ini config (some invalid options):
```
; file: invalid.conf
; this config only includes invalid config key
rpc-password=523d4557fe5f9824ae71eaa026c618f41d2cdfd24688978585739e5f337c80ea
container-file=/home/mywallet.twl
container-password=secr3t=-passwUrtz
enable-cors=*
daemon-address=public.turtlenode.io
daemon-port=11898
# this is also valid comment
log-level=0
this-is-not-good=hahaha
wutz-in-invalid-flag=1
```

command: `./turtle-service -c invalid.conf --dump-config`
Output: 
```
There was an error parsing the specified configuration file. Please check the file and try again: 
One or more options in your config file was invalid!
```

### DIFF
--- a/src/WalletService/WalletServiceConfiguration.h
+++ b/src/WalletService/WalletServiceConfiguration.h
@@ -293,7 +293,8 @@ namespace PaymentService {
     // comments, first non space starts with # or ;
     static const std::regex cfgComment{R"x(\s*[;#])x"};
     std::smatch item;
-    std::string cfgKey, cfgValue;
+    std::string cfgKey;
+    std::string stdcfgValue;
 
     for (std::string line; std::getline(data, line);)
     {
@@ -306,6 +307,7 @@ namespace PaymentService {
         if (item.size() == 4) {
           cfgKey = item[1].str();
           cfgValue = item[2].str();
+          
           if(cfgKey.compare("daemon-address") == 0)
           {
             config.daemonAddress = cfgValue;
@@ -372,8 +374,8 @@ namespace PaymentService {
       throw std::runtime_error("The --config-file you specified does not exist, please check the filename and try again.");
     }
 
-    try{
-
+    try
+    {
       json j;
       data >> j;
 
--- a/src/WalletService/WalletServiceConfiguration.h
+++ b/src/WalletService/WalletServiceConfiguration.h
@@ -285,6 +285,84 @@ namespace PaymentService {
     }
   };
 
+   inline void handleIniConfig(std::ifstream& data, WalletServiceConfiguration& config){
+    // find key=value pair, respect whitespace before/after "="
+    // g0: full match, g1: match key, g2: match value
+    static const std::regex cfgItem{R"x(\s*(\S[^ \t=]*)\s*=\s*((\s?\S+)+)\s*$)x"};
+    
+    // comments, first non space starts with # or ;
+    static const std::regex cfgComment{R"x(\s*[;#])x"};
+    std::smatch item;
+    std::string cfgKey, cfgValue;
+
+    for (std::string line; std::getline(data, line);)
+    {
+      if (line.empty() || std::regex_match(line, item, cfgComment))
+      {
+        continue;
+      }
+      else if (std::regex_match(line, item, cfgItem))
+      {
+        if (item.size() == 4) {
+          cfgKey = item[1].str();
+          cfgValue = item[2].str();
+          if(cfgKey.compare("daemon-address") == 0)
+          {
+            config.daemonAddress = cfgValue;
+          }
+          else if (cfgKey.compare("daemon-port")  == 0)
+          {
+            config.daemonPort = stoi(cfgValue);
+          }
+          else if (cfgKey.compare("log-file")  == 0)
+          {
+            config.logFile = cfgValue;
+          }
+          else if (cfgKey.compare("log-level")  == 0)
+          {
+            config.logLevel = stoi(cfgValue);
+          }
+          else if (cfgKey.compare("container-file")  == 0)
+          {
+            config.containerFile = cfgValue;
+          }
+          else if (cfgKey.compare("container-password")  == 0)
+          {
+            config.containerPassword = cfgValue;
+          }
+          else if (cfgKey.compare("bind-address")  == 0)
+          {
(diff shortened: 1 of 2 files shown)


brandonlehmann:
If we don’t throw from the Json parse anymore... and then INI isn’t readable... does it still throw to stop program execution and warn the user? 

Part of me thinks it’s smarter to dump the INI file support and tell people to upgrade to the JSON format.

Or add `--legacy-config`  as it feels like bad form to have that support two types of files transparently as the INI should be deprecated and removed at some future point.

Thoughts?

rixombea:
I think it will get caught w/ this line:
https://github.com/turtlecoin/turtlecoin/blob/7d01471f9acdc9daa15ab2b3e534cfc9b2ada9d7/src/WalletService/ConfigurationManager.cpp#L40

the original json parser itself wasn't explicitly throw an exception.

the ini parser itself thrown error on invalid option as you can see on the "invalid config test" above, but since all options are finally passed to cxxopts, I think we only need to parse, do simple validation key validation & collect the options and let cxxopts perform full validations

and here's when I pass a very bad config file:
command: ``` $ ./turtle-service -c turtle-service --daemon-port=3232 --dump-config```
output: 
```
There was an error parsing the specified configuration file. Please check the file and try again: 
One or more options in your config file was invalid!
```

brandonlehmann:
> I think it will get caught w/ this line:
> [turtlecoin/src/WalletService/ConfigurationManager.cpp](https://github.com/turtlecoin/turtlecoin/blob/7d01471f9acdc9daa15ab2b3e534cfc9b2ada9d7/src/WalletService/ConfigurationManager.cpp#L40)
> 
> Line 40 in [7d01471](/turtlecoin/turtlecoin/commit/7d01471f9acdc9daa15ab2b3e534cfc9b2ada9d7)
> 
>  catch (std::exception& e) 
> the original json parser itself wasn't explicitly throw an exception.
> 
> the ini parser itself thrown error on invalid option as you can see on the "invalid config test" above, but since all options are finally passed to cxxopts, I think we only need to parse, do simple validation key validation & collect the options and let cxxopts perform full validations
> 
> and here's when I pass a very bad config file:
> command: ` $ ./turtle-service -c turtle-service --daemon-port=3232 --dump-config`
> output:
> 
> ```
> There was an error parsing the specified configuration file. Please check the file and try again: 
> One or more options in your config file was invalid!
> ```

You're right, it (JSON parser) didn't explicitly throw an error; but it did, which was caught in `ConfigurationManager.cpp` as you mentioned. The JSON library would throw and we didn't catch it in `WalletServiceConfiguration.h` as it was allowed to bubble up. As you've proven that it does stop execution when a bad file is found, I'm okay with it now.

src/WalletService/WalletServiceConfiguration.h
```diff
@@ -285,6 +285,86 @@ namespace PaymentService {
...
+          cfgKey = item[1].str();
+          cfgValue = item[2].str();
+          
+          if(cfgKey.compare("daemon-address") == 0)
+          {
+            config.daemonAddress = cfgValue;
+          }
+          else if (cfgKey.compare("daemon-port")  == 0)
+          {
+            config.daemonPort = std::stoi(cfgValue);
```
zpalmtree: std::stoi throws std::invalid_argument if given a non numeric argument - I guess this gets caught in ConfigurationManager, but we should probably catch it here so we can report a better error. It would be good to report the line it is failing to parse on, for easier debugging.

src/WalletService/WalletServiceConfiguration.h
```diff
@@ -285,6 +285,86 @@ namespace PaymentService {
...
+          {
+            config.rpcPassword = cfgValue;
+          }
+          else if (cfgKey.compare("server-root")  == 0)
+          {
+            config.serverRoot = cfgValue;
+          }
+          else
+          {
+            throw std::runtime_error("One or more options in your config file was invalid!");
```
zpalmtree: I think it would be good to report the line this is failing on as well, so the user knows which bit of the config file is incorrect.

src/WalletService/WalletServiceConfiguration.h
```diff
@@ -294,67 +374,79 @@ namespace PaymentService {
...
+        config.rpcPassword = j["rpc-password"].get<std::string>();
+      }
 
-    if (j.find("server-root") != j.end())
+      if (j.find("server-root") != j.end())
+      {
+        config.serverRoot = j["server-root"].get<std::string>();
+      }
+    }
+    catch (std::exception& e)
```
zpalmtree: If we have an invalid JSON file, will it report a helpful error from the JSON parse, if we then go on to try and unsuccessfully parse it as an ini?

Also, it might also be a good idea to instead catch only Json exceptions, rather than everything. For example - https://nlohmann.github.io/json/classnlohmann_1_1basic__json_a9a0aced019cb1d65bb49703406c84970.html
This way, if something weird happens, we're not reporting it as a json error.

PR #582, opened 2018-10-30
error handling + catch json::parse_error early

rixombea:
- error handling as suggested by @zpalmtree
- catch json parsing error before further processing

**test on bad config file:** `./turtle-service -c /usr/bin/gcc --dump-config`

Output: 
```
There was an error parsing the specified configuration file. Please check the file and try again: 
Bad/invalid config file
```

**test on invalid config value:** `./turtle-service -c invalid-value.conf --dump-config`

Output: 
```
There was an error parsing the specified configuration file. Please check the file and try again: 
Invalid value for bind-port
```

**test on invalid config key:** `./turtle-service -c invalid-key.conf --dump-config`

Output: 
```
There was an error parsing the specified configuration file. Please check the file and try again:
Unknown option: this-is-invalid-config-key
```
not doing full validation, since it'll be hands over to the cli handler

### DIFF
--- a/src/WalletService/WalletServiceConfiguration.h
+++ b/src/WalletService/WalletServiceConfiguration.h
@@ -285,7 +285,7 @@ namespace PaymentService {
     }
   };
 
-   inline void handleIniConfig(std::ifstream& data, WalletServiceConfiguration& config){
+  inline void handleIniConfig(std::ifstream& data, WalletServiceConfiguration& config){
     // find key=value pair, respect whitespace before/after "="
     // g0: full match, g1: match key, g2: match value
     static const std::regex cfgItem{R"x(\s*(\S[^ \t=]*)\s*=\s*((\s?\S+)+)\s*$)x"};
@@ -314,7 +314,11 @@ namespace PaymentService {
           }
           else if (cfgKey.compare("daemon-port")  == 0)
           {
-            config.daemonPort = std::stoi(cfgValue);
+            try{
+              config.daemonPort = std::stoi(cfgValue);
+            }catch(std::exception& e){
+              throw std::runtime_error(std::string("Invalid value for ") + cfgKey);
+            }
           }
           else if (cfgKey.compare("log-file")  == 0)
           {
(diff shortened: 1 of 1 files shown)


src/WalletService/WalletServiceConfiguration.h
```diff
@@ -314,15 +314,23 @@ namespace PaymentService {
           }
           else if (cfgKey.compare("daemon-port")  == 0)
           {
-            config.daemonPort = std::stoi(cfgValue);
+            try{
```
brandonlehmann: Please use consistent bracing that matches the rest of the style :) braces go on next line please :) Verify throughout the file.

src/WalletService/WalletServiceConfiguration.h
```diff
@@ -374,10 +392,17 @@ namespace PaymentService {
...
     }
 
+    json j;
+    bool isValidJson = true;
+
     try
     {
-      json j;
       data >> j;
+    }catch(json::parse_error& e){
```
brandonlehmann: How is the parse error returned to the user? You've caught it here and stuffed it in a bag never to be seen again.

rixombea: is it ok to do `std::cout` here? print the json.exception.parse_error before moving to ini parsing?

brandonlehmann: Tough call. It feels like without the information it's like a "Generic" error which @zpalmtree seemed against in the other PR comments. But throwing a `std::cout` there i think looks weird if it doesn't parse JSON but does parse INI... know what I mean? It also seems redundant if it doesn't parse with either one of them...

zpalmtree: Maybe we can do something like:

```
try
{
    parseJson();
}
catch (const json::parse::error &jsonError)
{
    try
    {
         parseIni();
    }
    catch (const std::exception &iniError)
    {
        std::cout << "Failed to parse config file as json: " << jsonError.what() << std::endl;
        std::cout << "Failed to parse config file as ini: " << iniError.what() << std::endl;
        throw std::runtime_error();
    }
}

This way we only report the json parse error if it's also an invalid ini.

rixombea:
closing this one, will only use ini parser to convert into new config format.

PR #583, opened 2018-10-31
config format conversion

rixombea:


src/Daemon/DaemonConfiguration.h
```diff
@@ -340,6 +340,265 @@ namespace {
...
+          updated += 1;
+        }
+        else if (cfgKey.compare("log-level") == 0 )
+        {
+          try
+          {
+            config.logLevel = std::stoi(cfgValue);
+            updated += 1;
+          }
+          catch(std::exception& e)
```
zpalmtree: So, if we can't parse it, we're skipping, instead of throwing an error? Did the original format support that? I feel like we should just be letting the exceptions get thrown, and catching them around the original function call, rather than having a ton of empty try/catches.

I'm not a massive fan of the update+=1 either, it seems kinda wrong to use an int for this when it's just being used as a bool. Why not updated true/false? Either

src/Daemon/DaemonConfiguration.h
```diff
@@ -340,6 +340,265 @@ namespace {
     }
   };
 
+  bool updateConfigFormat(const std::string configFile, DaemonConfiguration& config)
+  {
+    std::ifstream data(configFile);
+
+    if (!data.good())
+    {
+      throw std::runtime_error("The --config-file you specified does not exist, please check the filename and try again.");
```
zpalmtree: This gets thrown, and then eaten. Should this be eaten? Not sure. Maybe we should separate out how we want to do returns - we can either return false if we failed to parse/update, or throw an exception. Perhaps use one as convention for 'error, report to user', and the other for 'not an ini file, parse as json'?

src/Daemon/DaemonConfiguration.h
```diff
@@ -340,6 +340,265 @@ namespace {
...
+            // pass
+          }
+        }
+        else
+        {
+          for (auto c: cfgKey) 
+          {
+            if (static_cast<unsigned char>(c) > 127) 
+            {
+                throw std::runtime_error(std::string("Bad/invalid config file"));
```
zpalmtree: You don't need to create a string here, just `throw std::runtime_error("Bad/invalid config file");` works.

src/Daemon/Daemon.cpp
```diff
@@ -146,6 +146,19 @@ int main(int argc, char* argv[])
...
   {
+    try
+    {
+      if(updateConfigFormat(config.configFile, config))
+      {
+          std::cout << std::endl << "Updating daemon configuration format..." << std::endl;
+          asFile(config, config.configFile);
+      }
+    }
+    catch(std::exception& e)
```
zpalmtree: It's worth noting that even if this throws we may have a partial parse occuring, with some config settings being updated. Maybe worth returning a std::tuple<bool, Config> to avoid this?

```
const auto [parseSuccess, newConfig] = updateConfigFormat(config.configFile, config);

if (parseSuccess)
{
    config = newConfig;
    // etc
}

File: src/Daemon/DaemonConfiguration.h (hunk 1 above)
Comment (rixombea) [reply src/Daemon/DaemonConfiguration.h]: updated.

runtime_error will get caught and exit w/ proper message.
invalid_argument & out_of_range error thrown by stoi will get caught, and replace with runtime_error w/ better error message and will get caught.

File: src/Daemon/DaemonConfiguration.h (hunk 2 above)
Comment (rixombea) [reply src/Daemon/DaemonConfiguration.h]: runtime_error now properly handled (see above)

File: src/Daemon/Daemon.cpp (hunk 4 above)
Comment (rixombea) [reply src/Daemon/Daemon.cpp]: no more partial conversion in new revs, so I think it's not necessary.

File: src/Daemon/DaemonConfiguration.h (hunk 3 above)
Comment (rixombea) [reply src/Daemon/DaemonConfiguration.h]: yes sir

File: src/Daemon/DaemonConfiguration.h [hunk 5]
```diff
@@ -340,6 +340,263 @@ namespace {
...
+          }
+        }
+        else if (cfgKey.compare("no-console") == 0)
+        {
+          config.noConsole = cfgValue.at(0) == '1' ? true : false;
+          updated = true;
+        }
+        else if (cfgKey.compare("db-max-open-files") == 0)
+        {
+          try{
```
zpalmtree: Inconsistent bracing here.

src/Daemon/DaemonConfiguration.h
```diff
@@ -340,6 +340,263 @@ namespace {
...
+        else if (cfgKey.compare("db-threads") == 0)
+        {
+          try
+          {
+            config.dbThreads = std::stoi(cfgValue);
+            updated = true;
+          }
+          catch(std::exception& e)
+          {
+            throw std::runtime_error(std::string(e.what()) + " - Invalid value for option " + cfgKey );
```
zpalmtree: I see e.what() returns a `char *`, does the `+` operator not work for `char *`'s and `std::string`'s?

src/WalletService/WalletServiceConfiguration.h
```diff
@@ -374,81 +443,69 @@ namespace PaymentService {
...
-        // when failed reading as json, try reading it as old flat ini
-        // clear eof + fail bits
-        data.clear();
-        // reread from start
-        data.seekg(0, std::ios::beg);
-        handleIniConfig(data, config);
+      config.serverRoot = j["server-root"].get<std::string>();
     }
-  }
+  };
```
zpalmtree: Semicolons are not required after function definitions, only classes or structs. I'm not sure why there are semi colons in this file, probably ibmcd being used to javascript. I also don't think these functions should be inline. Not necessarily in scope for this PR, just noticed it.

src/Daemon/DaemonConfiguration.h
rixombea: updated

src/Daemon/DaemonConfiguration.h
rixombea: got this when I try:

```
error: invalid operands of types ‘const char*’ and ‘const char [29]’ to binary ‘operator+’
             throw std::runtime_error(e.what() + " - Invalid value for option " + cfgKey );
```

src/WalletService/WalletServiceConfiguration.h
rixombea: not sure, was just follow existing style in this file.

brandonlehmann: My JavaScript doesn’t use semicolons. Go look. Lol. 

Pick one:
Old habits die hard
The brace looked lonely
Reminder that’s the end of the function even though it’s nested well

As for the inline in a few places... I forget why I did that. I think Was getting a few warnings somewhere due to multiple declarations.

src/Daemon/DaemonConfiguration.h
brandonlehmann: @zpalmtree I believe that’s what we talked about in channel. INI should try to parse and then upgrade it to JSON syntax. Otherwise (aside from stoi), fail gracefully and move on to JSON parsing. If JSON parsing then fails throw a fit. INI is being deprecated :)

src/WalletService/WalletServiceConfiguration.h
rixombea: I tried removing the inline and was failed CI tests (multiple declarations)

brandonlehmann: Sounds familiar.

zpalmtree: I believe it's because we've got function implementations in a header file. Should probably just split it into a header/cpp pair, making functions inline is rarely the best fix.

src/Daemon/DaemonConfiguration.h
zpalmtree: Ok, no problem.

src/WalletService/WalletServiceConfiguration.h
brandonlehmann: That's exactly the reason -- but just header files are so nice.

zpalmtree: True - but unfortunately, it doesn't really jive with how C++ works without a lot of hackery :(

brandonlehmann: `inline` === a lot of hackery

zpalmtree: No, but if you want to keep things header only, it will get worse the more you do. There's a reason these header only libraries are horrible to read. There's also no real reason to do header only when we're not trying to pop it in as a dependency, and we're adding needless code bloat by inlining large functions.

brandonlehmann: Now that's a valid reason :)

rixombea: updated.

brandonlehmann:
@rixombea you’re up 😄

rixombea:
:point_up:

rixombea:
wait, I think we need to convert buffer values back to MB when calling `--dump-config` & `--save-config`, correct? @zpalmtree

zpalmtree:
Yeah... maybe we should just store them in bytes and be done with it

brandonlehmann:
I disagree @zpalmtree if the cli says MB in the help and that’s what we are used to using then the configs should be in MB as well.

rixombea:
OK, buffer values restored to MB for dump+save in new rev @brandonlehmann

zpalmtree:
Well, I would suggest updating the help. Just we're doing a lot of * 1000 / 1000 and it's confusing if we store internally as megabytes unless we label every variable as megabytes.

brandonlehmann:
Or revert what you did and update it in the database::init method where I forgot to... problem solved.

zpalmtree:
Yeah, but then we're carrying around a megabyte variable everywhere, and you need to update the vars to be dbReadBufferSizeMegabytes so it's obvious what we doing and we don't make this mistake again.

brandonlehmann:
Okay. Sounds good. :) I’ll touch it tomorrow :)

rixombea:
so, update help messages & take raw value? or...?

:confused:

brandonlehmann:
Neither.

brandonlehmann:
I pulled this into a branch and pretended that the DB MB/bytes "bug" never existed. I've preserved the commit history on the other branch.

rixombea:
much better!

and finally can close this one :smile:

brandonlehmann:
Thank you for your hard effort sir :)

PR #585, opened 2018-11-01
that's a big buffer

zpalmtree:
We weren't using megabytes when loading the default values, so we had a buffer of 256/10 bytes, lol


brandonlehmann:
You need to update the config and cli parsers to convert the MB values to byte sizes then :)

zpalmtree:
out of scope

brandonlehmann:
Negative ghost writer. It's well within scope.

zpalmtree:
Updated

PR #606, opened 2018-11-13
Adding daemon /queryblocksdetailed

brandonlehmann:
/queryblocks doesn't work like it *should*
/queryblockslite returns partial block information and obscures the coinbase transaction in shortInfo which is a BinaryArray

Enter, /queryblocksdetailed which returns the same data structure as /get_block_details_by_height, /get_blocks_details_by_heights, and /get_blocks_details_by_hashes using the same logic of the last known block

### DIFF
--- a/src/CryptoNoteCore/Core.cpp
+++ b/src/CryptoNoteCore/Core.cpp
@@ -448,6 +448,53 @@ bool Core::queryBlocksLite(const std::vector<Crypto::Hash>& knownBlockHashes, ui
   }
 }
 
+bool Core::queryBlocksDetailed(const std::vector<Crypto::Hash>& knownBlockHashes, uint64_t timestamp, uint32_t& startIndex,
+                           uint32_t& currentIndex, uint32_t& fullOffset, std::vector<BlockDetails>& entries) const {
+  assert(entries.empty());
+  assert(!chainsLeaves.empty());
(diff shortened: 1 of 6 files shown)


zpalmtree:
To add to what I just said, you could get rid of the requirement of needing to provide the genesis, to avoid having https://github.com/TurtlePay/blockchain-data-collection-agent/commit/115ec9a2d9c8bdff3df726ab9dd26fcd0c98d42f#diff-168726dbe96b3ce427e7fedce31bb0bcR52 this in your data collection agent.

brandonlehmann:
I'm not 100% sure I follow the "scan height + timestamp" piece if we're supplying lastknown block hashes... it does have the timestamp parameter though :)

As for requiring the genesis block -- I like having it as a nice stable foundation -- that way we know we haven't done something silly.

zpalmtree:
@brandonlehmann For if we want to start grabbing blocks from a certain height, and we don't know the hashes of those preceding blocks - timestamp is OK, but most users think in terms of heights, and there's always some error converting between the two.

Furthermore, it does the whole 'lite' fetch I assume, where if you want to start scanning from the block with timestamp of 300,000 - You'll still need all the hashes preceding it, which is pretty much a waste of our time.

It doesn't really apply to your use case of course.

PR #609, opened 2018-11-17
Allow a developer to say how many blocks they want from /queryblocksdetailed 

brandonlehmann:


src/CryptoNoteCore/Core.cpp
```diff
@@ -451,14 +451,32 @@ bool Core::queryBlocksLite(const std::vector<Crypto::Hash>& knownBlockHashes, ui
 }
 
 bool Core::queryBlocksDetailed(const std::vector<Crypto::Hash>& knownBlockHashes, uint64_t timestamp, uint32_t& startIndex,
-                           uint32_t& currentIndex, uint32_t& fullOffset, std::vector<BlockDetails>& entries) const {
+                           uint32_t& currentIndex, uint32_t& fullOffset, std::vector<BlockDetails>& entries, uint32_t blockCount) const {
```
zpalmtree: I assume currentIndex and fullOffset are block heights? Can we make these a uint64_t whilst we're here?

brandonlehmann: Look good @zpalmtree ?

zpalmtree: Yep

PR #620, opened 2018-11-27
LWMA3 Deep Freeze

brandonlehmann:
After numerous discussions regarding the DA changes here and watching a few other projects, putting LWMA3 on hold seems like the best course of action while we continue to monitor the DA situations with other projects.

### DIFF
--- a/src/config/CryptoNoteConfig.h
+++ b/src/config/CryptoNoteConfig.h
@@ -40,7 +40,7 @@ const uint64_t LWMA_2_DIFFICULTY_BLOCK_INDEX                 = 620000;
 const uint64_t LWMA_2_DIFFICULTY_BLOCK_INDEX_V2              = 700000;
 const uint64_t LWMA_2_DIFFICULTY_BLOCK_INDEX_V3              = 800000;
 
-const uint64_t LWMA_3_DIFFICULTY_BLOCK_INDEX                 = 1200000;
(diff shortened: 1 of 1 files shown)


crappyrules:
I think there is a need for an algorithm change.  Say what you want but 30mh showing up on the network after Monero switched algorithms.  There are less miners today on the network than there was when the network was running between 7-12m.  Granted smaller miners have given up to lower the number of miners, but you cannot feign ignorance of the possibility of unknown asics, fpgas or botnets and do nothing about it because the network falsely appears secure with higher hashrate.  Go ahead and ridicule away at my statements, but there are plenty of rumblings amongst the mining community.

brandonlehmann:
> I think there is a need for an algorithm change. Say what you want but 30mh showing up on the network after Monero switched algorithms. There are less miners today on the network than there was when the network was running between 7-12m. Granted smaller miners have given up to lower the number of miners, but you cannot feign ignorance of the possibility of unknown asics, fpgas or botnets and do nothing about it because the network falsely appears secure with higher hashrate. Go ahead and ridicule away at my statements, but there are plenty of rumblings amongst the mining community.

By *algorithm* change I infer that you're talking about the hashing algo and not DA.

That being said, the possibility of ASIC/FPGA/Botnets certainly exists; however, there are multiple other factors that can play a role in the network hashrate as well. Does that mean that an algo change is a bad idea? No. Does it mean that it's the best idea? No. All factors must be considered.

PR #621, opened 2018-11-27
Wallet Api for WalletBackend

zpalmtree:
Pop the [docs.yaml](https://github.com/turtlecoin/turtlecoin/blob/wallet-api/src/WalletApi/docs.yaml) file in here - http://editor.swagger.io/ - to view the documentation in a nicer format.

* Rewrites the daemon proxy
* Fixes some wallet backend bugs
* Exposes a RESTful API
* Also messes with the logger because it was being dumb before I ditched it (reference lifetime)

It's, uh, sorta tested.


crappyrules:
I have an idea for address aliasing that could be done on chain if anyone cares to hear it before you completely finish up your wallet re-write.

brandonlehmann:
> I have an idea for address aliasing that could be done on chain if anyone cares to hear it before you completely finish up your wallet re-write.

Oh yeah? I'm all ears :)

src/WalletApi/ApiDispatcher.cpp
```diff
@@ -250,12 +250,10 @@ void ApiDispatcher::middleware(
         res.set_header("Access-Control-Allow-Origin", m_corsHeader);
     }
 
-    /* TODO: Uncomment
```
brandonlehmann: This is pretty important, yeah?

RocksteadyTC:
Yeah let's hear it

fexra:
I like this idea but am conflicted on it as I'm not sure this is something that belongs in the core, especially if it reduces privacy. This seems something that would be better handled by app devs and smaller ecosystems that interact with TRTL. This would also keep the alias process decentralized so to speak.

My second issue is using the blockIndex as an second identifier, and allowing the main identifier (name) can be reused on different blocks. This will surely cause confusion or at the very least not provide a good oversight.

brandonlehmann:
Core or not core... hmmmm

I’d link it to the transaction hash instead of the block height because someone could time it just right and two people say they are RockSteady@1004102 with two transactions in the same block. There’s no way to tell which transaction contains the right alias so link against the txn hash is better IMO.

brandonlehmann:
At that point though you’re at Alais + @ + 64 characters. Might as well use the real address. :)

crappyrules:
Implementation is up for interpretation on the best approach.  But I do believe it should be a service built into the network so **all** wallets can recognize and support the alias and not dependent on a third party plug in that could just pack up it's balls and leave at any time it wants to. Might as well just use open alias at that point since the majority of the work is already done.

fexra:
On a complete other subject:

- It would be nice to get a JSON response back containing the newly created address on ``addresses/create` - probably even needed in case of multiple/many requests at a time.
- Postman logs a 201 Internal Server Error on ``addresses/create`` - most likely due to the description not matching.
- ``addresses/create`` takes some noticeable time to complete - 700-800ms on my machine - this is pretty long.

RocksteadyTC:
Yeah I'm gonna need you to make a new issue for that Fexra

zpalmtree:
* Create operations now return the created address
* Wallet no longer saves after adding an address, which makes the operation much faster

zpalmtree:
@crappyrules This issue with this is now you can no longer send to an alias without having scanned the blockchain for all aliases. If I create my wallet at height 100, but your alias was created at height 50, I don't know about it.

Furthermore, it also means we have to add the extra data to every transaction which we receive from the daemon, which currently is not included. It's a lot of overhead.

Also, we would have to store this alias information in the wallet file, if we don't want to store the tx_extra of every transaction - It wouldn't be *that* much data, but it would be a decent amount.

I don't think we can store aliases on the blockchain and interact with them from a wallet point of view. Potentially you could move this to the daemon side, as they should have all the blocks? (Discounting people wanting to send transactions when unsynced, which potentially can work) But then if you're using a remote daemon, you're straight up asking the daemon what address this user belongs to, which makes it very clear who you're sending to.

brandonlehmann:
So, what happens if the wallet crashes after i create the address and i didn't get a chance to save yet? Should it return the new keys, you know, just in case?

@zpalmtree @fexra

zpalmtree:
If this is an issue, then the developer can call /save.
If it isn't, then no harm?

crappyrules:
Fine. Google keys it is then.

brandonlehmann:
You’ve implemented it backwards if the user stretches it and sends the result of the stretching. This violates best practices and negates the intended effect.

include/WalletTypes.h
```diff
@@ -291,4 +291,109 @@ namespace WalletTypes
...
+        Crypto::PublicKey key;
+
+        /* The transaction hash of the transaction that contains this input */
+        Crypto::Hash parentTransactionHash;
+    };
+
+    inline void to_json(nlohmann::json &j, const WalletBlockInfo &w)
+    {
+        j = {
+            {"coinbaseTX", w.coinbaseTransaction},
```
brandonlehmann: I may have missed this before, but why are you serializing the coinbaseTransaction separately from the rest of the transactions?

src/CMakeLists.txt
```diff
@@ -16,12 +17,14 @@ file(GLOB_RECURSE Http HTTP/*)
 file(GLOB_RECURSE Logging Logging/*)
 file(GLOB_RECURSE miner Miner/*)
 file(GLOB_RECURSE Mnemonics Mnemonics/*)
+file(GLOB_RECURSE Nigel Nigel/*)
```
brandonlehmann: We need to pick an actual name for this before this actually gets merged I think :)

src/Nigel/Nigel.cpp
```diff
@@ -0,0 +1,405 @@
...
+        {
+        }
+    }
+
+    return false;
+}
+
+bool Nigel::getFeeInfo()
+{
+    const auto res = m_httpClient->Get("/feeinfo");
```
brandonlehmann: This is deprecated use `/fee`

src/Nigel/Nigel.cpp
```diff
@@ -0,0 +1,405 @@
...
+
+    getFeeInfo();
+
+    /* Now launch the background thread to constantly update the heights etc */
+    m_backgroundThread = std::thread(&Nigel::backgroundRefresh, this);
+}
+
+bool Nigel::getDaemonInfo()
+{
+    const auto res = m_httpClient->Get("/getinfo");
```
brandonlehmann: Deprecated use `/info`

src/Rpc/CoreRpcServerCommandsDefinitions.h
```diff
@@ -789,7 +817,7 @@ struct COMMAND_RPC_GET_WALLET_SYNC_DATA {
     uint64_t startTimestamp;
 
     void serialize(ISerializer &s) {
-      KV_MEMBER(blockIds);
+      s(blockIds, "blockHashCheckpoints");
```
brandonlehmann: I'm torn on whether the use of `Checkpoints` in the name is the best choice here. We've been calling it `lastKnownBlockHashes` or am I confused?

src/WalletApi/ApiDispatcher.cpp
```diff
@@ -0,0 +1,1524 @@
...
+
+std::string ApiDispatcher::hashPassword(const std::string password)
+{
+    using namespace CryptoPP;
+
+    /* Using SHA256 as the algorithm */
+    CryptoPP::PKCS5_PBKDF2_HMAC<CryptoPP::SHA256> pbkdf2;
+
+    /* Salt of all zeros (this is bad...) */
+    byte salt[16] = {};
```
brandonlehmann: This needs to be adjusted before merge -- perhaps it'd be best to create a set of random keys at startup and use one of the keys for salt. It doesn't have to be stored persistently anywhere (across restarts) but it could be used with the password supplied from the CLI (and then stored) as well as applied to the API request itself.

src/WalletApi/ParseArguments.cpp
```diff
@@ -0,0 +1,74 @@
...
+    options.add_options("Core")
+        ("h,help", "Display this help message", cxxopts::value<bool>(help)->implicit_value("true"))
+
+        ("v,version", "Output software version information", cxxopts::value<bool>(version)->default_value("false")->implicit_value("true"));
+
+    options.add_options("Network")
+        ("p,port", "The port to listen on for http requests",
+        cxxopts::value<uint16_t>(config.port)->default_value(std::to_string(CryptoNote::SERVICE_DEFAULT_PORT)), "<port>")
+
+        ("a,accept-external-requests", "Should we listen for requests from outside this machine",
```
brandonlehmann: This changes the *known* nomenclature across other tools in the core from `rpc-bind-ip` to `accept-external-requests`

I know this is a new package and all; however, that being said, the bigger problem I have here is that I may want to specify that I bind to a virtual IP or secondary NIC and I cannot do that without `rpc-bind-ip` , it appears my only options are `127.0.0.1` or `0.0.0.0` which is too limited.

src/WalletApi/ParseArguments.h
```diff
@@ -0,0 +1,24 @@
...
+//
+// Please see the included LICENSE file for more information.
+
+#pragma once
+
+#include <config/CryptoNoteConfig.h>
+
+struct Config
+{
+    /* Should we listen for requests from other PC's (0.0.0.0 vs 127.0.0.1) */
```
brandonlehmann: See previous comment.

include/WalletTypes.h
zpalmtree: The coinbase is not a normal transaction. We should not treat it as if it is.

src/CMakeLists.txt
zpalmtree: Suggestions?

src/Rpc/CoreRpcServerCommandsDefinitions.h
zpalmtree: lastKnownBlockHashes contains the latest 100 block hashes, whereas this contains the latest 100 block hashes, and the hash of any blocks known in increments of 5k.

include/WalletTypes.h
brandonlehmann: If by not “normal” you mean it has a different input type... by that same token one could argue that an input type in the future that is not or type 2 isn’t “normal”

src/Rpc/CoreRpcServerCommandsDefinitions.h
brandonlehmann: Increments of 5k makes that request massive doesn’t it ?

src/WalletApi/ApiDispatcher.cpp
zpalmtree: I think this will make it a lot harder to use - now the caller has to parse the CLI output rather than just putting a password in there and starting up the service. I wanted to get away from parsing the output by opening wallets via the API instead. Now something like a pool has to concern itself with the implementation of the API.

brandonlehmann: Why would they even need to know? They would need no knowledge of the salt. It’s only used internally... remember... the client doesn’t stretch the password and send it... ever.

src/Rpc/CoreRpcServerCommandsDefinitions.h
zpalmtree: If you're scanning from zero, that's only 200 more hashes, which isn't tiny, but not that much more. We could increase the increments if you want.

src/WalletApi/ApiDispatcher.cpp
zpalmtree: Right, I got confused because you wrote "as well as applied to the API request itself.". I'll fix this.

src/Rpc/CoreRpcServerCommandsDefinitions.h
brandonlehmann: I think that’s why the original did it in the n^2 fashion... it attempts to balance the request size with the chance of reorg... the deeper the hashes the less likely they are to change... it’s also less hashes for the daemon to go through overall which reduces daemon overhead

zpalmtree: Daemon overhead is not an issue in the majority of cases - it returns as soon as it finds one block, so if we supply 300 hashes and the daemon knows about the first, it won't check any of the others. The n^2 logic makes sense, but frankly the code to do it was hard to read which is why I went for this path, it's much simpler to understand.

src/WalletApi/ParseArguments.cpp
zpalmtree: The thought was to make it easier for people to understand how to receive requests from outside their machine. Maybe it's a good thing it's a tad more obscure, so they don't instantly get pwned.

src/Nigel/Nigel.cpp
zpalmtree: Fixed

src/Nigel/Nigel.cpp
zpalmtree: Fixed

src/WalletApi/ApiDispatcher.cpp
zpalmtree: Done

src/WalletApi/ParseArguments.cpp
zpalmtree: Done.

src/Rpc/CoreRpcServerCommandsDefinitions.h
brandonlehmann: Okay, I'll bite :)

src/CMakeLists.txt
brandonlehmann: @RocksteadyTC ? suggestions?

brandonlehmann: Nigel it is... lol

brandonlehmann:
#yolo

PR #628, opened 2018-12-06
Fix fork heights being ill formed when empty

zpalmtree:
Previously if the fork heights array was empty, we would be allocating an array of 0 size, which is technically not supported behaviour. Most compilers don't complain, but I got a report of this failing in MSVC. So, we can use a std::initializer_list instead.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,4 +1,4 @@
-os: Visual Studio 2017
+os: Visual Studio 2017 Preview
 
 environment:
   BOOST_ROOT: C:\Libraries\boost_1_67_0
--- a/src/Rpc/RpcServer.cpp
+++ b/src/Rpc/RpcServer.cpp
@@ -568,8 +568,8 @@ bool RpcServer::on_get_info(const COMMAND_RPC_GET_INFO::request& req, COMMAND_RP
   res.grey_peerlist_size = m_p2p.getPeerlistManager().get_gray_peers_count();
   res.last_known_block_index = std::max(static_cast<uint32_t>(1), m_protocol.getObservedHeight()) - 1;
   res.network_height = std::max(static_cast<uint32_t>(1), m_protocol.getBlockchainHeight());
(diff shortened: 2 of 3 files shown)


zpalmtree:
Well, this compiles, but it needed me to use `Visual Studio 2017 Preview` instead of `Visual Studio 2017` in appveyor. I'm not sure how often they update their images, and if this fix is in the mainline Visual Studio. We probably don't want people having to install the preview version to get the code compiling. Would be good if someone could test.

More info: https://developercommunity.visualstudio.com/content/problem/62652/cant-declare-constexpr-initializer-list.html

brandonlehmann:
I'm getting mixed results here.

This seems like an error caused by not following the fork guide correctly or that the fork guide does not explicitly explain. I'm not sure it's a good idea to push build environment updates for something so small that can be fixed by updating the fork guide (if required).

zpalmtree:
No, it's not an error with the fork guide. It's perfectly legal to not have any upgrade heights, but the current code makes that not compile on Windows, since the array is zero sized. It compiles on Linux fine. We shouldn't force the user to have upgrade heights if they don't want any.

brandonlehmann:
Tell me again why we can't use a `std::vector`?

zpalmtree:
Not a constexpr

brandonlehmann:
Can't we just put a 0 in the array? lol

zpalmtree:
I guess....

zpalmtree:
We're already requiring a pretty new VS for C++17 - Why not just leave this open till it's in mainline VS, rather than hacking around it?

brandonlehmann:
Because @RocksteadyTC comes through randomly and clicks merge on PRs... :)

RocksteadyTC:
come on now

brandonlehmann:
Checking license headers now to make sure that's legit.

RocksteadyTC:
I see green checks. JS

brandonlehmann:
Leaving open until mainline VS supports this...

RocksteadyTC:
?

brandonlehmann:
Holding still.

zpalmtree:
:weary:

PR #648, opened 2018-12-20
Miner usability improvements

zpalmtree:
* Command line args no longer mandatory (will prompt for address)
* Lowers scan time to 1 sec to reduce orphans
* Nicer, simpler, coloured output
* Boost no longer needed, bye bye
* Altered the indentation to make reviewing hard


zpalmtree:
This *may* be incorrect, but I also saw a small hashrate boost by swapping the mutex with a `std::atomic<>`. From roughly 5.2kh/s to 5.5kh/s.

brandonlehmann:
Which algo are you using that's giving you 5.5kh/s using the solo miner? I need to deploy this *elsewhere*

zpalmtree:
cn_turtle

PR #649, opened 2018-12-21
Update CryptoNoteProtocolHandler.cpp

19morpheus80:
Fixed in other branches but not this one; replaced 'TurtleCoin' with " << CryptoNote::CRYPTONOTE_NAME << "


zpalmtree:
Maybe it would just be better to rebase this branch to dev?

brandonlehmann:
True. I’ll handle.

PR #656, opened 2018-12-24
Reuse daemon connections in miner and zedwallet-beta

zpalmtree:
Easy way to verify that the miner/zedwallet-beta are only using one connection to communicate with the daemon:

```
lsof -i -a -p `pidof miner` -r1
```

If you turn on log level 4 in the daemon, you can also see when a new connection is created. This should occur only once.

This also changes the miner to use cpp-httplib.

### DIFF
--- a/external/cpp-httplib/httplib.h
+++ b/external/cpp-httplib/httplib.h
@@ -2,6 +2,7 @@
 //  httplib.h
 //
 //  Copyright (c) 2017 Yuji Hirose. All rights reserved.
+//  Copyright (c) 2018 The TurtleCoin Developers. All rights reserved.
 //  MIT License
 //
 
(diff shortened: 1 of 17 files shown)


zpalmtree:
Oh, I need to make sure this works on Windows (if it even compiles) before merging.

brandonlehmann:
@zpalmtree was Windows tested?

zpalmtree:
Not yet. Let me spin up a VM.

zpalmtree:
Seems to work. If Resource Monitor is to be believed, it's also reusing the same connection. And wow, I see how long it takes for a connection to be freed in there. Over a minute to free the socket after closing the program.

zpalmtree:
OK. Think we're good to merge again if this builds.

PR #657, opened 2018-12-27
don't need to remove that which we don't use

andrewnk:


michael-herwig:
Hey, just saw that commit and if you want to keep you dockers small for distribution you may consider [multi stage builds](https://docs.docker.com/develop/develop-images/multistage-build/). Further you may want to clear the apt cache everytime you updated it in order to install packages `apt-get clean`.

andrewnk:
I agree. I'm not a huge fan of how the Dockerfile is currently set up. It works, but I think we could improve it a bit. In addition to multi stage builds, some things I think would be cool are:

- a Dockerfile for each process (now that we can compile components separately), maybe an env variable to specify which program to compile
- a Dockerfile for running a public node
- docker-compose to wire the components together if need be
- the ability to pass all config values via env variables or file
- env variable to download and include the most recent checkpoints when running the daemon

There's a lot we could do. I've done a few, if not most, of these things in my own projects. I'm not totally sure if this repo is the place to do all those things or not. The current Dockerfile, while it is larger than it needs to be and can run multiple processes (which I am not a fan of), is simple enough for the novice to get up and running.

michael-herwig:
With multistage builds you should be able to just copy the executable you want into a fresh container, so compiling every app seperated may introduce some unnecessary overhead on initialization. With the binaries itself being not that big i would personally argue for making one base container containing all applications and then setup multiple derived containers with convinient entrypoints. You may also want to delete binaries not in use from the derived container if necessary. However if the size does matter that much it would be a good idea to use [alpine](https://hub.docker.com/_/alpine/).

Regarding a Dockerfile for public nodes I guess there is not that much to it, except using `EXPOSE` to clearify the ports that should be used?

For docker-compose I wrote a little configuration for mining
```
version: '3.3'

services:
daemon:
    image: registry.gitlab.com/galaxia-project/blockchain/xi
    container_name: xi_daemon
    restart: always
    expose:
    - "31018"
    - "21018"
    volumes:
    - ./.data:/data
    entrypoint: xi-daemon --data-dir /data --rpc-bind-ip 0.0.0.0 --fee-address Xizwr3KCvHXdSGi8kQtbdQfUKPZbqaYtZAHdh4BGqp63CkGZBGdaZWiQJysqqXJTnSPckRccAzNMaf2U3Lt2ZU9W4axVJqoatg --fee-amount 2

miner:
    image: registry.gitlab.com/galaxia-project/blockchain/xi
    container_name: xi_miner
    restart: always
    depends_on: 
    - daemon
    entrypoint: xi-miner --address Xizwr3KCvHXdSGi8kQtbdQfUKPZbqaYtZAHdh4BGqp63CkGZBGdaZWiQJysqqXJTnSPckRccAzNMaf2U3Lt2ZU9W4axVJqoatg --daemon-host daemon --log-level 3
```
but now that you wrote, it kinda seems bloated to me. In my opinion you have a good point that it would be beneficial to look into moving more CLI configurations into config files and env variables.

andrewnk:
My point in compiling individual binaries it not really about the final size, but rather reducing the compile time and unnecessary components. Why compile the miner if I don't need it? I agree multi stage builds is the way to go. We could just as easily pass a build argument to determine which binary to compile. Something similar to this:

```
ARG DEFAULT_BRANCH=master
ARG DEFAULT_BINARY=TurtleCoind

FROM ubuntu:latest as builder

ARG DEFAULT_BRANCH
ENV BRANCH=${DEFAULT_BRANCH}

ARG DEFAULT_BINARY
ENV BINARY=${DEFAULT_BINARY}

# install build dependencies
RUN apt-get update && \
    apt-get install -y \
      build-essential \
      python-dev \
      git \
      cmake \
      libboost-all-dev && \
    mkdir -p /opt/turtlecoin

WORKDIR /opt/turtlecoin

# clone
RUN git clone -b $BRANCH --single-branch https://github.com/turtlecoin/turtlecoin.git . && \
    mkdir build

WORKDIR /opt/turtlecoin/build

# build
RUN cmake -DCMAKE_CXX_FLAGS="-g0 -Os -fPIC -std=gnu++11" .. && \
    make -j$(nproc) $BINARY && \
    strip src/$BINARY

FROM frolvlad/alpine-glibc
ARG DEFAULT_BINARY
ENV BINARY=${DEFAULT_BINARY}

# add user
RUN addgroup -S turtlecoin && adduser -S turtlecoin -G turtlecoin -h /home/turtlecoin && \
    mkdir -p /home/turtlecoin/.TurtleCoin /home/turtlecoin/bin

#copy binary from builder
COPY --from=builder /opt/turtlecoin/build/src/$BINARY /home/turtlecoin/bin/$BINARY

WORKDIR /home/turtlecoin

# fix ownership
RUN chown -R turtlecoin:turtlecoin .

USER turtlecoin

#EXPOSE

ENTRYPOINT ["/bin/sh", "-c", "/home/turtlecoin/bin/$BINARY"]

#CMD
```

You could then build with the binary you wish to compile:
```docker build --build-arg DEFAULT_BINARY=TurtleCoind -t turtlecoind .```
```docker build --build-arg DEFAULT_BINARY=zedwallet -t zedwallet .```

As an aside, running on an alpine image with glibc like the code above, the daemon image size ends up being 22.9MB and the zedwallet image is 18.9MB.

michael-herwig:
This Dockerfile already looks very good. Reducing the compile time dependent on the build as well. The downside would be if you want to provide container for every single executable. Since `$BUILD` is used in the make command you would have to compile every single library that is shared accross two executables again, incresing your compile time. At Xi we moved away from multi stage builds for that reason. If we already have a pipeline building for ubuntu why not make a dockerfile simply copying the prebuilt binaries from its host. However this surely makes it more difficult for people to create the dockers themselves even that the build environment itself is a container. I can see you are using the latest ubuntu dist and install its default compiler that seems to be gcc7, you may want to consider upgrading it. We have our linux build environments documented [here](https://gitlab.com/galaxia-project/blockchain/develop-environments), maybe you find something usefull there. Besides thanks for that good work on the dockerfile. I already learned some new tweaks im going to use from now on.

andrewnk:
Thanks for the feedback! You’re absolutely right, I think the approach above may call for more workarounds when extending and would require the user to jump through more hoops than necessary. Your suggestions coupled with what Rock said on discord made me realize that zedwallet and TurtleCoind are really the only binaries that are needed for most applications and, therefore, building an image with just those in a base image is probably the way to go.

michael-herwig:
You could also build both in a builder image and reference it twice in the multi stage build. That way you would only once compile both binaries and could create two different containers for each application without the need of recompilation (ie. ` docker build --target zedwallet -t ... .`).

PR #660, opened 2018-12-28
WalletBackend fixes, code cleanup

zpalmtree:
* Fixes resetting not being performed correctly
* Fixes transactions being able to be half added causing duplicate transactions to be added
* Fixes splitAmountsIntoDenominations overflowing, causing incorrect amounts/fees to be sent
* Adds assertions that transaction outputs are PRETTY_AMOUNTS
* Adds assertions that transaction fee is as expected
* Fixes bug where the daemon returns a positive block height for a timestamp of zero
* Fixes off by one in in-memory GetBlocksByHeight error causing blocks to be skipped

===

Since it wasn't wanting to link nicely on some platforms due to the circular dependencies, I refactored the code location a little bit.

* Pulled WalletError and ValidateParameters out into a separate folder, Errors.
* Pulled SubWallets + SubWallet into a separate folder, SubWallets.
* And pulled a few things from around into a Utilities folder.

The Common folder sorta covers this, but it has a lot of old code that we don't want to be using anymore. Easier if we add new stuff into the Utilities folder, and try to remove or replace bits in Common.

* Also cleaned up the cmake a little bit. Alphabetizing, and removed some unneeded target_link_libraries.

Sorry this isn't in a separate PR... it would be if the damn thing would build. Not sure why it decided now to break, when I'm fairly sure I didn't change any includes.


zpalmtree:
One more thing I need to fix.

zpalmtree:
Fixed

brandonlehmann:
Merge conflicts

zpalmtree:
Resolved

PR #670, opened 2019-01-03
Add docs link on startup

zpalmtree:


brandonlehmann:
Need to find a more permanent home for those :)

zpalmtree:
I agree

PR #687, opened 2019-01-13
Fix fall-through in BlockchainSynchronizer.cpp

thinkpol2:
If switch case UpdateConsumersResult::nothingChanged is triggered while m_node.getKnownBlockCount() != m_node.getLocalBlockCount, switch processing will fall-through to case UpdateConsumersResult::addedNewBlocks.

### DIFF
--- a/src/Transfers/BlockchainSynchronizer.cpp
+++ b/src/Transfers/BlockchainSynchronizer.cpp
@@ -539,8 +539,7 @@ void BlockchainSynchronizer::processBlocks(GetBlocksResponse& response) {
         m_logger(DEBUGGING) << "Blockchain updated, resume blockchain synchronization";
         std::this_thread::sleep_for(std::chrono::milliseconds(100));
       } 
-        break;
-      }
+      break;
 
     case UpdateConsumersResult::addedNewBlocks:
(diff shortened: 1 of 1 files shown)


brandonlehmann:
Nice. Will have to give this a go to see if everything still syncs properly. I'm sure it does but have to be safe. We don't get into this portion of the code that often.

zpalmtree:
Strange. The way this code is formatted, it seems like not breaking is intentional. However, it doesn't really make sense to not break. Personally I would have put a comment there if it was, but you know how the codebase is...

Maybe remove the `else` and just put the `break` on the next line rather than having the break on both codepaths?

i.e.

```
if (m_node.getKnownBlockCount() != m_node.getLocalBlockCount()) {
    m_logger(DEBUGGING) << "Blockchain updated, resume blockchain synchronization";
    std::this_thread::sleep_for(std::chrono::milliseconds(100));
}

break;
```

brandonlehmann:
I agree that putting the break outside the if makes sense... i've yet to build that update

thinkpol2:
I thought the same.. it looks like it belongs there, but I could not figure out why you'd want it to fall through like that.

PR #695, opened 2019-01-18
Speed up mnemonics lib compilation

zpalmtree:
To compile the mnemonics lib before:

```
real	2m0.206s
user	2m1.771s
sys	0m3.910s
```

After:

```
real	0m10.551s
user	0m15.931s
sys	0m2.576s
```


zpalmtree:
Fixed

PR #696, opened 2019-01-19
Fix strange console locking issue in Windows under non english locale...

ngeojiajun:
When the turtlecoind's logging system is >=3, it will output debug messages including socket error messages. However, under certain locale (mine is ja_JP) errorMessage(int) will produce output in native system language and cause the console to be locked.

Fix:
Modified FormatMessage call to use en_US locale regardless of user native language.

Hope it solves the problem


zpalmtree:
I'm not familiar with how the windows locale system works, is it possible to uninstall a locale? I know on Linux if you try and output with en_US.utf8 for example when the user doesn't have that installed, it crashes the program.

ngeojiajun:
In windows, "locale" sometimes means display language. Also, en_US is always available in windows regardless of display language.

zpalmtree:
Great, thanks for the response :+1:

Issue #705, opened 2019-01-26
JSON library is a bottleneck when saving


zpalmtree:
```
Started saving
Serialized wallet to JSON: 10375 milliseconds
Converted wallet to string: 9701 milliseconds
Derived PBKDF2 Key: 1994 milliseconds
Performed AES Encryption: 6614 milliseconds
Wrote data to file: 2549 milliseconds
Restarted wallet synchronizer: 0 milliseconds
Total time: 37 seconds
```

Turns out nlohmann/json is nice to use, but not the fastest. This is with a 500mb~ wallet, 20 seconds spent serializing to JSON.

Might be worth looking into faster libraries. If we're reallly focused on speed, we could even just output the JSON directly, it's pretty damn simple if we don't have anything we need to escape.

zpalmtree:
Possible good candidate: RapidJSON - https://github.com/Tencent/rapidjson - header only, fast.

Benchmarks: https://github.com/miloyip/nativejson-benchmark

PR #706, opened 2019-01-27
Replace wallet saving/loading with rapidjson library

zpalmtree:
Resolves: #705 

For the upsides, see #705 

Downsides: Extra code, now we have two JSON libraries.
It shouldn't be *too* much effort to replace the rest of the code with rapidjson.

However, rapidjson has a much less friendly interface, easier to make mistakes serializing, and requires a few template classes just to make just getting a value out of it streamlined.

Most data we currently return with nlohmann from the wallet-api is pretty lightweight, but others could potentially grow pretty large, for example, getting all transactions in a large container.

These would certainly benefit from using rapidjson.

So, unsure if we should selectively use rapidjson where needed, or get rid of nlohmann json and deal with the worse API everywhere.


brandonlehmann:
Cmake macro is wrong :(

zpalmtree:
It's correct, but I guess these Cmake versions don't support it.

brandonlehmann:
I got u fam

brandonlehmann:
Windows error that one isn't a CMake one :(

zpalmtree:
Compiles now. Windows.h macro fuckery.

zpalmtree:
Don't merge yet, wallet sync status is not saving correctly

zpalmtree:
Should be good to merge now.

PR #718, opened 2019-01-31
Fixes #717 fork countdown imprecision

### LINKED ISSUE #717: Fork Countdown Does Not Display Accurately
### System Information

**Operating System**: Manjaro Linux
**OS Version**: 18.02 Illyria
**CPU Make, Model, & Speed**: i7 7700k 4.2ghz
**Amount of RAM**: 16GB
**Hard Drive Type**: SSD
**Hard Drive Size**: 2TB
**Volume Type**: EXT4

### TurtleCoin Build Information

**Version:** v0.12.1.1281
**Binary Source:** self compiled
**If self compiled, please provide all the build commands you used**
```
yay -S turtlecoin-git --noconfirm
```

### Issue Report

**Binary Reported:** TurtleCoind

#### Expected Behavior

Fork Countdown displays days to current fork.

#### Current Behavior

Fork Countdown displays days to current fork with too much accuracy. It always displays the whole number to the hundredths place, e.g. `next fork in 2.00 days.` It also rounds down to the nearest whole day even if there are 23 hours estimated. For example, 2 days 23 hours would display as `2.00 days`. Finally, when the countdown reaches less than 1 full day the timer displays `next fork in 0.00 hours`.

#### Steps to Reproduce

Use the `status` command in the daemon when a fork height is near.

#### Possible Solution

Fix the countdown to only display whole numbers. Eg, `next fork in 2 days` vs `next fork in 2.00 days`

Fix the countdown to display `next fork in <1 day` instead of `next fork in 0.00 hours`.

thinkpol2:
Fixes #717

### DIFF
--- a/src/Utilities/FormatTools.cpp
+++ b/src/Utilities/FormatTools.cpp
@@ -151,7 +151,7 @@ std::string get_fork_time(
         }
     }
 
-    const float days = ((float)(next_fork - height) / CryptoNote::parameters::EXPECTED_NUMBER_OF_BLOCKS_PER_DAY);
+    const float days = (static_cast<float>(next_fork - height) / CryptoNote::parameters::EXPECTED_NUMBER_OF_BLOCKS_PER_DAY);
 
     std::stringstream stream;
 
--- a/src/Utilities/FormatTools.cpp
+++ b/src/Utilities/FormatTools.cpp
@@ -151,7 +151,7 @@ std::string get_fork_time(
         }
     }
 
-    const float days = (next_fork - height) / CryptoNote::parameters::EXPECTED_NUMBER_OF_BLOCKS_PER_DAY;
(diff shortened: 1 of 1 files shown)


src/Utilities/FormatTools.cpp
```diff
@@ -151,7 +151,7 @@ std::string get_fork_time(
         }
     }
 
-    const float days = (next_fork - height) / CryptoNote::parameters::EXPECTED_NUMBER_OF_BLOCKS_PER_DAY;
+    const float days = ((float)(next_fork - height) / CryptoNote::parameters::EXPECTED_NUMBER_OF_BLOCKS_PER_DAY);
```
zpalmtree: Can we use `static_cast<float>(next_fork - height)` instead? :smiley:

thinkpol2: Updated.

PR #719, opened 2019-01-31
Stop faux socket reuse in httplib and wallet api (#703)

### LINKED ISSUE #703: You can open two wallet-api's on the same port with no issue
Since the port is already in use, this should fail. Probably need to tweak something in external/cpp-httlib

brandonlehmann:
Resolves #703

### DIFF
--- a/external/cpp-httplib/httplib.h
+++ b/external/cpp-httplib/httplib.h
@@ -610,7 +610,7 @@ socket_t create_socket(const char* host, int port, Fn fn, int socket_flags = 0)
 
        // Make 'reuse address' option available
        int yes = 1;
-       setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, (char*)&yes, sizeof(yes));
+       setsockopt(sock, SOL_SOCKET, SO_EXCLUSIVEADDRUSE, (char*)&yes, sizeof(yes));
 
        // bind or connect
        if (fn(sock, *rp)) {
--- a/external/cpp-httplib/httplib.h
+++ b/external/cpp-httplib/httplib.h
@@ -608,10 +608,6 @@ socket_t create_socket(const char* host, int port, Fn fn, int socket_flags = 0)
           continue;
        }
 
-       // Make 'reuse address' option available
-       int yes = 1;
-       setsockopt(sock, SOL_SOCKET, SO_EXCLUSIVEADDRUSE, (char*)&yes, sizeof(yes));
-
        // bind or connect
        if (fn(sock, *rp)) {
           freeaddrinfo(result);
--- a/src/WalletApi/ApiDispatcher.cpp
+++ b/src/WalletApi/ApiDispatcher.cpp
@@ -210,11 +210,9 @@ void ApiDispatcher::start()
 {
     if (!m_server.listen(m_host, m_port))
     {
-      std::ostringstream oss;
-      oss << "Could not bind service to " << m_host << ":" << m_port 
-          << std::endl << "Is another service using this address and port?";
-      std::cout << std::endl << oss.str() << std::endl << std::endl;
-      throw std::runtime_error(oss.str());
+      std::cout << "Could not bind service to " << m_host << ":" << m_port 
+                << "\nIs another service using this address and port?\n";
+      exit(1);
     }
 }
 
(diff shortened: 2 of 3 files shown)


external/cpp-httplib/httplib.h
```diff
@@ -610,7 +610,7 @@ socket_t create_socket(const char* host, int port, Fn fn, int socket_flags = 0)
 
        // Make 'reuse address' option available
        int yes = 1;
-       setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, (char*)&yes, sizeof(yes));
+       setsockopt(sock, SOL_SOCKET, SO_EXCLUSIVEADDRUSE, (char*)&yes, sizeof(yes));
```
zpalmtree: Doesn't compile on Linux: `error: use of undeclared identifier 'SO_EXCLUSIVEADDRUSE'`. We don't need this on Linux anyway, so perhaps use an ifdef.

src/WalletApi/WalletApi.cpp
```diff
@@ -41,6 +42,8 @@ int main(int argc, char **argv)
         /* Launch the API */
         apiThread = std::thread(&ApiDispatcher::start, api.get());
 
+        std::this_thread::sleep_for(std::chrono::milliseconds(250));
```
zpalmtree: Can we add a comment explaining the reason for this sleep.

src/WalletApi/ApiDispatcher.cpp
```diff
@@ -208,7 +208,14 @@ ApiDispatcher::ApiDispatcher(
...
 void ApiDispatcher::start()
 {
-    m_server.listen(m_host, m_port);
+    if (!m_server.listen(m_host, m_port))
+    {
+      std::ostringstream oss;
+      oss << "Could not bind service to " << m_host << ":" << m_port 
+          << std::endl << "Is another service using this address and port?";
+      std::cout << std::endl << oss.str() << std::endl << std::endl;
+      throw std::runtime_error(oss.str());
```
zpalmtree: I assume this exception isn't being caught, since the catch is in another thread?
If it is, do we need to print the message twice?
If not, can we use `exit()` instead?

src/WalletApi/ApiDispatcher.cpp
```diff
@@ -208,7 +208,14 @@ ApiDispatcher::ApiDispatcher(
 
 void ApiDispatcher::start()
 {
-    m_server.listen(m_host, m_port);
+    if (!m_server.listen(m_host, m_port))
+    {
+      std::ostringstream oss;
+      oss << "Could not bind service to " << m_host << ":" << m_port 
+          << std::endl << "Is another service using this address and port?";
```
zpalmtree: Minor complaint, but it's preferable to use `\n` instead of `std::endl` where possible - `std::endl` flushes the stream as well as printing a newline, which is unneeded here.

src/WalletApi/ApiDispatcher.cpp
brandonlehmann: True. It doesn’t get caught as it’s on another thread and passing in global exception pointed felt weird. 

It doesn’t actually print twice as the throw isn’t caught in the main thread. 

Exit would only exit the thread.

src/WalletApi/ApiDispatcher.cpp
brandonlehmann: Fair enough

brandonlehmann: Fair enough

external/cpp-httplib/httplib.h
brandonlehmann: Adding ifdef

src/WalletApi/WalletApi.cpp
brandonlehmann: No. Lol

RocksteadyTC: I can add one

brandonlehmann: I see what you did there.

src/WalletApi/ApiDispatcher.cpp
zpalmtree: Exit terminates the whole program on my machine, and the docs agree:
`Terminates the process normally, performing the regular cleanup for terminating programs.`
http://www.cplusplus.com/reference/cstdlib/exit/

src/WalletApi/WalletApi.cpp
RocksteadyTC: image

Issue #727, opened 2019-02-06
LiteBlocks Peers Shutdown Exploit


michael-herwig:
A daemon can enforce all connected daemons to shut down their connection to all daemons they are connected to using an exploit in the p2p lite blocks implementation.

#### Expected Behavior

Once a daemon notifies their peers about a new block using the `NOTIFY_NEW_LITE_BLOCK::request` the notified peer asks the publisher about missing transactions that may not be known to the receiver.

#### Current Behavior

An incoming lite block notification will force the receiver to ask every connected peer about missing transactions.

#### Steps to Reproduce

1. Send a lite block with random transaction hashes into the network.

#### Possible Solution

Only ask the publisher of the lite block about missing transactions.

#### Explanation

An attacker sends a lite block with random transaction hashes. Note the block does not be valid at all as this is only checked after all transaction details are known and these do not exist. All receivers will not know the transactions and ask every connected peer about the transactions using `NOTIFY_MISSING_TXS::request`. The request will be handled by `handle_notify_missing_txs` trying to get the transactions acquired. As they don't exist the method will declare the request as toxic and shut down the connection. Basically, you enforce all daemons to send an ill-formed request, as peers are required to send requests if and only if they already know the peer serves the data. Further, this exploit sends a huge payload into the network as you can encode an arbitrary amount of transactions into the lite block.

brandonlehmann:
Very nice find! Thanks for pointing this out.

ngeojiajun:
Actually it also flood the node resulting very high latency for RPC request. Especially when sending tx that requires 2 RPC calls will take maximum 3 minutes to send if the node is flooded with NOTIFY_MISSING_TX message. At worse case, node needed to be restarted to make it responsive to RPC. Do not sure is it related.

brandonlehmann:
See #730

Issue #729, opened 2019-02-06
LiteBlocks Alternative Chains Not Resolvable


michael-herwig:
When a daemon receives missing transactions of a lite block they will be tried to push into the transaction pool as the `NOTIFY_NEW_TRANSACTIONS::request` is used. That way alternative blocks may never get fully known as their transactions is invalid on the main chain.

#### Expected Behavior

The knowledge of known transactions is queried from the entire blockchain.

#### Current Behavior

The transactions will only be queried from the transaction pool.

#### Steps to Reproduce

1. Mine a block at a specific height including a transaction.
2. Mine a second block at the previous height with the same transaction.

#### Possible Solution

Query the entire blockchain for transaction hashes in a lite block. Implement a dedicated message type for handling missing transactions from a lite block. You can find our implementation [here](https://gitlab.com/galaxia-project/blockchain/xi/blob/lite-blocks/src/CryptoNoteProtocol/CryptoNoteProtocolHandler.h#L89) and [here](https://gitlab.com/galaxia-project/blockchain/xi/blob/lite-blocks/src/CryptoNoteCore/Core.cpp#L944)

#### Explanation

As the `NOTIFY_NEW_TRANSACTIONS::request` is used the transaction missing will always be pushed to the transaction pool. The transaction may have spent keys or is already mined in the main chain. Thus the lite block will never be able to gather all pieces of information as the transactions will never be available in the pool.

brandonlehmann:
Another good find. Thank you sir!

brandonlehmann:
See #730

Issue #728, opened 2019-02-06
LiteBlocks Payload Duplicaton


michael-herwig:
Once a daemon has a missing transaction in a lite block he requests them but does not add the lite block after.

#### Expected Behavior

After a peer sent the missing transactions the lite block gets added to the blockchain and will be propagated through the p2p network.

#### Current Behavior

The transactions will be pushed to the transaction pool.

#### Steps to Reproduce

1. Send a lite block with a transaction not known to another peer.

#### Possible Solution

Cache the lite block in the peer context who published the block and try to add it once the publisher sent missing transactions. You can look at our implementation [here](https://gitlab.com/galaxia-project/blockchain/xi/blob/lite-blocks/src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp#L590).

#### Explanation

Due to the hop increment of other nodes the node which has missed some transactions will not be notified again. Even that it has received the transactions potentially. On the next sync request, the node sees itself behind the network height and start a `NOTIFY_REQUEST_GET_OBJECTS::request` which will send the entire blocks with all transactions again which were just requested.

brandonlehmann:
Thanks again @michael-herwig  We'll work to get the fix incorporated.

brandonlehmann:
See #730

PR #730, opened 2019-02-06
lite block fixes

michael-herwig:
This is how far I got until a friend shown up that wanna drink with me :). Please have a look and give some feedback.

This should fix all lite block issues except the alternative chain problem as the missing transactions notification still pushes only into the transaction pool and some transactions may still are missing that would be valid on alternative chains.

### DIFF
--- a/src/CryptoNoteCore/Core.cpp
+++ b/src/CryptoNoteCore/Core.cpp
@@ -863,7 +863,7 @@ std::string Core::getPaymentIDFromExtra(const std::vector<uint8_t> &extra)
     return std::string();
 }
 
-boost::optional<BinaryArray> Core::getTransaction(const Crypto::Hash& hash) const {
+std::optional<BinaryArray> Core::getTransaction(const Crypto::Hash& hash) const {
     throwIfNotInitialized();
     auto segment = findSegmentContainingTransaction(hash);
     if(segment != nullptr) {
@@ -871,7 +871,7 @@ boost::optional<BinaryArray> Core::getTransaction(const Crypto::Hash& hash) cons
     } else if(transactionPool->checkIfTransactionPresent(hash)) {
         return transactionPool->getTransaction(hash).getTransactionBinaryArray();
     } else {
-        return boost::none;
+        return std::nullopt;
     }
 }
 
--- a/src/CryptoNoteCore/Core.cpp
+++ b/src/CryptoNoteCore/Core.cpp
@@ -863,6 +863,18 @@ std::string Core::getPaymentIDFromExtra(const std::vector<uint8_t> &extra)
     return std::string();
 }
 
+boost::optional<BinaryArray> Core::getTransaction(const Crypto::Hash& hash) const {
+    throwIfNotInitialized();
+    auto segment = findSegmentContainingTransaction(hash);
+    if(segment != nullptr) {
+        return segment->getRawTransactions({hash})[0];
+    } else if(transactionPool->checkIfTransactionPresent(hash)) {
+        return transactionPool->getTransaction(hash).getTransactionBinaryArray();
+    } else {
+        return boost::none;
+    }
+}
+
 void Core::getTransactions(const std::vector<Crypto::Hash>& transactionHashes, std::vector<BinaryArray>& transactions,
                            std::vector<Crypto::Hash>& missedHashes) const {
   assert(!chainsLeaves.empty());
--- a/src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp
+++ b/src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp
@@ -796,6 +796,7 @@ int CryptoNoteProtocolHandler::handle_notify_new_lite_block(int command, NOTIFY_
     req.current_blockchain_height = arg.current_blockchain_height;
     req.blockHash = CachedBlock(newBlockTemplate).getBlockHash();
     req.missing_txs = std::move(need_txs);
+    context.m_pending_lite_block = PendingLiteBlock{arg, {req.missing_txs.begin(), req.missing_txs.end()}};
 
     if (!post_notify<NOTIFY_MISSING_TXS>(*m_p2p, req, context)) {
         logger(Logging::ERROR)
--- a/src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp
+++ b/src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp
@@ -642,7 +642,7 @@ int CryptoNoteProtocolHandler::doPushLiteBlock(NOTIFY_NEW_LITE_BLOCK::request ar
     else {
         if(context.m_pending_lite_block.has_value()) {
             context.m_pending_lite_block = std::nullopt;
-            logger(Logging::DEBUGGING) << context << " Peer has a pending lite block but didnt provide all necassary transactions, dropping connection.";
+            logger(Logging::DEBUGGING) << context << " Peer has a pending lite block but didn't provide all necessary transactions, dropping the connection.";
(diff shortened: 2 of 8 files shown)


src/CryptoNoteCore/Core.cpp
```diff
@@ -863,6 +863,18 @@ std::string Core::getPaymentIDFromExtra(const std::vector<uint8_t> &extra)
     return std::string();
 }
 
+boost::optional<BinaryArray> Core::getTransaction(const Crypto::Hash& hash) const {
```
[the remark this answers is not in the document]
michael-herwig: does cmake enforce c++17?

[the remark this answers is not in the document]
michael-herwig: ok it does gonna change that

brandonlehmann:
For clarity, this should resolve #727 and #728 correct?

michael-herwig:
yes and partially #729, will add some explanation tomorrow

brandonlehmann:
Awesome. Good work!

michael-herwig:
I think this should do it for the three tickets. I didn't test it yet but the implementation should do the following things now:

* missing transactions of a lite block will only be requested from the publisher
* if a missing transaction request is sent the initial lite block receiver caches the lite block request as well as the request of the missing transaction
* if a node gets a new transactions it checks if there is a cached pending lite block request for this peer, if so the transactions are not considered as new pool transaction. Instead, they are pushed into the lite block routine.
* the lite block routine now can handle an additional context of provided transactions, it will search for provided transactions, blockchain transaction and pool transaction to fill all necessary transaction details.
* the push lite block routine has some additional sanity checks for checking if all requested missing transactions were provided or a peer tries to push a new lite block while one is already pending

src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp
```diff
@@ -413,18 +413,22 @@ int CryptoNoteProtocolHandler::handle_notify_new_transactions(int command, NOTIF
...
-  }
-
-  if (arg.txs.size()) {
-    //TODO: add announce usage here
-    relay_post_notify<NOTIFY_NEW_TRANSACTIONS>(*m_p2p, arg, &context.m_connection_id);
+  if(context.m_pending_lite_block.has_value()) {
+      logger(Logging::TRACE) << context << " Pending lite block detected, handling request as missing lite block transactions response";
+      return doPushLiteBlock(context.m_pending_lite_block->request, context, std::move(arg.txs));
+  } else {
+      for (auto tx_blob_it = arg.txs.begin(); tx_blob_it != arg.txs.end();) {
```
zpalmtree: Might be less error prone and a tad easier to read if we use `std::remove_if` here?

E.g.

```cpp
const auto it = std::remove_if(arg.txs.begin(), arg.txs.end(), [&m_core](const auto &tx)
{
    bool failed = !m_core.addTransactionToPool(tx);

    if (failed)
    {
        logger(Logging::DEBUGGING) << context << "Tx verification failed";
    }

    return failed;
});

if (it != arg.txs.end())
{
    arg.txs.erase(it, arg.txs.end());
}
```

src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp
```diff
@@ -413,18 +413,22 @@ int CryptoNoteProtocolHandler::handle_notify_new_transactions(int command, NOTIF
...
+          if (!m_core.addTransactionToPool(*tx_blob_it)) {
+              logger(Logging::DEBUGGING) << context << "Tx verification failed";
+              tx_blob_it = arg.txs.erase(tx_blob_it);
+          } else {
+              ++tx_blob_it;
+          }
+      }
+      if (arg.txs.size()) {
```
zpalmtree: I'd prefer explicitly checking the size here.

src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp
```diff
@@ -567,6 +571,121 @@ int CryptoNoteProtocolHandler::processObjects(CryptoNoteConnectionContext& conte
...
+    BlockTemplate newBlockTemplate;
+    if(!fromBinaryArray(newBlockTemplate, arg.blockTemplate)) { // deserialize blockTemplate
+        logger(Logging::WARNING) << context << "Deserialization of Block Template failed, dropping connection" ;
+        context.m_state = CryptoNoteConnectionContext::state_shutdown;
+        return 1;
+    }
+
+    std::unordered_map<Crypto::Hash, BinaryArray> provided_txs;
+    provided_txs.reserve(missingTxs.size());
+    for(std::size_t i = 0; i < missingTxs.size(); ++i) {
```
zpalmtree: Is there any reason to use a for loop over a ranged based loop here?

E.g.

```cpp
for (const auto &tx: missingTxs)
{
    provided_txs[getBinaryArrayHash(tx)] = tx;
}
```

Could use std::move here, but personally I'm not a fan of std::move when we're using the tx on the same line - makes it easy to invoke undefined behaviour if we refactor slightly.

I'm not seeing us using the `CachedTransaction i_provided_transaction{std::move(missingTxs[i])};` line either?

src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp
```diff
@@ -567,6 +571,121 @@ int CryptoNoteProtocolHandler::processObjects(CryptoNoteConnectionContext& conte
...
+    }
+
+    std::vector<BinaryArray> have_txs;
+    std::vector<Crypto::Hash> need_txs;
+
+    if(context.m_pending_lite_block.has_value()) {
+        for(const auto& requestedTxHash : context.m_pending_lite_block->missed_transactions) {
+            if(provided_txs.find(requestedTxHash) == provided_txs.end()) {
+                logger(Logging::DEBUGGING) << context << "Peer didn't provide a missing transaction, previously "
+                                                         "acquired for a lite block, dropping conneciton.";
```
zpalmtree: Typo, `conneciton`.

src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp
```diff
@@ -567,6 +571,121 @@ int CryptoNoteProtocolHandler::processObjects(CryptoNoteConnectionContext& conte
...
+
+    /*
+   * here we are finding out which txs are
+   * present in the pool and which are not
+   * further we check for transactions in
+   * the blockchain to accept alternative
+   * blocks.
+   */
+    for (const auto transactionHash: newBlockTemplate.transactionHashes) {
+        auto providedSerach = provided_txs.find(transactionHash);
```
zpalmtree: Typo, `providedSerach`.

src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp
michael-herwig: Yes I forgot to delete the CachedTransaction line there. For using a range based loop, if you want to move the object I tend to use a classic for in favor to not use `for(auto&& ...)`. Anway the double move is actually a bug in there gotta fix that.

src/CryptoNoteProtocol/CryptoNoteProtocolHandler.cpp
michael-herwig: `const auto` is missing `&` as well

zpalmtree:
Looks good. Thanks for the PR! Do we want to spin up a testnet @brandonlehmann @RocksteadyTC ?

brandonlehmann:
Probably a good idea.

RocksteadyTC:
I'm free for testnet today if we want to give this a shot.

PR #744, opened 2019-02-27
Blockchain Bloat Exploit Fix

### LINKED ISSUE #743: Blockchain Bloat Exploit
Transaction validation misses vital checks on sigantures count.

#### Expected Behavior
The signatures of a transaction are checked against the expected number of signatures.

#### Current Behavior
Signatures are validated to be valid keys and only a partial amount of signatures is checked against the outputs of the transaction.

#### Steps to Reproduce
1. Send a transaction and append an arbitrary amount of valid keys as signatures.

#### Possible Solution
Check the count of provided signatures against the number of expected signatures. In case of coinbase transactions zero otherwise the number of outputs. You can see a working implementation [here](https://gitlab.com/galaxia-project/blockchain/xi/blob/develop/src/CryptoNoteCore/Core.cpp#L1731), and [here](https://gitlab.com/galaxia-project/blockchain/xi/blob/develop/src/CryptoNoteCore/Core.cpp#L1544).

#### Explanation
Signatures are created for every output of a transaction that has an attached input. These signatures must be valid keys and encode a valid signature of the key image as well as outputs attached. All signatures are validated to be valid keys but only the first signatures are checked against the outputs. Thus embedding more signatures would not only lead to another bloat attack but further would be computational expensive as every point on the elliptic curve needs to be validated even if not further needed.

brandonlehmann:
Resolves #743 

Thanks to @michael-herwig for bringing this to our attention and providing a path for resolution.

### DIFF
--- a/src/CryptoNoteCore/Core.cpp
+++ b/src/CryptoNoteCore/Core.cpp
@@ -1,5 +1,6 @@
 // Copyright (c) 2012-2017, The CryptoNote developers, The Bytecoin developers
 // Copyright (c) 2014-2018, The Monero Project
+// Copyright (c) 2018, The Galaxia Project Developers
 // Copyright (c) 2018, The TurtleCoin Developers
 //
 // Please see the included LICENSE file for more information.
@@ -1391,7 +1392,7 @@ bool Core::getGlobalIndexesForRange(
                 getBinaryArrayHash(toBinaryArray(block.baseTransaction))
             );
         }
-        
+
         indexes = mainChain->getGlobalIndexes(transactionHashes);
 
         return true;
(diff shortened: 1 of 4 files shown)


brandonlehmann:
Hold this before merging. This **may** require a fork to implement or we risk invalidating older blocks.

brandonlehmann:
I'll be adding code to support this for planned fork at 1,400,000

brandonlehmann:
Updated to activate at 1,400,000

zpalmtree:
Should we bump version number and supported height with this, or want to do it in a separate PR?

brandonlehmann:
Why do you have to make sense all the time.

brandonlehmann:
Version bumped to 0.14.0 + support height index changed to 9 (1400000)

PR #752, opened 2019-03-09
WIP: Display Incoming Unlock Time in Blocks and Local Time using wallet-api + zedwallet-beta

dirtybits:
So far so good

### DIFF
--- a/src/zedwallet++/CommandImplementations.cpp
+++ b/src/zedwallet++/CommandImplementations.cpp
@@ -418,11 +418,13 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
         stream << "Payment ID: " << tx.paymentID << "\n";
     }
     
-    /* Display unlock time (if applicable) */
-    if (tx.unlockTime == 0)
+    /* Don't display unlock time */
+    if (tx.unlockTime == 0 ^ (tx.unlockTime < std::time(nullptr) && 
+                              tx.unlockTime > CryptoNote::parameters::DIFFICULTY_TARGET))
     {
         std::cout << SuccessMsg(stream.str()) << std::endl;
     }
+    /* Display Unlock time */
     else
     {
         std::cout << SuccessMsg(stream.str());
@@ -445,7 +447,7 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
             }
         }
         /* Here we treat Unlock as Unix time, and treat it that way in the future */
-        else
+        else if (tx.unlockTime > std::time(nullptr) && tx.unlockTime > Cryptonote::parameters::DIFFICULTY_TARGET)
         {
             std::cout << InformationMsg("Unlocks at ")
                       << InformationMsg(ZedUtilities::unixTimeToDate(tx.unlockTime))
--- a/src/zedwallet++/CommandImplementations.cpp
+++ b/src/zedwallet++/CommandImplementations.cpp
@@ -427,7 +427,8 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
     if (tx.unlockTime != 0)
     {
         std::cout << SuccessMsg(stream.str());
-
+        
+        /* Unlock time should be treated as a block, and is in the future */
         if (tx.unlockTime < CryptoNote::parameters::CRYPTONOTE_MAX_BLOCK_NUMBER)
         {
             int64_t difference = tx.unlockTime - tx.blockHeight;
@@ -437,6 +438,7 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
                 std::cout << InformationMsg("(Unlocks in ")
                           << InformationMsg(difference)
                           << InformationMsg(" blocks)")
+                          << std::endl
                           << std::endl;
             }
         }
@@ -447,6 +449,7 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
         std::cout << InformationMsg("(Unlocks at ")
                   << InformationMsg(ZedUtilities::unixTimeToDate(tx.timestamp))
                   << InformationMsg(")")
+                  << std::endl
                   << std::endl;
     }
 }
--- a/src/zedwallet++/CommandImplementations.cpp
+++ b/src/zedwallet++/CommandImplementations.cpp
@@ -433,7 +433,8 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
 
             if (difference > 0)
             {
-                int64_t unlockInUnixTime = tx.timestamp + tx.blockHeight*2 + tx.unlockTime;
+                unsigned short blocksPerMinute = 60 / CryptoNote::parameters::DIFFICULTY_TARGET;
+                int64_t unlockInUnixTime = tx.timestamp + (difference*blocksPerMinute);
 
                 std::cout << InformationMsg("Unlocks in ")
                           << InformationMsg(difference)
--- a/src/zedwallet++/CommandImplementations.cpp
+++ b/src/zedwallet++/CommandImplementations.cpp
@@ -447,7 +447,7 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
             }
         }
         /* Here we treat Unlock as Unix time, and treat it that way in the future */
-        else if (tx.unlockTime > std::time(nullptr) && tx.unlockTime > CryptoNote::parameters::CRYPTONOTE_MAX_BLOCK_NUMBER)
+        else if (tx.unlockTime > std::time(nullptr))
         {
             std::cout << InformationMsg("Unlocks at ")
                       << InformationMsg(ZedUtilities::unixTimeToDate(tx.unlockTime))
--- a/src/zedwallet++/CommandImplementations.cpp
+++ b/src/zedwallet++/CommandImplementations.cpp
@@ -426,10 +426,15 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
     else
     {
         std::cout << SuccessMsg(stream.str());
-        /* Let's treat Unlock as a block and as a time, and treat it that way in the future */
+
+        /* Here we treat Unlock as a block, and treat it that way in the future */
         if (tx.unlockTime < CryptoNote::parameters::CRYPTONOTE_MAX_BLOCK_NUMBER)
         {
             int64_t difference = tx.unlockTime - tx.blockHeight;
+            std::cout << "test"
+                      << std::endl
+                      << InformationMsg(ZedUtilities::unixTimeToDate(tx.unlockTime))
+                      << std::endl;
 
             if (difference > 0)
             {
@@ -442,6 +447,15 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
                           << std::endl
                           << std::endl;
             }
+        }
+        /* Here we treat Unlock as Unix time, and treat it that way in the future */
+        else
+        {
+
+            std::cout << InformationMsg("Unlocks at ")
+                      << InformationMsg(ZedUtilities::unixTimeToDate(tx.unlockTime))
+                      << std::endl
+                      << std::endl;
         }
     }
 }
--- a/src/zedwallet++/CommandImplementations.cpp
+++ b/src/zedwallet++/CommandImplementations.cpp
@@ -1,4 +1,5 @@
-// Copyright (c) 2018, The TurtleCoin Developers
+// Copyright (c) 2018-2019, The TurtleCoin Developers
+// Copyright (c) 2018-2019, The Catalyst Developers
 // 
 // Please see the included LICENSE file for more information.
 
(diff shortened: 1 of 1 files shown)


dirtybits:
Adding display of unlock time in blocks into incoming transfers via zedwallet-beta

src/zedwallet++/CommandImplementations.cpp
```diff
@@ -416,7 +417,21 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
         stream << "Payment ID: " << tx.paymentID << "\n";
     }
 
-    std::cout << SuccessMsg(stream.str()) << std::endl;
+    std::cout << SuccessMsg(stream.str());
+
+    /* Display unlock time */
+    if (tx.unlockTime != 0 && tx.unlockTime < 50000000)
```
zpalmtree: Can we use the MAX_BLOCK_NUMBER, or whatever that constant is called, from CryptoNoteConfig.h here?

src/zedwallet++/CommandImplementations.cpp
```diff
@@ -416,7 +417,21 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
...
+    {
+        int64_t difference = tx.unlockTime - tx.blockHeight;
+        if (difference > 0)
+        {
+            std::cout << "(Unlocks in " << difference << " blocks)" << std::endl;
+        }
+    }
+    else
```
zpalmtree: I think we should combine these if's, something like:

```cpp
if (tx.unlockTime != 0)
{
    /* Unlock time should be treated as a block, and is in the future */
    if (tx.unlockTime < CryptoNote::parameters::MAX_BLOCK_NUMBER)
    {
        int64_t difference = tx.unlockTime - tx.blockHeight;

        if (difference > 0)
        {
            std::cout << "(Unlocks in " << difference << " blocks)" << std::endl;
        }
    }
    /* Unlock time should be treated as a time, and is in the future */
    else if (tx.unlockTime > std::time(nullptr))
    {
        std::cout << "(Unlocks at " << ZedUtilities::unixTimeToDate(tx.timestamp) << ")" << std::endl;
    }
}
```

This way we will only print the unlock time if there is an unlock time, and the unlock time is in the future. I think with the code you have currently, this will print a timestamp even if the unlock time is zero.

src/zedwallet++/CommandImplementations.cpp
```diff
@@ -416,7 +417,21 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
...
+    {
+        std::cout << "(Unlocked at " << ZedUtilities::unixTimeToDate(tx.timestamp) << ")" << std::endl;
```
zpalmtree: Can we put some colour on these? Either include it in the stringstream above so it gets coloured green, or manually apply a SuccessMsg/InformationMsg. I think it probably would look better in yellow.

src/zedwallet++/CommandImplementations.cpp
dirtybits: Yes.

src/zedwallet++/CommandImplementations.cpp
```diff
@@ -416,7 +418,37 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
...
+            {
+                std::cout << InformationMsg("(Unlocks in ")
+                          << InformationMsg(difference)
+                          << InformationMsg(" blocks)")
+                          << std::endl;
+            }
+        }
+    }
+    /* Unlock time should be treated as a time, and is in the future */
+    else if (tx.unlockTime > std::time(nullptr))
```
zpalmtree: This if will never get hit, since you have `if (tx.unlockTime != 0)` above, which will always be true if there is any unlock time. You need to nest this under the `if (tx.unlockTime < CryptoNote::parameters::CRYPTONOTE_MAX_BLOCK_NUMBER)` if.

Furthermore, once you've done that, you can turn the `if (tx.unlockTime != 0)` into an `else` with the above `if (tx.unlockTime == 0)` if

src/zedwallet++/CommandImplementations.cpp
dirtybits: Done.

src/zedwallet++/CommandImplementations.cpp
dirtybits: Done.

dirtybits:
I'm confident it's okay now

dirtybits:
Why even have an unlock time, though? It's sort of smart-contract like, but what are some other use cases?

dirtybits:
Besides withholding funds from someone for a set amount of time.

src/zedwallet++/CommandImplementations.cpp
```diff
@@ -415,8 +417,33 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
...
+            {
+                int64_t unlockInUnixTime = tx.timestamp + (difference*CryptoNote::parameters::DIFFICULTY_TARGET);
+
+                std::cout << InformationMsg("Unlocks in ")
+                          << InformationMsg(difference)
+                          << InformationMsg(" blocks, at approximately ")
+                          << InformationMsg(ZedUtilities::unixTimeToDate(unlockInUnixTime))
+                          << std::endl
+                          << std::endl;
+            }
```
zpalmtree: Looks good now, but can we also print the unlock time when it's a timestamp?

dirtybits: Not sure what you mean, it's already printing a timestamp as far as I'm aware.

dirtybits: Oh, do you perhaps mean when it actually unlocks, we show the timestamp?

zpalmtree: I mean, when this is false: ```
if (tx.unlockTime < CryptoNote::parameters::CRYPTONOTE_MAX_BLOCK_NUMBER)
```

so the unlockTime is a unix timestamp, not a block height, also print that case.

**Comment by dirtybits** [comment]
Screenshot from 2019-03-17 00-48-49

**Comment by dirtybits** [comment]
It's working on my system

**Comment by brandonlehmann** [comment]
Uh, excuse me. “Works on my machine” is my tagline.

**Comment by dirtybits** [comment]
That's why I said system ;), also I've seen people on /g/ use that same phrase, so it's not yours

**Comment by brandonlehmann** [comment]
My first documented use predates 4chan... don’t make me pull out the BBS logs.

**Comment by dirtybits** [comment]
Pull out the BBS logs, and I'll show you the memes I've started (not exactly proud of them but I will still show them)

**Comment by brandonlehmann** [comment]
Any movement on the necessary changes?

**Comment by dirtybits** [comment]
- top incoming transfer uses block 180000 as it's unlockTime
- the bottom incoming transfer uses 500000001 as it's unlockTime (1 more than MAX_BLOCK_NUMBER) and is treated as Unix time
Screenshot from 2019-03-27 20-22-14

**Comment by dirtybits** [comment]
maybe we should do
```cpp
else if (tx.unlockTime > std::time(nullptr) && tx.unlockTime > tx.timestamp)
```

dirtybits:
```cpp
/* Here we treat Unlock as Unix time, and treat it that way in the future */
        else if (tx.unlockTime > std::time(nullptr))
        {
            std::cout << InformationMsg("Unlocks at ")
                      << InformationMsg(ZedUtilities::unixTimeToDate(tx.unlockTime))
                      << std::endl
                      << std::endl;
        }
        else
        {
            std::cout << InformationMsg("The unlock time was set at ")
                      << InformationMsg(ZedUtilities::unixTimeToDate(tx.unlockTime))
                      << InformationMsg(", which is before the timestamp of when it was sent")
                      << InformationMsg(", so it's already unlocked")
                      << std::endl
                      << std::endl;
        }
```

dirtybits:
Good to go.

src/zedwallet++/CommandImplementations.cpp
```diff
@@ -415,8 +417,44 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
     {
         stream << "Payment ID: " << tx.paymentID << "\n";
     }
+    
+    /* Don't display unlock time */
+    if (tx.unlockTime == 0 ^ (tx.unlockTime < std::time(nullptr) &&
```
zpalmtree: I think we can refactor this block to be a little simpler and get rid of this xor. Not tested if it compiles.

```
/* Don't display unlock time */
int64_t difference = tx.unlockTime - tx.blockHeight;

/* Here we treat Unlock as a block, and treat it that way in the future */
if (tx.unlockTime != 0 && difference > 0 && tx.unlockTime < CryptoNote::parameters::CRYPTONOTE_MAX_BLOCK_NUMBER)
{
    int64_t unlockInUnixTime = tx.timestamp + (difference * CryptoNote::parameters::DIFFICULTY_TARGET);

    std::cout << InformationMsg("Unlocks in ")
              << InformationMsg(difference)
              << InformationMsg(" blocks, at approximately ")
              << InformationMsg(ZedUtilities::unixTimeToDate(unlockInUnixTime))
              << std::endl
              << std::endl;
}
/* Here we treat Unlock as Unix time, and treat it that way in the future */
else if (tx.unlockTime > std::time(nullptr))
{
    std::cout << InformationMsg("Unlocks at ")
              << InformationMsg(ZedUtilities::unixTimeToDate(tx.unlockTime))
              << std::endl
              << std::endl;
}
else
{
    std::cout << SuccessMsg(stream.str()) << std::endl;
}
```

src/zedwallet++/CommandImplementations.cpp
```diff
@@ -415,8 +417,44 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
...
+                          << InformationMsg(ZedUtilities::unixTimeToDate(unlockInUnixTime))
+                          << std::endl
+                          << std::endl;
+            }
+        }
+        /* Here we treat Unlock as Unix time, and treat it that way in the future */
+        else if (tx.unlockTime > std::time(nullptr))
+        {
+            std::cout << InformationMsg("Unlocks at ")
+                      << InformationMsg(ZedUtilities::unixTimeToDate(tx.unlockTime))
```
zpalmtree: To fit with the other list transfer items, could we print this as `Unlock time: 2019-03-03 20:20` (for example)

src/zedwallet++/CommandImplementations.cpp
```diff
@@ -415,8 +417,44 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
...
 
-    std::cout << SuccessMsg(stream.str()) << std::endl;
+            if (difference > 0)
+            {
+                int64_t unlockInUnixTime = tx.timestamp + (difference*CryptoNote::parameters::DIFFICULTY_TARGET);
+
+                std::cout << InformationMsg("Unlocks in ")
+                          << InformationMsg(difference)
+                          << InformationMsg(" blocks, at approximately ")
+                          << InformationMsg(ZedUtilities::unixTimeToDate(unlockInUnixTime))
```
zpalmtree: To fit with the other list transfer items, could we print this as
```
Unlock height: 123456
Approximate unlock time: 2019-03-03 20:20
```

dirtybits: Indeed we can

src/zedwallet++/CommandImplementations.cpp
dirtybits: Sure we can

src/zedwallet++/CommandImplementations.cpp
dirtybits: Printing 
```cpp
std::cout << SuccessMsg(stream.str()) << std::endl;
```
 first looks better in my opinion.

zpalmtree: I don't mind if you do it first, but could we get rid of the xor?

dirtybits: Yes, that was supposed to be an "or" but I mistakenly put a "xor"

src/zedwallet++/CommandImplementations.cpp
```diff
@@ -415,8 +417,45 @@ void printIncomingTransfer(const WalletTypes::Transaction tx)
     {
         stream << "Payment ID: " << tx.paymentID << "\n";
     }
+    
+    /* Don't display unlock time */
+    if (tx.unlockTime == 0 || (tx.unlockTime < std::time(nullptr) &&
```
zpalmtree: I might be wrong, but I think this only catches the case where the unlock time is zero, or the timestamp is in the past. It doesn't catch the case where the unlock time is a block, in the past.

If that occurs, then since it will fail this check: `if (difference > 0)`, nothing will be printed. You could either pull the `difference` variable out to check against that in the case it's being treated as a block, or use the method i suggested earlier with printing the base case last.

dirtybits: You're right

dirtybits:
Appears to be working as expected:
If treated as block height:
case 1: unlockTime treated as a block in the future
case 2: unlockTime not printed because it's a block height in the past
If treated as Unix time
case 3: unlockTime treated as Unix time in the future
case 4: unlockTime not printed because it's a Unix time in the past

4 Incoming transfers and their respective cases from top to bottom:
case 3
case 1
case 2
case 4
Screenshot from 2019-03-28 22-42-39

zpalmtree:
On case 3, that's an exact unlock time, not an approximate timestamp, right?
The approximate timestamp is for when we calculated unlock time, from an unlock block height.
i.e., should be `Unlock time: 2019-03-29 09:33` rather than approximate unlock time. Other than that, looks perfect.

dirtybits:
Right. Fixed.

PR #753, opened 2019-03-11
Algo tweak to Soft-Shell to enable xmr-stak

davehlong:
xmr-stak alignment requires that scratchpad is divisible by 64 and division by 64 results in even number

### DIFF
--- a/src/crypto/hash.h
+++ b/src/crypto/hash.h
@@ -151,11 +151,7 @@ namespace Crypto {
     }
 
     uint32_t scratchpad = CN_SOFT_SHELL_MEMORY + (static_cast<uint32_t>(offset) * CN_SOFT_SHELL_PAD_MULTIPLIER);
-	scratchpad = (static_cast<uint64_t>(scratchpad / 64)) * 64;
-	if ((scratchpad / 64) % 2 != 0)
-	{
-		scratchpad -= 64;
-	}
+	scratchpad = (static_cast<uint64_t>(scratchpad / 128)) * 128;
     uint32_t iterations = CN_SOFT_SHELL_ITER + (static_cast<uint32_t>(offset) * CN_SOFT_SHELL_ITER_MULTIPLIER);
     uint32_t pagesize = scratchpad;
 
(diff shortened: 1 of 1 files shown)


src/crypto/hash.h
```diff
@@ -151,6 +151,11 @@ namespace Crypto {
     }
 
     uint32_t scratchpad = CN_SOFT_SHELL_MEMORY + (static_cast<uint32_t>(offset) * CN_SOFT_SHELL_PAD_MULTIPLIER);
+	scratchpad = (static_cast<uint64_t>(scratchpad / 64)) * 64;
+	if ((scratchpad / 64) % 2 != 0)
+	{
```
brandonlehmann: Question: If we've divided it by 64 above, chopped off the remainder via the conversion to uint64_t and then multiplied it again by 64 to make it based on 64bytes... what is this modulus operation for?

brandonlehmann: Per discussion in discord. This needs checked to validate that iterations remains >= (scratchpad / 2)

davehlong: > Question: If we've divided it by 64 above, chopped off the remainder via the conversion to uint64_t and then multiplied it again by 64 to make it based on 64bytes... what is this modulus operation for?

if the result of the division by 64 is not an even number, xmr-stak does not produce the correct hash, so the modulus is there to check for that, then reduce the scratchpad down by 64 so when it's dividsed by 64 it's aligned with the math in xmr-stak.

brandonlehmann: I’m confused

137 / 64 = 2.140625
Casting it a uint pops off the remainder
2 * 64 = 128 

It will always be an even number, yes?

brandonlehmann: I’m confused

137 / 64 = 2.140625
Casting it a uint pops off the remainder
2 * 64 = 128 

It will always be an even number, yes?

davehlong: scratchpad will always be an even number, but scratchpad/64 may not be, this could be odd or even

brandonlehmann: Can the code be simplified by taking it to 128 instead of 64 then?

davehlong: Code has been simplified by changing to /128 * 128 and dropping the mod op

brandonlehmann: Thanks! That feels cleaner :)

Issue #709, opened 2019-01-28
Add a logger to WalletBackend


zpalmtree:
A logger for WalletBackend would be very desirable. If possible, it would be nice if we could emulate the logger in https://github.com/turtlecoin/turtlecoin-wallet-backend-js/blob/master/lib/Logger.ts

I think it would be worth extending the logger to also add a `Component` variable, so we can use it more effectively in other parts of the codebase.

One thing I don't like about the current logger is that we have to pass it around everywhere, instead of just being able to import a file and modify a global object. I'm not sure the best way to do this in C++, or even if it is possible.

Desires for logger:
* Components, so we can filter messages for just one area of the codebase.
* Log levels (including Disabled)
* Categories, so we can do further filtering on a component
* A callback to handle the logger messages, so we can decide whether to log to a file, console, or something else dynamically.

The logger linked above has most of these already, so is a good starting point to work from.

PR #756, opened 2019-03-17
Add an optional logger to wallet-beta + zedwallet-beta

zpalmtree:
Uses a global variable, so don't have to pass the variable around everywhere...

Nice and configurable, you can provide a callback to print data in a specific way if you like.
Keeps with the current theme where log level 4 is the highest, 0 is the lowest.

Possibly need to add some more data to the logger, but this is a good start.

Fixes: #709


brandonlehmann:
If I had to guess, the order of linking is what AppVeyor is upset about.

PR #758, opened 2019-03-22
macos-travis-fix

hooftly:
Fixes compilation errors for OSX Clang.

### DIFF
--- a/src/Miner/BlockchainMonitor.cpp
+++ b/src/Miner/BlockchainMonitor.cpp
@@ -59,7 +59,7 @@ void BlockchainMonitor::waitBlockchainUpdate()
             nextBlockHash = requestLastBlockHash();
         }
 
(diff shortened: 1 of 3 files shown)


src/Miner/BlockchainMonitor.cpp
```diff
@@ -59,7 +59,7 @@ void BlockchainMonitor::waitBlockchainUpdate()
             nextBlockHash = requestLastBlockHash();
         }
 
-        if (lastBlockHash.value() != nextBlockHash.value())
+        if (lastBlockHash.operator*() != nextBlockHash.operator*())
```
zpalmtree: Could we use the operator directly, i.e. `*lastBlockHash != *nextBlockHash`?

I personally find using `.operator` very odd.

hooftly: Switched it up.

PR #761, opened 2019-03-29
WalletBackend (zedwallet++ and wallet-api) support for connecting to SSL daemons

### LINKED ISSUE #713: Add https support to cpp-httplib/nigel
It looks pretty easy to add SSL support to Nigel and thus wallet-api and zedwallet-beta.
https://github.com/yhirose/cpp-httplib#openssl-support

Relevant code: https://github.com/turtlecoin/turtlecoin/blob/development/src/Nigel/Nigel.cpp#L62

Currently, daemon's don't support SSL, unless proxied through something, so I'm not 100% sure on the best way to test this.

Could link into #712, as the blockchain cache API *does* support SSL

brandonlehmann:
Resolves #713 

Please see the updated build directions on how to enable SSL support. At this time, OpenSSL is linked against as a shared library. As such, I consider it an "advanced user" option as it decreases the portability of the binaries (in my testing at least).

I've done some testing of both zedwallet++ and wallet-api and everything appears to work correctly. Additional testing is appreciated.

The easiest way to enable SSL on a conventional daemon is to pass it through [ngrok](https://ngrok.com/) like so: `ngrok http 11898`

ngrok will then display both http and https (ssl) hostnames and will handle the encryption for you. Just remember that when you're connecting to the ngrok hostname to test SSL that you tell the wallets that the port is `443`.

### DIFF
--- a/src/Nigel/Nigel.cpp
+++ b/src/Nigel/Nigel.cpp
@@ -27,7 +27,7 @@ inline std::shared_ptr<httplib::Client> getClient(const std::string daemonHost, 
 #ifdef CPPHTTPLIB_OPENSSL_SUPPORT
     if (daemonSSL)
     {
-        return (std::shared_ptr<httplib::Client>) std::make_shared<httplib::SSLClient>(daemonHost.c_str(), daemonPort, timeout.count());
+        return std::make_shared<httplib::SSLClient>(daemonHost.c_str(), daemonPort, timeout.count());
     }
     else
     {
--- a/src/Nigel/Nigel.cpp
+++ b/src/Nigel/Nigel.cpp
@@ -1,5 +1,5 @@
-// Copyright (c) 2018, The TurtleCoin Developers
-// 
+// Copyright (c) 2018-2019, The TurtleCoin Developers
+//
 // Please see the included LICENSE file for more information.
 
 ////////////////////////
@@ -23,21 +23,50 @@ using json = nlohmann::json;
 ////////////////////////////////
 
 Nigel::Nigel(
-    const std::string daemonHost, 
-    const uint16_t daemonPort) : 
-    Nigel(daemonHost, daemonPort, std::chrono::seconds(10))
+    const std::string daemonHost,
+    const uint16_t daemonPort) :
+    Nigel(daemonHost, daemonPort, false, std::chrono::seconds(10))
 {
 }
 
 Nigel::Nigel(
-    const std::string daemonHost, 
+    const std::string daemonHost,
     const uint16_t daemonPort,
+    const bool daemonSSL) :
+    Nigel(daemonHost, daemonPort, daemonSSL, std::chrono::seconds(10))
+{
+}
+
+Nigel::Nigel(
+    const std::string daemonHost,
+    const uint16_t daemonPort,
+    const std::chrono::seconds timeout) :
+    Nigel(daemonHost, daemonPort, false, timeout)
+{
+}
+
+Nigel::Nigel(
+    const std::string daemonHost,
+    const uint16_t daemonPort,
+    const bool daemonSSL,
     const std::chrono::seconds timeout) :
     m_timeout(timeout),
     m_daemonHost(daemonHost),
     m_daemonPort(daemonPort),
-    m_httpClient(std::make_shared<httplib::Client>(daemonHost.c_str(), daemonPort, timeout.count()))
+    m_daemonSSL(daemonSSL)
 {
+#ifdef CPPHTTPLIB_OPENSSL_SUPPORT
+    if (m_daemonSSL)
+    {
+        m_httpsClient = std::make_shared<httplib::SSLClient>(daemonHost.c_str(), daemonPort, timeout.count());
+    }
+    else
+    {
+#endif
+        m_httpClient = std::make_shared<httplib::Client>(daemonHost.c_str(), daemonPort, timeout.count());
+#ifdef CPPHTTPLIB_OPENSSL_SUPPORT
+    }
+#endif
 }
 
 Nigel::~Nigel()
@@ -51,6 +80,11 @@ Nigel::~Nigel()
 
 void Nigel::swapNode(const std::string daemonHost, const uint16_t daemonPort)
 {
+    swapNode(daemonHost, daemonPort, false);
+}
+
+void Nigel::swapNode(const std::string daemonHost, const uint16_t daemonPort, const bool daemonSSL)
+{
     stop();
 
     m_localDaemonBlockCount = 0;
@@ -60,10 +94,24 @@ void Nigel::swapNode(const std::string daemonHost, const uint16_t daemonPort)
 
     m_daemonHost = daemonHost;
     m_daemonPort = daemonPort;
+    m_daemonSSL = daemonSSL;
 
-    m_httpClient = std::make_shared<httplib::Client>(
-        daemonHost.c_str(), daemonPort, m_timeout.count()
-    );
+#ifdef CPPHTTPLIB_OPENSSL_SUPPORT
+    if (m_daemonSSL)
+    {
+        m_httpsClient = std::make_shared<httplib::SSLClient>(
+            daemonHost.c_str(), daemonPort, m_timeout.count()
+        );
+    }
+    else
+    {
+#endif
+        m_httpClient = std::make_shared<httplib::Client>(
+            daemonHost.c_str(), daemonPort, m_timeout.count()
+        );
+#ifdef CPPHTTPLIB_OPENSSL_SUPPORT
+    }
+#endif
 
     init();
 }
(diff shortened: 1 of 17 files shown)


brandonlehmann:
Just remembered that when swapping nodes that I need to dispose of the previous handler if you're swapping SSL/non-SSL or vice versa

src/zedwallet++/ParseArguments.h
```diff
@@ -30,6 +30,9 @@ struct Config
 
     /* Controls what level of messages to log */
     Logger::LogLevel logLevel = Logger::DISABLED;
+    
+    /* Use SSL with daemon */
+    bool ssl;
```
zpalmtree: Looks to me like this doesn't get initialized if ssl support is disabled at compile time? Default initialization for bools is false, but I would prefer for that to be explicit.

brandonlehmann: Updated.

brandonlehmann:
Rebased, resolved conflicts, and squashed.

src/zedwallet++/GetInput.cpp
```diff
@@ -338,7 +342,11 @@ std::tuple<std::string, uint16_t> getDaemonAddress()
             continue;
         }
 
-        return {host, port};
+#ifdef CPPHTTPLIB_OPENSSL_SUPPORT
+        ssl = ZedUtilities::confirm("Does this daemon support SSL?", false);
```
zpalmtree: If you hit enter (default of localhost), this line will not be hit. Maybe that's intentional (or a happy coincidence) since it's unlikely localhost supports SSL?

src/Nigel/Nigel.cpp
```diff
@@ -85,9 +135,23 @@ std::tuple<bool, std::vector<WalletTypes::WalletBlockInfo>> Nigel::getWalletSync
         {"startTimestamp", startTimestamp}
     };
 
-    const auto res = m_httpClient->Post(
-        "/getwalletsyncdata", j.dump(), "application/json"
-    );
+    std::shared_ptr<httplib::Response> res;
+#ifdef CPPHTTPLIB_OPENSSL_SUPPORT
```
zpalmtree: Perhaps we could extract the #ifdefs into a single function? Something like..
```
auto getClient()
{
    #ifdef CPPHTTPLIB_OPENSSL_SUPPORT
    if (m_daemonSSL)
    {
        return m_httpsClient;
    }
    else
    {
    #endif
        return m_httpClient;
    #ifdef CPPHTTPLIB_OPENSSL_SUPPORT
    }
    #endif
}

const auto client = getClient();
const auto res = client->Post(
    "/getwalletsyncdata", j.dump(), "application/json"
);
```

Not 100% if `auto` for the return type will work, might need a little bit of template hackery.
This also has the advantage that any new functions added can't forget to wrap in an #ifdef.

src/Nigel/Nigel.h
```diff
@@ -33,6 +33,17 @@ class Nigel
         Nigel(
             const std::string daemonHost,
             const uint16_t daemonPort,
+            const bool daemonSSL);
+
+        Nigel(
```
zpalmtree: How come we're making these alternate constructors, are we using all these different combinations of parameters?

src/Nigel/Nigel.cpp
```diff
@@ -50,6 +79,11 @@ Nigel::~Nigel()
 //////////////////////
 
 void Nigel::swapNode(const std::string daemonHost, const uint16_t daemonPort)
```
zpalmtree: Again, any reason for the override here? Not seeing any usage where we don't pass the SSL arg.

src/WalletApi/ApiDispatcher.h
```diff
@@ -327,7 +327,7 @@ class ApiDispatcher
         //////////////////////////
 
         /* Extracts {host, port, filename, password}, from body */
```
zpalmtree: Outdated comment :smiley:

src/WalletApi/ApiDispatcher.cpp
```diff
@@ -883,8 +883,9 @@ std::tuple<Error, uint16_t> ApiDispatcher::setNodeInfo(
 
     const std::string daemonHost = tryGetJsonValue<std::string>(body, "daemonHost");
     const uint16_t daemonPort = tryGetJsonValue<uint16_t>(body, "daemonPort");
+    const bool daemonSSL = tryGetJsonValue<bool>(body, "daemonSSL");
```
zpalmtree: I don't see this value (and a few other calls) documented on https://turtlecoin.github.io/wallet-api-docs/ , would you care to update them? :smiley: 

Furthermore, should this parameter be optional, instead of mandatory, like with zedwallet? It will break existing code if it is mandatory.

src/WalletApi/ApiDispatcher.cpp
```diff
@@ -1557,7 +1560,12 @@ std::tuple<std::string, uint16_t, std::string, std::string>
         daemonPort = tryGetJsonValue<uint16_t>(body, "daemonPort");
     }
 
-    return {daemonHost, daemonPort, filename, password};
+    if (body.find("daemonSSL") != body.end())
+    {
+      daemonSSL = tryGetJsonValue<bool>(body, "daemonSSL");
```
zpalmtree: Indentation is off here

src/Nigel/Nigel.cpp
zpalmtree: That's pointing to the wrong line, sorry.

src/zedwallet++/GetInput.cpp
brandonlehmann: Correct. I juggled with this in my mind for a bit and considered it unlikely to be an issue for the time being.

src/Nigel/Nigel.cpp
brandonlehmann: I’ve been mulling over the best way to simplify that. I’ll give the auto route a try but no promises.

src/Nigel/Nigel.cpp
brandonlehmann: See above.

src/Nigel/Nigel.h
brandonlehmann: Left those there for the lazy in all of us. They can be removed at this point.

src/WalletApi/ApiDispatcher.cpp
brandonlehmann: Is it not optional already as it’s a try call? If not, does that mean that port is required?

brandonlehmann: Also regarding docs, baby steps.

src/WalletApi/ApiDispatcher.cpp
brandonlehmann: Pffft

src/WalletApi/ApiDispatcher.h
brandonlehmann: Pfffft

src/WalletApi/ApiDispatcher.cpp
zpalmtree: Hmm, good point about port. It's not optional, no. Perhaps the naming is poor.. it throws if it can't get the value, however, it returns a much more informative error message than the default error message when a value is not present.

src/WalletApi/ApiDispatcher.cpp
brandonlehmann: Resolved in incoming commit(s)

src/WalletApi/ApiDispatcher.h
brandonlehmann: Resolved in incoming commit(s)

src/Nigel/Nigel.h
brandonlehmann: Removed extra constructor(s) in incoming commit(s)

src/WalletApi/ApiDispatcher.cpp
brandonlehmann: I've made `daemonPort` and `daemonSSL` optional in the incoming commit(s). IMO, both should be optional to your point. I've also renamed `tryGetJsonValue` to `getJsonValue` since it throws :)

src/Nigel/Nigel.cpp
brandonlehmann: Since httplib::SSLClient extends httplib::Client, looks like casting httplib::SSLClient as httplib::Client works okay. Incoming commit(s) greatly simplify this code.

src/Nigel/Nigel.cpp
```diff
@@ -18,26 +18,49 @@
...
+////////////////////////////////
+/*   Inline helper methods    */
+////////////////////////////////
+
+inline std::shared_ptr<httplib::Client> getClient(const std::string daemonHost, const uint16_t daemonPort, const bool daemonSSL, const std::chrono::seconds timeout)
+{
+#ifdef CPPHTTPLIB_OPENSSL_SUPPORT
+    if (daemonSSL)
+    {
+        return (std::shared_ptr<httplib::Client>) std::make_shared<httplib::SSLClient>(daemonHost.c_str(), daemonPort, timeout.count());
```
zpalmtree: I believe this cast is not needed, we should be able to do just 
```
return std::make_shared<httplib::SSLClient>(daemonHost.c_str(), daemonPort, timeout.count());
```

In addition to being simpler, C style casts should not be used over safer C++ style casts.

brandonlehmann: Leave us C dinosaurs alone.

RocksteadyTC:
I'm Johnny Knoxville and this is TurtleCoin

brandonlehmann:
*sigh* now we have to start locking conversations.

src/zedwallet++/CommandImplementations.cpp
```diff
@@ -1,5 +1,4 @@
 // Copyright (c) 2018-2019, The TurtleCoin Developers
-// Copyright (c) 2018-2019, The Catalyst Developers
```
dirtybits: :(

zpalmtree: Sorry, didn't see this -- was probably a rebase mistake. @brandonlehmann Can we fix?

dirtybits: No worries, thank you.

PR #764, opened 2019-03-30
Block Transactions Blob Shuffle

### LINKED ISSUE #748: Block Transactions Blob Shuffle
Transactions can be shuffled in the block blob to enforce daemons get incompatible chains.

### Expected Behavior

The block header transactions hash array order is checked against the order of the included blobs.

### Current Behavior

The order is not checked such that multiple daemons can retrieve different blobs resulting in different cache states.

### Steps to Reproduce
1. Create a custom daemon shuffling transaction blobs randomly in mined blocks.
2. Create multiple, none published, small transactions with the same amounts and mine them into a block.
3. Send a differently shuffled block to each daemon, will not work using lite blocks.
4. Compute the global indices of the new amounts in the databases of the daemons according to the order they received.
5. Send transactions using these indices such that they are valid on a partial state of received orderings.

### Possible Solution

Check the transaction blobs ordering against the order in the block header. You can see a running example [here](https://gitlab.com/galaxia-project/blockchain/xi/blob/develop/src/CryptoNoteCore/Core.cpp#L718-734)

### Explanation

The transactions pushed to caches are extracted from the blob, see [here](https://github.com/turtlecoin/turtlecoin/blob/development/src/CryptoNoteCore/Core.cpp#L992). As these can have different orders the cache gets a slightly different state. This is then abused to create transactions that are only compatible on chains with a given state by an attacker. This step can then be repeated until most daemons get out of sync with different local information about transactions, mining blocks that are not valid for the others.

brandonlehmann:
Resolves #748 ... I think :)

### DIFF
--- a/src/CryptoNoteCore/Core.cpp
+++ b/src/CryptoNoteCore/Core.cpp
@@ -1,7 +1,7 @@
 // Copyright (c) 2012-2017, The CryptoNote developers, The Bytecoin developers
 // Copyright (c) 2014-2018, The Monero Project
-// Copyright (c) 2018, The Galaxia Project Developers
-// Copyright (c) 2018, The TurtleCoin Developers
+// Copyright (c) 2018-2019, The Galaxia Project Developers
+// Copyright (c) 2018-2019, The TurtleCoin Developers
 //
 // Please see the included LICENSE file for more information.
 
@@ -37,6 +37,7 @@
 
 #include <Utilities/FormatTools.h>
 #include <Utilities/LicenseCanary.h>
+#include <Utilities/Container.h>
 
 #include <unordered_set>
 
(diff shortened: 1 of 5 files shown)


src/Utilities/Utilities.h
```diff
@@ -96,4 +101,21 @@ namespace Utilities
 
         return false;
     }
+
+    /* Verify that the items in a collection are all unique */
+    template<typename T>
+    bool is_unique(T begin, T end)
+    {
+        std::set<typename T::value_type> set{};
```
zpalmtree: If we use `std::unordered_set`, it will both be potentially faster, and I believe, should remove the requirement for the `<` operator.

brandonlehmann: I may be mistaken but one of the points is to make sure that the set is actually ordered correctly.

src/Utilities/Utilities.h
```diff
@@ -96,4 +101,21 @@ namespace Utilities
 
         return false;
     }
+
+    /* Verify that the items in a collection are all unique */
+    template<typename T>
+    bool is_unique(T begin, T end)
```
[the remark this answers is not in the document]
brandonlehmann: Yeah, I missed that header :)

src/Utilities/Utilities.h
zpalmtree: I am not seeing where this requirement is checked, we are not utilizing the set after building it, and the name of the function does not suggest anything to do with ordering?

src/Utilities/Utilities.h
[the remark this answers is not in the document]
brandonlehmann: Oh sneaky, you just added it :)

src/Utilities/Utilities.h
brandonlehmann: Fair enough

zpalmtree: That's not to say we shouldn't be checking that they are ordered -- that seems to be what the original issue is referring to. Just that is not being verified here.

src/CryptoNoteCore/Core.cpp
```diff
@@ -1021,6 +1022,47 @@ std::error_code Core::addBlock(const CachedBlock& cachedBlock, RawBlock&& rawBlo
...
+    }
+
+    if (blockTemplate.transactionHashes.size() != transactions.size())
+    {
+      return error::BlockValidationError::TRANSACTION_INCONSISTENCY;
+    }
+
+    for (const auto &transaction: transactions)
+    {
+      auto search = std::find(blockTemplate.transactionHashes.begin(), blockTemplate.transactionHashes.end(), transaction.getTransactionHash());
```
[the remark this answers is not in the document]
brandonlehmann: Indeed

src/CryptoNoteCore/Core.cpp
```diff
@@ -1021,6 +1022,47 @@ std::error_code Core::addBlock(const CachedBlock& cachedBlock, RawBlock&& rawBlo
...
+    {
+      return error::BlockValidationError::TRANSACTION_DUPLICATES;
+    }
+
+    if (blockTemplate.transactionHashes.size() != transactions.size())
+    {
+      return error::BlockValidationError::TRANSACTION_INCONSISTENCY;
+    }
+
+    for (const auto &transaction: transactions)
```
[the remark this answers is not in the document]
brandonlehmann: Looks like it

brandonlehmann:
Pushing new commits & rebase after build test

src/Utilities/Utilities.h
brandonlehmann: Now it's being verified elsewhere.

Issue #694, opened 2019-01-18
Daemon had corrupted blockchain 2 times during last 2 days on v0.12.0 on different servers


minerrocks:
Daemons stopped working periodically on different servers/public cloud instances/, and then failed to start, giving logs:
```
2019-Jan-17 13:22:18.073971 INFO    [Core] Importing blocks from blockchain storage
2019-Jan-17 13:22:18.084632 ERROR   [Core] Corrupted blockchain. Block with index 1152103 and hash 77e0f0b7ff661854f238460f15fd95052c3765e145ceea735fab68d7abb01fee has previous block hash 3c1f4bea6df864466c47b4f273bddf35c5ef28cb1dd61af487ef0d362566b083, but parent has hash 0838b65872c658a7f3867ba90c67f5356e4f538d67dcd73c1f7f8c300f505568. Resynchronize your daemon please.
2019-Jan-17 13:22:18.152946 INFO    [RocksDBWrapper] Closing DB.
2019-Jan-17 13:22:18.161456 ERROR   [daemon] Exception: Blockchain storage is corrupted
```

### System Information

**Operating System**: Ubuntu
**OS Version**: 16.04
**CPU Make, Model, & Speed**: Intel Xeon Processor (Skylake, IBRS) 2100 MHz
**Amount of RAM**: 8 Gb
**Hard Drive Type**: Attached volume (public cloud)
**Hard Drive Size**: 80 Gb
**Volume Type**: ext4

### TurtleCoin Build Information

**Version:** 0.12.0.1280
**Binary Source:** precompiled -> https://github.com/turtlecoin/turtlecoin/releases/download/v0.12.0/turtlecoin-v0.12.0-linux.tar.gz

### Issue Report

**Binary Reported:** TurtleCoind

#### Expected Behavior

Daemon started and running fine.

#### Current Behavior

1) Daemon failed to start, giving logs:
```
2019-Jan-17 13:22:18.073971 INFO    [Core] Importing blocks from blockchain storage
[...]

PR #769, opened 2019-04-05
Local Blockchain Cache Enhancements

brandonlehmann:
This PR entails a number of enhancements:

* Builds & Links [sqlite3](https://www.sqlite.org/index.html) into the project
* **Experimental:** Enables the use of of a sqlite3 database for the local blockchain cache in the daemon via the `--sqlite` command line argument (replaces `blocks.bin` and `blockindexes.bin` with `blocks.bin.sqlite3`)
  * Conversion between the two formats is **not** currently supported. Switching between the two mechanisms will require a **full re-synchronization** of the local copy of the blockchain
* Adds the command line argument `--resync` to the daemon that deletes all of the relevant data from the data directory which is useful for forcing a resync without having to look for the data directory on the different platforms
* Adds the command line argument `--rewind #` to the daemon whereby it will [rewind](https://en.wikipedia.org/wiki/Be_Kind_Rewind) the local blockchain cache to a point where the blockchain syncronization will **restart** at the specified block (inclusive).
  * This is useful in attempting to *recover from arbitrary database corruption* (see #694 and countless Discord discussions in almost every channel) as it removes blockchain data starting from the specified height
* Altered the core "Corrupted Blockchain Database" message to provide possible recovery solutions including using the `--rewind #` and `--resync` options
* Altered the daemon to always create the directory specified by `--data-dir` if it doesn't exist thereby preventing the dreaded message about the directory not existing.


src/CryptoNoteCore/MainChainStorageSqlite.cpp
```diff
@@ -0,0 +1,231 @@
...
+            throw std::runtime_error("Failed to reset the autoincroment value of the table");
+        }
+    }
+
+    std::unique_ptr<IMainChainStorage> createSwappedMainChainStorageSqlite(const std::string &dataDir, const Currency &currency)
+    {
+        fs::path blocksFilename = fs::path(dataDir) / currency.blocksFileName();
+        fs::path indexesFilename = fs::path(dataDir) / currency.blockIndexesFileName();
+
+        std::unique_ptr<IMainChainStorage> storage(new MainChainStorageSqlite(blocksFilename.string() + ".sqlite3", indexesFilename.string()));
```
zpalmtree: We should use `std::make_unique` over using `new` here - 

```
const auto storage = std::make_unique<IMainChainStorage>(blocksFilename.string() + ".sqlite3", indexesFilename.string());
```
should do the trick.

src/Daemon/Daemon.cpp
```diff
@@ -221,6 +237,30 @@ int main(int argc, char* argv[])
     }
     CryptoNote::Currency currency = currencyBuilder.currency();
 
+    /* If we were told to rewind the blockchain to a certain height
+       we will remove blocks until we're back at the height specified */
+    if (config.rewindToHeight > 0)
```
zpalmtree: Does zero make it explode? I think that might be a common operation for users not familiar with the software. I see we're using zero as the default elsewhere, could maybe flip on the resync arg instead of rewind when given `== 0`?

src/Daemon/DaemonConfiguration.cpp
```diff
@@ -330,6 +344,11 @@ namespace DaemonConfig{
             throw std::runtime_error(std::string(e.what()) + " - Invalid value for " + cfgKey );
           }
         }
+        else if (cfgKey.compare("sqlite") == 0)
+        {
+          config.useSqliteForLocalCaches = cfgValue.at(0) == '1' ? true : false;
```
zpalmtree: Don't need the `true : false` here, can just do `config.useSqliteForLocalCache = cfgValue.at(0) == '1';`

include/CryptoNote.h
```diff
@@ -124,4 +118,35 @@ inline void from_json(const nlohmann::json &j, CryptoNote::KeyInput &k)
     k.keyImage = j.at("k_image").get<Crypto::KeyImage>();
 }
 
+inline void to_json(nlohmann::json &j, const CryptoNote::RawBlock &block)
```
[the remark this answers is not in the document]
brandonlehmann: Likely, I plan to refactor this later anyways as I want to build out the same type of stuff for the rest of the stuff there if someone doesn't beat me to it.

src/CryptoNoteCore/MainChainStorageSqlite.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+namespace CryptoNote
+{
+    MainChainStorageSqlite::MainChainStorageSqlite(const std::string &blocksFilename, const std::string &indexesFilename)
+    {
+        int resultCode = sqlite3_open(blocksFilename.c_str(), &m_db);
+
+        if (resultCode != SQLITE_OK)
+        {
+            sqlite3_close(m_db);
```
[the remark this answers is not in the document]
brandonlehmann: According to the examples, yes.

src/CryptoNoteCore/MainChainStorageSqlite.cpp
```diff
@@ -0,0 +1,231 @@
...
+            throw std::runtime_error("Failed to prepare getBlockByIndex statement");
+        }
+
+        while((resultCode = sqlite3_step(stmt)) == SQLITE_ROW)
+        {
+            const std::string rawBlockString = std::string(reinterpret_cast<const char*>(sqlite3_column_text(stmt, 0)));
```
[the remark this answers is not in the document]
brandonlehmann:
```
const unsigned char *sqlite3_column_text(sqlite3_stmt*, int iCol);
const void *sqlite3_column_text16(sqlite3_stmt*, int iCol);
```

Do you have a better preference?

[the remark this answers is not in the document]
zpalmtree: Nope, that's fine.

src/Daemon/Daemon.cpp
brandonlehmann: Using `0` would wipe out the genesis, and while that's the same as a resync, personally I'd prefer that they *knowingly* are forcing a resync -- hence the separation of the options.

zpalmtree: Can we provide an error message then, if that's your preference?

src/Daemon/DaemonConfiguration.cpp
brandonlehmann: This follows the same format as the rest. Altering this would trigger further requests for refactoring outside the scope of this PR :)

zpalmtree: Meh... we can fix that later if needed, I don't think we should add more complex code just because it's the style of the file

src/Daemon/Daemon.cpp
brandonlehmann: The only way I can think of to handle that is to carry around another bool variable that triggers to say "yes, rewind" when `--rewind` is found in the command line arguments. Then we have to check for that, I'd consider that unnecessary complexity that the `--help` explains

zpalmtree: Can't we just assert it here: https://github.com/turtlecoin/turtlecoin/pull/769/files#diff-4ec476f1f17243e736b8e16f2c98082eR120

```
if (config.rewindHeight == 0) { // something }`
```

brandonlehmann: See, now that makes more sense.

src/CryptoNoteCore/MainChainStorageSqlite.cpp
```diff
@@ -0,0 +1,231 @@
...
+        {
+            sqlite3_close(m_db);
+            throw std::runtime_error("Failed to pop the last block off the database");
+        }
+    }
+
+    RawBlock MainChainStorageSqlite::getBlockByIndex(uint32_t index) const
+    {
+        sqlite3_stmt *stmt;
+        RawBlock rawBlock = RawBlock();
```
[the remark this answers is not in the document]
brandonlehmann: k

src/CryptoNoteCore/MainChainStorageSqlite.cpp
brandonlehmann: Nope:

```
error C2280: 'std::unique_ptr<CryptoNote::IMainChainStorage,std::default_delete<_Ty>>::unique_ptr(const std::unique_ptr<_Ty,std::default_delete<_Ty>> &)': attempting to reference a deleted function```

File: src/Daemon/DaemonConfiguration.cpp (hunk 3 above)
Comment (brandonlehmann) [reply src/Daemon/DaemonConfiguration.cpp]: It was only 5 instances of it, I replaced em.

File: src/CryptoNoteCore/MainChainStorageSqlite.cpp (hunk 1 above)
Comment (zpalmtree) [reply src/CryptoNoteCore/MainChainStorageSqlite.cpp]: Whoops, should be 

```
auto storage = std::make_unique<MainChainStorageSqlite>(blocksFilename.string() + ".sqlite3", indexesFilename.string());
```

That compiles on linux at least. Also gives a much more informative error message...

Comment (brandonlehmann) [reply src/CryptoNoteCore/MainChainStorageSqlite.cpp]: Acknowledged. It's good.

File: src/Daemon/Daemon.cpp (hunk 2 above)
Comment (brandonlehmann) [reply src/Daemon/Daemon.cpp]: Handled

File: src/Daemon/DaemonConfiguration.cpp [hunk 8]
```diff
@@ -111,6 +114,21 @@ namespace DaemonConfig{
...
+        uint32_t rewindHeight = cli["rewind"].as<uint32_t>();
+        if (rewindHeight == 0)
+        {
+          std::cout << CryptoNote::getProjectCLIHeader()
+            << "Please use the `--resync` option instead of `--rewind 0` to completely reset the synchronization state." << std::endl;
+          exit(1);
+        }
+        else
+        {
+          config.rewindToHeight = rewindHeight;
```
zpalmtree: I'm getting `2019-Apr-06 18:59:43.035292 ERROR   Exception: Deserialization error` after using `--rewind`, saving, and reopening.

To reproduce:

* Launch with `--sqlite`, sync a few blocks (I did about 1000)
* Launch with `--sqlite --resync 300`
* Let that complete then close the daemon (I synced to about 1300 in that short amount of time)
* Reopen with `--sqlite`
* Should see
```
2019-Apr-06 18:59:42.990049 INFO    Loading Checkpoints for faster initial sync...
2019-Apr-06 18:59:42.990204 INFO    Loaded 255 default checkpoints
2019-Apr-06 18:59:42.990308 INFO    Opening DB in a/DB
2019-Apr-06 18:59:43.010648 INFO    DB opened in a/DB
2019-Apr-06 18:59:43.011054 INFO    Initializing core...
2019-Apr-06 18:59:43.012789 INFO    Closing DB.
2019-Apr-06 18:59:43.035292 ERROR   Exception: Deserialization error
```

brandonlehmann: Does it do the same without using SQLite (**experimental**)?

src/Daemon/Daemon.cpp
```diff
@@ -186,6 +189,19 @@ int main(int argc, char* argv[])
     }
   }
 
+  /* If we were given the resync arg, we're deleting everything */
+  if (config.resync)
+  {
+    std::error_code ec;
+    fs::remove_all(fs::path(config.dataDirectory), ec);
```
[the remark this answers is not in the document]
brandonlehmann: It does say it deletes everything. :)

src/Daemon/DaemonConfiguration.cpp
zpalmtree: It does not.

src/Daemon/Daemon.cpp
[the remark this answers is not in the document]
zpalmtree: Let me know how well people reading instructions goes ;)

src/Daemon/DaemonConfiguration.cpp
brandonlehmann: Please define "close the daemon" in step 3.

zpalmtree: Typing `exit`.

brandonlehmann: Did you keep a copy of the .sqlite3 file by chance?

zpalmtree: I could reproduce it 100% so I can make one now if you like

brandonlehmann: Please do, I can't seem to reproduce myself

src/Daemon/Daemon.cpp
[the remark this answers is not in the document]
brandonlehmann: Aye. I'm working a change now to delete just the files we want to delete.

src/Daemon/DaemonConfiguration.cpp
zpalmtree: https://files.catbox.moe/mwfgrc.sqlite3

zpalmtree: Reproducing this time, I saw
```
2019-Apr-07 17:28:31.477007 INFO    Blockchain storage and root segment are on different chains. Cutting root segment to common block index
```
but didn't see that in previous fails, so possibly unrelated

zpalmtree: If I relaunch with `--rewind 300` again it will continue syncing fine, but `--rewind 301` would not work, so I assume the ~300 block is messed up.

brandonlehmann: Aye, that's "normal" if you happened to rewind to a point where there was a fork in the chain :) At least we know that's handled well.

brandonlehmann: Data in your sqlite3 file is well-formed, opened without errors and the block data is fine in the DB. I don't think this file contains the error you mentioned originally.

src/Daemon/Daemon.cpp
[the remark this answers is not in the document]
brandonlehmann: See if the last commit I pushed does it for ya

src/Daemon/DaemonConfiguration.cpp
zpalmtree: It is opening the DB folder which is failing, not the sqlite file.

zpalmtree: Here - https://github.com/turtlecoin/turtlecoin/blob/master/src/CryptoNoteCore/Core.cpp#L114

brandonlehmann: If I’m following the code right, it’s not that it’s opening the rocksdb there, that’s where it’s importing from the local cache (mainChainStorage) and it’s throwing when the data is malformed — as if what was in the SQLite DB was improperly saved. To confirm, that exact error needs replicated and the sqlitedb file checked to see if maybe the data was corrupted during the save. The file you provided above didn’t show this issue so maybe another pass would help?

I suppose, if all else fails, we could check the last few blocks on open to confirm they deserialize correctly and if not pop them off automatically.

zpalmtree: Continuing to follow the code through, that's getting called from here:
https://github.com/turtlecoin/turtlecoin/blob/master/src/CryptoNoteCore/Core.cpp#L2060

It seems to be that for some reason the mainchain storage is getting corrupted after performing a rewind. It is not at the rewind block that it is failing, but rather the top block of the storage that is failing to deserialise.

brandonlehmann: Aye. That’s what I’m saying. If we ch ci the last ones and pop them off if bad crisis averted.

src/Daemon/Daemon.cpp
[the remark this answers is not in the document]
zpalmtree: Works well

src/Daemon/DaemonConfiguration.cpp
brandonlehmann: > Continuing to follow the code through, that's getting called from here:
> https://github.com/turtlecoin/turtlecoin/blob/master/src/CryptoNoteCore/Core.cpp#L2060
> 
> It seems to be that for some reason the mainchain storage is getting corrupted after performing a rewind. It is not at the rewind block that it is failing, but rather the top block of the storage that is failing to deserialise.

Aye, that one is fine :)

brandonlehmann: I think I’ve got it fixed, running some tests now before pushing it up.

brandonlehmann: Pushed. I'm not seeing the gap created by the autoincrement after rewinding now :)

zpalmtree: Works fine

Issue #708, opened 2019-01-28
Prune spent inputs after some period of time from WalletBackend


zpalmtree:
Currently we keep all spent inputs forever, so in case the chain forks, we can revert them to unspent (if they are >= forkHeight). However, there's a point at which the chain is no longer going to fork, say 5000 blocks.

Removing old inputs will reduce the file size significantly, and help keep querying speedy.

I do wonder the best way to implement this. There's not really much point checking every block, since it's a waste of cycles to check and remove a couple of inputs each time. Adding a new thread and checking every N minutes wouldn't hurt.

I think the best way to do it would be to check every N blocks. To keep it simple, probably do it on the same thread as the [findTransactionsInBlocks](https://github.com/turtlecoin/turtlecoin/blob/development/src/WalletBackend/WalletSynchronizer.cpp#L451)  function.

This should be pretty easy to implement for someone who wants a relatively easy task to complete.

zpalmtree:
For inspiration... https://github.com/turtlecoin/turtlecoin-wallet-backend-js/commit/9890ebb16cdce301e0148536a55571c620bf0fae

PR #773, opened 2019-04-11
Prune spent inputs

zpalmtree:
Resolves: #708


src/WalletBackend/Constants.h
```diff
@@ -45,6 +45,10 @@ namespace Constants
        blocks */
     const uint32_t BLOCK_HASH_CHECKPOINTS_INTERVAL = 5000;
 
+    /* The amount of blocks since an input has been spent that we remove it
+       from the container */
+    const uint64_t PRUNE_SPENT_INPUTS_INTERVAL = 5000;
```
brandonlehmann: There’s 2,880 blocks in a day so I’m not sure where this arbitrary 5,000 blocks came from. That being said, I think I would rather see this calculated based on the number of blocks in a day.

zpalmtree: fixed

PR #776, opened 2019-04-12
Prioritizes Regular Transactions over Fusion Transactions

### LINKED ISSUE #767: Fusions being prioritized over regular transactions
As the title says, fusions (zero fee) are being included over transactions with a fee. The relevant chunk of code is here:

https://github.com/turtlecoin/turtlecoin/blob/development/src/CryptoNoteCore/Core.cpp#L2499

It could also do with being cleaned up a bit, the logic is a little hard to follow.

brandonlehmann:
And cleans up the code a little bit.

Resolves #767

### DIFF
--- a/src/CryptoNoteCore/Core.cpp
+++ b/src/CryptoNoteCore/Core.cpp
@@ -57,25 +57,6 @@ std::vector<T> preallocateVector(size_t elements) {
 }
 UseGenesis addGenesisBlock = UseGenesis(true);
 
-class TransactionSpentInputsChecker {
-public:
-  bool haveSpentInputs(const Transaction& transaction) {
-    for (const auto& input : transaction.inputs) {
-      if (input.type() == typeid(KeyInput)) {
-        auto inserted = alreadSpentKeyImages.insert(boost::get<KeyInput>(input).keyImage);
-        if (!inserted.second) {
-          return true;
-        }
-      }
-    }
-
-    return false;
-  }
-
-private:
-  std::unordered_set<Crypto::KeyImage> alreadSpentKeyImages;
-};
-
 inline IBlockchainCache* findIndexInChain(IBlockchainCache* blockSegment, const Crypto::Hash& blockHash) {
   assert(blockSegment != nullptr);
   while (blockSegment != nullptr) {
@@ -2481,6 +2462,43 @@ bool Core::validateBlockTemplateTransaction(
     return true;
 }
 
+bool Core::addTransactionToBlockTemplate(
+    TransactionSpentInputsChecker &spentInputsChecker,
+    const CachedTransaction transaction,
+    const size_t maxTotalSize,
+    const uint64_t height,
+    BlockTemplate &block,
+    size_t &transactionsSize,
+    uint64_t &fee) const
+{
+    if (maxTotalSize < transactionsSize + transaction.getTransactionBinaryArray().size())
+    {
+        return false;
+    }
+
+    if (!validateBlockTemplateTransaction(transaction, height))
+    {
+        transactionPool->removeTransaction(transaction.getTransactionHash());
+
+        return false;
+    }
+
+    if (!spentInputsChecker.haveSpentInputs(transaction.getTransaction()))
+    {
+        transactionsSize += transaction.getTransactionBinaryArray().size();
+
+        fee += transaction.getTransactionFee();
+
+        block.transactionHashes.emplace_back(transaction.getTransactionHash());
+
+        return true;
+    }
+    else
+    {
+        return false;
+    }
+}
+
 void Core::fillBlockTemplate(
     BlockTemplate& block,
     const size_t medianSize,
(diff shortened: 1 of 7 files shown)


src/CryptoNoteCore/TransactionPool.cpp
```diff
@@ -141,6 +129,36 @@ std::vector<CachedTransaction> TransactionPool::getPoolTransactions() const {
   return result;
 }
 
+std::vector<CachedTransaction> TransactionPool::getNormalTransactions() const
```
zpalmtree: Maybe we could make one function, that returns both the normal and cached transactions, to avoid iterating the vector twice? Could make two functions from that one function, if preferred..

src/CryptoNoteCore/Core.h
```diff
@@ -31,6 +31,25 @@
 
 namespace CryptoNote {
 
+class TransactionSpentInputsChecker {
```
zpalmtree: Why are we moving this into the header? I prefer class implementations in the *.cpp (preferably in another file..)

src/CryptoNoteCore/Core.cpp
```diff
@@ -2481,6 +2462,43 @@ bool Core::validateBlockTemplateTransaction(
...
+bool Core::addTransactionToBlockTemplate(
+    TransactionSpentInputsChecker &spentInputsChecker,
+    const CachedTransaction transaction,
+    const size_t maxTotalSize,
+    const uint64_t height,
+    BlockTemplate &block,
+    size_t &transactionsSize,
+    uint64_t &fee) const
+{
+    if (maxTotalSize < transactionsSize + transaction.getTransactionBinaryArray().size())
```
zpalmtree: A little bit of love can make this line easier to understand with some renames.. -
```cpp
/* Not enough room left in the block for this transaction */
if (currentTransactionsSize + transaction.getTransactionBinaryArray().size() > maxTransactionsSize)
```

src/CryptoNoteCore/Core.cpp
```diff
@@ -2481,6 +2462,43 @@ bool Core::validateBlockTemplateTransaction(
...
+    }
+
+    if (!validateBlockTemplateTransaction(transaction, height))
+    {
+        transactionPool->removeTransaction(transaction.getTransactionHash());
+
+        return false;
+    }
+
+    if (!spentInputsChecker.haveSpentInputs(transaction.getTransaction()))
```
zpalmtree: Should we be removing the transaction if it's a double spend?

src/CryptoNoteCore/Core.cpp
```diff
@@ -2489,58 +2507,38 @@ void Core::fillBlockTemplate(
...
+    TransactionSpentInputsChecker spentInputsChecker;
 
-    auto transactionBlobSize = transaction.getTransactionBinaryArray().size();
-    if (currency.fusionTxMaxSize() < transactionsSize + transactionBlobSize) {
-      continue;
-    }
+    std::vector<CachedTransaction> normalTransactions = transactionPool->getNormalTransactions();
 
-    if (!validateBlockTemplateTransaction(transaction, height))
+    for (auto &transaction : normalTransactions)
```
zpalmtree: I think this can be a const iterator.

src/CryptoNoteCore/Core.cpp
```diff
@@ -2481,6 +2462,43 @@ bool Core::validateBlockTemplateTransaction(
     return true;
 }
 
+bool Core::addTransactionToBlockTemplate(
+    TransactionSpentInputsChecker &spentInputsChecker,
+    const CachedTransaction transaction,
```
zpalmtree: Any issue with us taking the transaction by reference here?

src/CryptoNoteCore/Core.cpp
```diff
@@ -2489,58 +2507,38 @@ void Core::fillBlockTemplate(
...
-        continue;
-    }
-
-    if (!spentInputsChecker.haveSpentInputs(transaction.getTransaction())) {
-      block.transactionHashes.emplace_back(transaction.getTransactionHash());
-      transactionsSize += transactionBlobSize;
-      logger(Logging::TRACE) << "Fusion transaction " << transaction.getTransactionHash() << " included to block template";
+        if (addTransactionToBlockTemplate(spentInputsChecker, transaction, maxTotalSize, height, block, transactionsSize, fee))
+        {
+            logger(Logging::TRACE) << "Transaction " << transaction.getTransactionHash() << " included to block template";
```
zpalmtree: Grammar on these two logs is a little off.. maybe 'included in block template' and 'not included in block template'?

src/CryptoNoteCore/Core.cpp
zpalmtree: I'm not sure if there's a better way to handle this. We iterate through every transaction, even once the block is full. Of course, it's possible we try and add a transaction, and fail, and then try and add a smaller transaction and succeed, but at some point, we're going to fail to add any transactions, but keep iterating.

Plus, it looks like we're validating each one when we add it, which is going to add up if we're frequently polling for the block template.

Ideally, we would have some fancy algorithm which sorts by fee per byte, then does some sort of best fit algorithm, but that might want to be split into another PR.

src/CryptoNoteCore/Core.cpp
```diff
@@ -2481,6 +2462,43 @@ bool Core::validateBlockTemplateTransaction(
     return true;
 }
 
+bool Core::addTransactionToBlockTemplate(
```
zpalmtree: meh.... not a big fan of taking multiple params by reference and manipulating them all. Maybe an object to pass in and return? `[bool addedToBlock, BlockTemplateState blockTemplateState]`?

src/CryptoNoteCore/Core.h
```diff
@@ -31,6 +31,25 @@
...
+          return true;
+        }
+      }
+    }
+
+    return false;
+  }
+
+private:
+  std::unordered_set<Crypto::KeyImage> alreadSpentKeyImages;
```
zpalmtree: Could we fix this typo?

src/CryptoNoteCore/TransactionPool.cpp
brandonlehmann: I agree. I thought about this after the fact.

src/CryptoNoteCore/Core.h
brandonlehmann: As it stands right now, `TransactionSpentInputsChecker` is an argument for the added method and thus, the header needs to know about it :)

src/CryptoNoteCore/Core.cpp
brandonlehmann: Not here, no as it's referring only to the `TransactionSpentInputChecker` that's created in this method. This is only checking to make sure that you aren't double spending within the same block. If; however, you attempt to double spend elsewhere, the `validateBlockTemplateTransaction` call will remove it from the pool.

src/CryptoNoteCore/Core.cpp
brandonlehmann: Could easily return once the block is full; however, unless every single byte is used, it isn't really "full" so there's always a chance :)

That being said, the method is short-circuited to return if the transaction won't fit so it's not actually checking (validating) it every time. The size check is first particularly for this reason.

As for the fancy algorithm... see https://github.com/turtlecoin/turtlecoin/blob/81e56e079666b3e93a1e172ac8ae31842eb04af4/src/CryptoNoteCore/TransactionPool.cpp#L27

src/CryptoNoteCore/Core.cpp
brandonlehmann: Good, because I don't even want this here :)

src/CryptoNoteCore/Core.cpp
brandonlehmann: Yup. I've deleted the method :)

src/CryptoNoteCore/Core.cpp
brandonlehmann: Flipped the comparison so it reads like a human and provided comments.

PR #777, opened 2019-04-12
Check transaction input (outputs) unlock times against the last block height @ 1.6M

### LINKED ISSUE #749: Timestamp Locked Input Transactions
Transactions with timestamp encoded unlock can be abused to force block creation that is only valid for a partial group of the chain based on their local timestamp.

### Expected Behavior

A block will never contain a transaction which unlocks after the timestamp of the block.

### Current Behavior

The transactions included are based on the local timestamp of the daemon instance.

### Steps to Reproduce
1. Send a lot of small transactions using locked outputs that are close to the current time.

### Possible Solution

Check the transaction unlock timestamp in the fill block template routine against the chosen timestamp that is used for the template. Check if a timestamp locked input is unlocked in a block using the timestamp of that block.

### Explanation

If a block contains transactions with inputs locked on timestamp base the unlock is validated against the local timestamp. But this timestamp is not necessarily the same on all daemons, thus if chosen toxic you can let people mine blocks that are only partially valid in the blockchain.

brandonlehmann:
Resolves #749

### DIFF
--- a/src/CryptoNoteCore/BlockchainCache.cpp
+++ b/src/CryptoNoteCore/BlockchainCache.cpp
@@ -809,6 +809,18 @@ bool BlockchainCache::isTransactionSpendTimeUnlocked(uint64_t unlockTime, uint32
     return blockIndex + currency.lockedTxAllowedDeltaBlocks() >= unlockTime;
   }
 
+  if (blockIndex >= CryptoNote::parameters::TRANSACTION_INPUT_BLOCKTIME_VALIDATION_HEIGHT)
+  {
+    /* Get the last block timestamp from an existing method call */
+    const std::vector<uint64_t> lastBlockTimestamps = getLastTimestamps(1);
+
+    /* Pop the last timestamp off the vector */
+    const uint64_t lastBlockTimestamp = lastBlockTimestamps.at(0);
+
+    /* Compare our delta seconds plus our last time stamp against the unlock time */
+    return lastBlockTimestamp + currency.lockedTxAllowedDeltaSeconds() >= unlockTime;
+  }
+
   // interpret as time
   return static_cast<uint64_t>(time(nullptr)) + currency.lockedTxAllowedDeltaSeconds() >= unlockTime;
 }
--- a/src/CryptoNoteCore/Core.cpp
+++ b/src/CryptoNoteCore/Core.cpp
@@ -1256,12 +1256,17 @@ void Core::actualizePoolTransactionsLite(const TransactionValidatorState& valida
   auto& pool = *transactionPool;
   auto hashes = pool.getTransactionHashes();
 
+  TransactionValidatorState validator = validatorState;
+
   for (auto& hash : hashes) {
     auto tx = pool.getTransaction(hash);
 
     auto txState = extractSpentOutputs(tx);
 
-    if (hasIntersections(validatorState, txState) || tx.getTransactionBinaryArray().size() > getMaximumTransactionAllowedSize(blockMedianSize, currency)) {
+    if (hasIntersections(validatorState, txState) ||
+        tx.getTransactionBinaryArray().size() > getMaximumTransactionAllowedSize(blockMedianSize, currency) ||
+        !isTransactionValidForPool(tx, validator))
+    {
       pool.removeTransaction(hash);
       notifyObservers(makeDelTransactionMessage({ hash }, Messages::DeleteTransaction::Reason::NotActual));
     }
--- a/src/CryptoNoteCore/DatabaseBlockchainCache.cpp
+++ b/src/CryptoNoteCore/DatabaseBlockchainCache.cpp
@@ -959,11 +959,22 @@ bool DatabaseBlockchainCache::isTransactionSpendTimeUnlocked(uint64_t unlockTime
   return isTransactionSpendTimeUnlocked(unlockTime, getTopBlockIndex());
 }
 
-// TODO: pass time
 bool DatabaseBlockchainCache::isTransactionSpendTimeUnlocked(uint64_t unlockTime, uint32_t blockIndex) const {
   if (unlockTime < currency.maxBlockHeight()) {
     // interpret as block index
     return blockIndex + currency.lockedTxAllowedDeltaBlocks() >= unlockTime;
+  }
+
+  if (blockIndex >= CryptoNote::parameters::TRANSACTION_INPUT_BLOCKTIME_VALIDATION_HEIGHT)
+  {
+    /* Get the last block timestamp from an existing method call */
+    const std::vector<uint64_t> lastBlockTimestamps = getLastTimestamps(1);
+
+    /* Pop the last timestamp off the vector */
+    const uint64_t lastBlockTimestamp = lastBlockTimestamps.at(0);
+
+    /* Compare our delta seconds plus our last time stamp against the unlock time */
+    return lastBlockTimestamp + currency.lockedTxAllowedDeltaSeconds() >= unlockTime;
   }
 
   // interpret as time
--- a/src/config/CryptoNoteConfig.h
+++ b/src/config/CryptoNoteConfig.h
@@ -140,6 +140,7 @@ const uint64_t MAX_EXTRA_SIZE_V2_HEIGHT                      = 1300000;
    and other possible unexpected behavior */
 const uint64_t TRANSACTION_SIGNATURE_COUNT_VALIDATION_HEIGHT = 1400000;
 const uint64_t BLOCK_BLOB_SHUFFLE_CHECK_HEIGHT               = 1600000;
+const uint64_t TRANSACTION_INPUT_BLOCKTIME_VALIDATION_HEIGHT = 1600000;
 
 /* This describes how many blocks of "wiggle" room transactions have regarding
    when the outputs can be spent based on a reasonable belief that the outputs


brandonlehmann:
> I'm not seeing this method getting called in `validateBlockTemplateTransaction`. It's a little dirty to get it working in there, so I think a simpler solution is simply including the future time limit in the unlocked transaction calculation.
> 
> Since a following block cannot have a timestamp less than the current block timestamp - future block time limit, this means that once a timestamp unlocked transaction is in the pool, it is valid, and we don't need to recheck on adding to the template.

I'm not 100% sure what you're trying to say here...

## Add Transaction to Block

1) Calls validateTransaction https://github.com/turtlecoin/turtlecoin/blob/87b9ddca0ca34182db18d0aa18c379a8cc1403d9/src/CryptoNoteCore/Core.cpp#L1099
2) Calls `cache->extractKeyOutputKeys` https://github.com/turtlecoin/turtlecoin/blob/87b9ddca0ca34182db18d0aa18c379a8cc1403d9/src/CryptoNoteCore/Core.cpp#L1848
3) Calls `cache->isTransactionSpendTimeUnlocked`
https://github.com/turtlecoin/turtlecoin/blob/87b9ddca0ca34182db18d0aa18c379a8cc1403d9/src/CryptoNoteCore/DatabaseBlockchainCache.cpp#L984
4) We've altered to compare against the last block time versus local current time

## Add Transaction To Pool

1) Calls validateTransaction https://github.com/turtlecoin/turtlecoin/blob/87b9ddca0ca34182db18d0aa18c379a8cc1403d9/src/CryptoNoteCore/Core.cpp#L1512
2) Calls `cache->extractKeyOutputKeys` https://github.com/turtlecoin/turtlecoin/blob/87b9ddca0ca34182db18d0aa18c379a8cc1403d9/src/CryptoNoteCore/Core.cpp#L1848
3) Calls `cache->isTransactionSpendTimeUnlocked` https://github.com/turtlecoin/turtlecoin/blob/87b9ddca0ca34182db18d0aa18c379a8cc1403d9/src/CryptoNoteCore/DatabaseBlockchainCache.cpp#L984
4) We've altered to compare against the last block time versus local current time

### .

I'm not sure how or why you believe FTL comes into play here. We already have the DELTA blocks defined in the config as 1 block that defines how *far in the future* we're willing to accept transactions with inputs that are *almost* unlocked (based on a locked **time**). This delta is to account for **user experience** as if there is no delta, my local wallet time may lead me to believe that my funds are *spendable* but are actually not by a window of up to 30s. Remember, we're not talking about funds that unlock at a specific block height here, they unlock at a specific *time*. 

We've also checked the transaction as it enters the local pool and we are checking the transactions when they come in with a new block. There is **absolutely no** reason to incur the extra overhead again to hit the local database again during blocktemplate creation to go and check that all of the inputs are unlocked again after we've already done it as the transactions come into our local pool. That's adding code and processes that serve no purpose. If, somehow, someone manages to include transactions in a **mined** block that contains transactions with inputs that are not unlocked yet, **consensus** will reject that block because we're checking for such in the necessary `addBlock` calls.

zpalmtree:
The issue is that an attacker can prevent valid blocks being mined by fiddling with block timestamps (and this may happen accidentally anyway).

Say the last block timestamp was 10:00:00 am.

Our attacker sends transaction(s) using inputs which unlock at 9:59:00am.

Last blocktimestamp + allowed delta = 10:00:30am which is > unlock time, so we add it to the pool.

Then, the attacker makes a block with a timestamp of 9:58:00am which is a diff of 120 seconds, < FTL = 180 seconds so the block is valid.

Now, if we mine a block, and include the previously valid transaction in our block, other nodes will not validate it, since last blocktimestamp (9:58:00) + allowed delta = 9:58:30am which is not > unlock time.

Now, until that transaction is dropped from the pool, no valid blocks will be made which include that transaction, and the chain might freeze, allowing the attacker to build up a large alt chain meanwhile.

You mentioned previously that every transaction in the pool is revalidated when a new block comes in. This would avoid this, however, it felt like that was incidental and we wanted a better way to handle it.

Thinking about it again I'm not sure if using FTL would fix the issue actually.

brandonlehmann:
> The issue is that an attacker can prevent valid blocks being mined by fiddling with block timestamps (and this may happen accidentally anyway).
> 
> Say the last block timestamp was 10:00:00 am.
> 
> Our attacker sends transaction(s) using inputs which unlock at 9:59:00am.
> 
> Last blocktimestamp + allowed delta = 10:00:30am which is > unlock time, so we add it to the pool.
> 
> Then, the attacker makes a block with a timestamp of 9:58:00am which is a diff of 120 seconds, < FTL = 180 seconds so the block is valid.
> 
> Now, if we mine a block, and include the previously valid transaction in our block, other nodes will not validate it, since last blocktimestamp (9:58:00) + allowed delta = 9:58:30am which is not > unlock time.
> 
> Now, until that transaction is dropped from the pool, no valid blocks will be made which include that transaction, and the chain might freeze, allowing the attacker to build up a large alt chain meanwhile.
> 
> You mentioned previously that every transaction in the pool is revalidated when a new block comes in. This would avoid this, however, it felt like that was incidental and we wanted a better way to handle it.
> 
> Thinking about it again I'm not sure if using FTL would fix the issue actually.

As discussed on Discord, I've made a small change to make sure that all transactions in the transaction pool are validated upon block add. This will help prevent a transaction from getting stuck in the pool as explained here without having to touch fillBlockTemplate.

Still up for discussion is whether or not validating the transactions in the pool actually holds up the RPC or not as whether we clear the pool and slowly add them back in as they are validated or loop through the pool, remove them, and attempt to re-add them are the same in the thread process. There would be no change to how the RPC would respond as we're still blocking the same thread either way.

bobbieltd:
Sorry to put in this commit. I have this error with latest code.
2019-Apr-29 00:29:22.228111 INFO    Starting p2p net loop...
2019-Apr-29 00:29:22.228217 INFO    Starting node_server
2019-Apr-29 00:29:28.414368 INFO    [80.211.84.120:11897 OUT] Your TurtleCoin node is syncing with the network (99.99% complete) You are 2 blocks (0 days) behind the current peer you're connected to. Slow and steady wins the race!
2019-Apr-29 00:29:28.414570 INFO    New Top Block Detected: 1447680
2019-Apr-29 00:29:30.182655 INFO    [80.211.84.120:11897 OUT] Successfully synchronized with the TurtleCoin Network.
2019-Apr-29 00:29:30.182832 INFO
2019-Apr-29 00:29:30.182940 INFO    ===[ TurtleCoin Tip! ]=============================
2019-Apr-29 00:29:30.183042 INFO     Always exit TurtleCoind and zedwallet with the "exit" command to preserve your chain and wallet data.
2019-Apr-29 00:29:30.183152 INFO     Use the "help" command to see a list of available commands.
2019-Apr-29 00:29:30.183257 INFO     Use the "backup" command in zedwallet to display your keys/seed for restoring a corrupted wallet.
2019-Apr-29 00:29:30.183366 INFO     If you need more assistance, you can contact us for support at http://chat.turtlecoin.lol
2019-Apr-29 00:29:30.183480 INFO    ===================================================
2019-Apr-29 00:29:30.183480 INFO
2019-Apr-29 00:29:30.183721 INFO

████████╗██╗  ██╗██████╗ ████████╗██╗    ██████╗ █████╗ █████╗ ██╗███╗   ██╗
╚══██╔══╝██║  ██║██╔══██╗╚══██╔══╝██║    ██╔═══╝██╔═══╝██╔══██╗██║████╗  ██║
   ██║   ██║  ██║██████╔╝   ██║   ██║    ████╗  ██║    ██║  ██║██║██╔██╗ ██║
   ██║   ██║  ██║██╔══██╗   ██║   ██║    ██╔═╝  ██║    ██║  ██║██║██║╚██╗██║
   ██║   ╚█████╔╝██║  ██║   ██║   ██████╗██████╗╚█████╗╚█████╔╝██║██║ ╚████║
   ╚═╝    ╚════╝ ╚═╝  ╚═╝   ╚═╝   ╚═════╝╚═════╝ ╚════╝ ╚════╝ ╚═╝╚═╝  ╚═══╝

TurtleCoind: /usr/local/src/turtlecoin/src/CryptoNoteCore/Core.cpp:273: virtual uint64_t CryptoNote::Core::getBlockTimestampByIndex(uint32_t) const: Assertion `!(timestamps.size() == 1)' failed.
Aborted (core dumped)

bobbieltd:
It happens because of getBlockTemplate RPC.
Log :
2019-Apr-29 00:45:33.956603 TRACE   JSON-RPC request: {"id":"0","jsonrpc":"2.0","method":"getblocktemplate","params":{"reserve_size":1,"wallet_address":"TRTLxxx....
TurtleCoind: /usr/local/src/turtlecoin/src/CryptoNoteCore/Core.cpp:273: virtual uint64_t CryptoNote::Core::getBlockTimestampByIndex(uint32_t) const: Assertion `!(timestamps.size() == 1)' failed.
Aborted (core dumped)

brandonlehmann:
Please open a [new issue](https://github.com/turtlecoin/turtlecoin/issues/new?template=bug_report.md) with all relevant information. I am unable to replicate this error.

brandonlehmann:
Also, for reference, this code hasn't activated yet as it's waiting for block 1.6M

PR #785, opened 2019-04-23
Support Blockchain cache API in Nigel

### LINKED ISSUE #712: Support Blockchain cache API in Nigel
It would be nice to support the blockchain cache API (https://github.com/TurtlePay/blockchain-cache-api) transparently for wallets.

You can find the documentation here: https://docs.turtlepay.io/blockapi/

We would probably want to make an IDaemon interface, and then instantiate it with either a ConventionalDaemon or a BlockchainCacheAPI.

This is pretty easy to do, and I've done this exact same thing in TypeScript, so you can easily piggy back off the differences:

Typescript interface: <https://github.com/turtlecoin/turtlecoin-wallet-backend-js/blob/master/lib/IDaemon.ts>
Typescript conventional daemon: <https://github.com/turtlecoin/turtlecoin-wallet-backend-js/blob/master/lib/ConventionalDaemon.ts>
Typescript blockchain cache api: <https://github.com/turtlecoin/turtlecoin-wallet-backend-js/blob/master/lib/BlockchainCacheApi.ts>

You might have to slightly tweak the `--remote-node` support, but this should hopefully allow us to sync directly from a blockchain cache API instead of a conventional daemon.

Note that currently /get_transaction_status is unimplemented in the cache api, but it should be in there pretty soon. It's not essential to wallet operation anyway.

## Bonus

For bonus points, extend `KeyOutput` in `WalletBlockInfo` to take an optional `globalIndex`. The blockchain API returns this from `/getWalletSyncData`, whilst a standard daemon does not. If we pass this through to `processTranasctionOutputs`, we can then skip getting the global indexes from the daemon, which will speed up syncing when we have a wallet with a lot of transactions belonging to us.

Relevant code: https://github.com/turtlecoin/turtlecoin/blob/development/src/WalletBackend/WalletSynchronizer.cpp#L261

brandonlehmann:
Resolves #712

To test non-ssl, you can sync against: `node.trtlpay.com:80`
To test SSL (when built with SSL), you can sync against: `blockapi.turtlepay.io:443`

**Bonus:** It auto-detects if it's a blockchain cache so a user doesn't have to do anything special

### DIFF
--- a/include/WalletTypes.h
+++ b/include/WalletTypes.h
@@ -1,5 +1,5 @@
 // Copyright (c) 2018, The TurtleCoin Developers
-// 
+//
 // Please see the included LICENSE file for more information.
 
 #pragma once
@@ -13,6 +13,7 @@
 
 #include <unordered_map>
 #include <optional>
+#include <sstream>
 
 namespace WalletTypes
 {
@@ -27,8 +28,8 @@ namespace WalletTypes
     };
 
     /* A coinbase transaction (i.e., a miner reward, there is one of these in
-       every block). Coinbase transactions have no inputs. 
-       
+       every block). Coinbase transactions have no inputs.
+
        We call this a raw transaction, because it is simply key images and
        amounts */
     struct RawCoinbaseTransaction
@@ -105,7 +106,7 @@ namespace WalletTypes
 
         /* The transaction key we took from the key outputs */
         Crypto::PublicKey key;
-        
+
         /* If spent, what height did we spend it at. Used to remove spent
            transaction inputs once they are sure to not be removed from a
            forked chain. */
@@ -309,8 +310,8 @@ namespace WalletTypes
             /* A map of public keys to amounts, since one transaction can go to
                multiple addresses. These can be positive or negative, for example
                one address might have sent 10,000 TRTL (-10000) to two recipients
-               (+5000), (+5000) 
-               
+               (+5000), (+5000)
+
                All the public keys in this map, are ones that the wallet container
                owns, it won't store amounts belonging to random people */
             std::unordered_map<Crypto::PublicKey, int64_t> transfers;
@@ -495,7 +496,20 @@ namespace WalletTypes
         r.keyOutputs = j.at("outputs").get<std::vector<KeyOutput>>();
         r.hash = j.at("hash").get<Crypto::Hash>();
         r.transactionPublicKey = j.at("txPublicKey").get<Crypto::PublicKey>();
-        r.unlockTime = j.at("unlockTime").get<uint64_t>();
+
+        /* We need to try to get the unlockTime from an integer in the json
+           however, if that fails because we're talking to a blockchain
+           cache API that encodes unlockTime as a string (due to json
+           integer encoding limits), we need to attempt this as a string */
+        try
+        {
+            r.unlockTime = j.at("unlockTime").get<uint64_t>();
+        }
+        catch (...)
+        {
+            std::istringstream unlockTime(j.at("unlockTime").get<std::string>());
+            unlockTime >> r.unlockTime;
+        }
     }
 
     inline void to_json(nlohmann::json &j, const RawTransaction &r)
@@ -515,7 +529,20 @@ namespace WalletTypes
         r.keyOutputs = j.at("outputs").get<std::vector<KeyOutput>>();
         r.hash = j.at("hash").get<Crypto::Hash>();
         r.transactionPublicKey = j.at("txPublicKey").get<Crypto::PublicKey>();
-        r.unlockTime = j.at("unlockTime").get<uint64_t>();
+
+        /* We need to try to get the unlockTime from an integer in the json
+           however, if that fails because we're talking to a blockchain
+           cache API that encodes unlockTime as a string (due to json
+           integer encoding limits), we need to attempt this as a string */
+        try
+        {
+            r.unlockTime = j.at("unlockTime").get<uint64_t>();
+        }
+        catch (...)
+        {
+            std::istringstream unlockTime(j.at("unlockTime").get<std::string>());
+            unlockTime >> r.unlockTime;
+        }
         r.paymentID = j.at("paymentID").get<std::string>();
         r.keyInputs = j.at("inputs").get<std::vector<CryptoNote::KeyInput>>();
     }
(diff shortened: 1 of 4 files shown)


include/WalletTypes.h
```diff
@@ -495,7 +496,20 @@ namespace WalletTypes
...
+
+        /* We need to try to get the unlockTime from an integer in the json
+           however, if that fails because we're talking to a blockchain
+           cache API that encodes unlockTime as a string (due to json
+           integer encoding limits), we need to attempt this as a string */
+        try
+        {
+            r.unlockTime = j.at("unlockTime").get<uint64_t>();
+        }
+        catch (...)
```
zpalmtree: Please catch the specific exception thrown rather than using a catch all.

include/WalletTypes.h
```diff
@@ -495,7 +496,20 @@ namespace WalletTypes
...
+        {
+            std::istringstream unlockTime(j.at("unlockTime").get<std::string>());
```
zpalmtree: Please use `std::stoull` or similar for int conversion. You may wish to catch `std::invalid_argument` if it's possibly not an int, or `std::out_of_range` if this can be negative/exceed uint64_t limits.

src/Nigel/Nigel.cpp
```diff
@@ -107,6 +108,19 @@ std::tuple<bool, std::vector<WalletTypes::WalletBlockInfo>> Nigel::getWalletSync
         {"startTimestamp", startTimestamp}
     };
 
+    /* The blockchain cache does not support sending
+       both the startHeight and startTimestamp properties
+       and will kick a 500 error if we do. So, let's make sure
+       that we're only sending one of them */
+    if (m_isBlockchainCache && startHeight == 0)
```
zpalmtree: Are you sure..? We're sending both here - <https://github.com/turtlecoin/turtlecoin-wallet-backend-js/blob/master/lib/BlockchainCacheApi.ts#L157>

include/WalletTypes.h
```diff
@@ -515,7 +529,20 @@ namespace WalletTypes
...
+
+        /* We need to try to get the unlockTime from an integer in the json
+           however, if that fails because we're talking to a blockchain
+           cache API that encodes unlockTime as a string (due to json
+           integer encoding limits), we need to attempt this as a string */
+        try
+        {
+            r.unlockTime = j.at("unlockTime").get<uint64_t>();
+        }
+        catch (...)
```
zpalmtree: Same comments as above.

src/Nigel/Nigel.cpp
brandonlehmann: Give it a go, it 500s for me.

include/WalletTypes.h
brandonlehmann: I found some article that says that’s slower :) I’ll be happy to revisit. 

As for the catch, sure.

zpalmtree: stringstreams are nearly always the slowest way to do things related to parsing.

src/Nigel/Nigel.cpp
zpalmtree: Works fine here:

```
curl -X POST "https://blockapi.turtlepay.io/getwalletsyncdata" -H  "accept: application/json" -H  "Content-Type: application/json" -d "{\"startTimestamp\":0,\"startHeight\":100,\"blockHashCheckpoints\":[\"0606a15147159e3f01dd90aa78828cf8587caed36203cd764701b56eb6ff6fd8\",\"1d61bf052ee2fa3720078ee01cf30e01ba305096bf3e3d3668fbc09bbeafa244\"]}"
```

It throws a 400 if `startHeight` and `startTimestamp` are both non zero, however - https://github.com/TurtlePay/blockchain-cache-api/blob/master/index.js#L778

A 500 should only be returned if an exception was thrown processing the request - https://github.com/TurtlePay/blockchain-cache-api/blob/master/index.js#L788

brandonlehmann: 400/500 either way it doesn’t work :)

I’ll update the comment to say 400. :smile:

zpalmtree: But it does work provided one is zero. If that's not the case, then that's a bug.

brandonlehmann: Bah.

brandonlehmann: fixed

include/WalletTypes.h
brandonlehmann: caught

include/WalletTypes.h
brandonlehmann: > You may wish to catch `std::invalid_argument` if it's possibly not an int, or `std::out_of_range` if this can be negative/exceed uint64_t limits.

On second thought, I do not wish. What would we do? Catch that error and re-throw it? This method doesn't have an error_code return, it just throws.

include/WalletTypes.h
brandonlehmann: Handled with tighter catch

include/WalletTypes.h
zpalmtree: I'm not sure exactly where this code is being used, but we don't want to stop saving/loading/syncing because a block is funky - we want to replace the garbage data with a reasonable value and carry on regardless.

brandonlehmann: What? It's the `from_json` method -- it already fails if data is malformed and kicks the bucket. The throw is caught further up in the stack here: https://github.com/turtlecoin/turtlecoin/blob/be2d97f14a51bc44c48d5d07f4df9def8b7c4952/src/Nigel/Nigel.cpp#L129

Because of here: https://github.com/turtlecoin/turtlecoin/blob/be2d97f14a51bc44c48d5d07f4df9def8b7c4952/include/WalletTypes.h#L476 and here: https://github.com/turtlecoin/turtlecoin/blob/be2d97f14a51bc44c48d5d07f4df9def8b7c4952/include/WalletTypes.h#L477

zpalmtree: It works fine parsing the `uint64_t` from the current daemon, because that returns the incorrect value, which is a valid `uint64_t`, iirc.

If the cache is returning it as a string to allow values > 2^64, then this line will throw. That will be caught, but as you indicated, this is called in `getWalletSyncData`. The blocks will not be downloaded, and so the code will retry, and the same error will occur. Syncing will get stuck.

brandonlehmann: The cache doesn't make it a string because its > 2^64 (nor because it's a negative value). It makes it a string because javascript (node.js) doesn't support uint64_t as an integer type and thus cannot represent it in json accurately.

https://tools.ietf.org/html/rfc7159#section-6
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

> The MAX_SAFE_INTEGER constant has a value of 9007199254740991 (9,007,199,254,740,991 or ~9 quadrillion). The reasoning behind that number is that JavaScript uses double-precision floating-point format numbers as specified in **IEEE 754 and can only safely represent numbers between -(2^53 - 1) and 25^3 - 1.**

zpalmtree: OK

Issue #714, opened 2019-01-29
Replace all usage of nlohmann::json with rapidjson


zpalmtree:
We recently migrated from nlohmann::json for wallet saving/loading in #706 

Whilst nlohmann::json is very nice to use, it doesn't perform so well with large data sets. It would be nice to replace all usages of nlohmann::json with rapidjson, so we can get rid of another dependency.

It's mainly used in ApiDispatcher, but there are also a few uses in `include/CryptoTypes.h` and make `include/WalletTypes.h`

Edit [@brandonlehmann]: It's also used in turtle-service and the daemon for dumping and/or loading a configuration file.

Macroshock:
Jesus I was replacing all usage of JsonValue with nlohmann::json

Well at least I didnt get very far

zpalmtree:
oof

brandonlehmann:
I feel like an overall win was had here.

brandonlehmann:
> Jesus I was replacing all usage of JsonValue with nlohmann::json
> 
> Well at least I didnt get very far

You, whoever you are, are a saint. While you're at it check out #603 :)

Macroshock:
> You, whoever you are, are a saint. While you're at it check out #603 :)

This is what motivated the removal of JsonValue, the original idea was to implement uint64_t serialization in the JsonValue library but after a while I decided that removing JsonValue was a better idea since nlohmann was already being used in the project.

Well I guess it's time to check this RapidJson thing.

brandonlehmann:
> > You, whoever you are, are a saint. While you're at it check out #603 :)
> 
> This is what motivated the removal of JsonValue, the original idea was to implement uint64_t serialization in the JsonValue library but after a while I decided that removing JsonValue was a better idea since nlohmann was already being used in the project.
> 
> Well I guess it's time to check this RapidJson thing.

Bless your soul weary traveler. Come and be known http://chat.turtlecoin.lol we have cookies

zpalmtree:
Not suggesting we use either of these, just in case anyone is interested in even *more* performance, I found these two:
https://github.com/chadaustin/sajson
https://github.com/lemire/simdjson

brandonlehmann:
Are they actively maintained ?

zpalmtree:
Neither are suitable for our use case, they're just *really* fast. I thought someone might find it interesting. simdjson certainly is - I found it on HN this morning.

brandonlehmann:
Cool

[...]

PR #787, opened 2019-04-24
nlohmann -> rapidjson (daemon+wallet config)

rixombea:
`--dump-config`, `--save-config`, `--config-file` was tested, no regression.

baby steps for: #714

### DIFF
--- a/src/Daemon/DaemonConfiguration.cpp
+++ b/src/Daemon/DaemonConfiguration.cpp
@@ -4,9 +4,14 @@
 
 #include "DaemonConfiguration.h"
 #include <cxxopts.hpp>
-#include <json.hpp>
+
+#include <rapidjson/document.h>
+#include <rapidjson/istreamwrapper.h>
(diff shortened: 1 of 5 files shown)


src/Daemon/DaemonConfiguration.cpp
```diff
@@ -582,191 +587,265 @@ namespace DaemonConfig{
...
+    IStreamWrapper isw(data);
+    
+    Document j;
+    j.ParseStream(isw);
 
-    json j;
-    data >> j;
-
-    if (j.find("data-dir") != j.end())
+    if (j.HasMember("data-dir") )
```
brandonlehmann: I hate to be that guy, but there's a space that snuck in there and I'm getting :triggered:

rixombea: updated

brandonlehmann: Thanks!

PR #789, opened 2019-04-24
Wallet Synchronization Back-off Mechanism

brandonlehmann:
Implemented a back-off mechanism that cuts the number of blocks requested in half during sync operations if something fails. Once it passes, it moves back up to full power again.

### DIFF
--- a/src/Nigel/Nigel.cpp
+++ b/src/Nigel/Nigel.cpp
@@ -76,6 +76,7 @@ void Nigel::swapNode(const std::string daemonHost, const uint16_t daemonPort, co
 {
     stop();
 
+    m_blockCount = CryptoNote::BLOCKS_SYNCHRONIZING_DEFAULT_COUNT;
     m_localDaemonBlockCount = 0;
     m_networkBlockCount = 0;
     m_peerCount = 0;
@@ -89,6 +90,19 @@ void Nigel::swapNode(const std::string daemonHost, const uint16_t daemonPort, co
     m_nodeClient = getClient(m_daemonHost, m_daemonPort, m_daemonSSL, m_timeout);
 
     init();
+}
+
+void Nigel::decreaseRequestedBlockCount()
+{
+    if (m_blockCount > 1)
+    {
+        m_blockCount = m_blockCount / 2;
+    }
+}
+
+void Nigel::resetRequestedBlockCount()
+{
+    m_blockCount = CryptoNote::BLOCKS_SYNCHRONIZING_DEFAULT_COUNT;
 }
 
 std::tuple<bool, std::vector<WalletTypes::WalletBlockInfo>> Nigel::getWalletSyncData(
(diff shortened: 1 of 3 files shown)


src/Nigel/Nigel.cpp
```diff
@@ -91,11 +92,26 @@ void Nigel::swapNode(const std::string daemonHost, const uint16_t daemonPort, co
     init();
 }
 
+void Nigel::decreaseRequestedBlockCount()
+{
+    if (m_blockCount > 1)
+    {
+        m_blockCount = m_blockCount / 2;
```
zpalmtree: Can be `m_blockCount /= 2`

src/Nigel/Nigel.cpp
```diff
@@ -105,7 +121,8 @@ std::tuple<bool, std::vector<WalletTypes::WalletBlockInfo>> Nigel::getWalletSync
     json j = {
         {"blockHashCheckpoints", blockHashCheckpoints},
         {"startHeight", startHeight},
-        {"startTimestamp", startTimestamp}
+        {"startTimestamp", startTimestamp},
+        {"blockCount", blockCount}
```
zpalmtree: We can just use `m_blockCount` here directly instead of using another var.

src/Nigel/Nigel.h
```diff
@@ -115,6 +121,10 @@ class Nigel
         /* If we should stop the background thread */
         std::atomic<bool> m_shouldStop = false;
 
+        /* Stores how many blocks we'll try to sync at a given time
+           and automatically moves up and down if we run into issues */
```
zpalmtree: Just being picky here.. but it's not Nigel that is automatically setting the block count.

src/Nigel/Nigel.cpp
brandonlehmann: That’s not very readable

src/Nigel/Nigel.cpp
brandonlehmann: No, you can’t because std::atomic<uint64_t> doesn’t have a json serializer and I didn’t feel like writing one.

src/Nigel/Nigel.h
brandonlehmann: Yes, that’s you being picky.

src/Nigel/Nigel.cpp
zpalmtree: `m_blockCount.load()`

src/Nigel/Nigel.cpp
zpalmtree: Do you write `i = i + 1` instead of `i++`? I don't see the difference.

src/Nigel/Nigel.cpp
brandonlehmann: Make up your mind :)

src/Nigel/Nigel.cpp
brandonlehmann: There is no `/=` operator for std::atomic (https://en.cppreference.com/w/cpp/atomic/atomic)

zpalmtree: ok

PR #790, opened 2019-04-25
nlohmann -> rapidjson (zedwallet++)

rixombea:
`ab_add`, `ab_list`, `ab_delete` was tested, no regression.

baby steps for: #714

### DIFF
--- a/src/zedwallet++/AddressBook.h
+++ b/src/zedwallet++/AddressBook.h
@@ -2,15 +2,10 @@
 // 
 // Please see the included LICENSE file for more information.
 
-#include "json.hpp"
-
 #include <string>
-
 #include <vector>
-
 #include <WalletBackend/WalletBackend.h>
 
-using nlohmann::json;
 
 struct AddressBookEntry
 {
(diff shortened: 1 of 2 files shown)


src/zedwallet++/AddressBook.h
```diff
@@ -2,15 +2,15 @@
...
+#include "rapidjson/writer.h"
 
 #include <string>
-
 #include <vector>
-
 #include <WalletBackend/WalletBackend.h>
 
-using nlohmann::json;
+using namespace rapidjson;
```
zpalmtree: I'm not a big fan of `using namespace ...` unless it saves a lot of complexity.

src/zedwallet++/AddressBook.h
```diff
@@ -2,15 +2,15 @@
 // 
 // Please see the included LICENSE file for more information.
 
-#include "json.hpp"
+#include <JsonHelper.h>
+#include "rapidjson/document.h"
+#include "rapidjson/writer.h"
```
zpalmtree: Do we need to include this here when we're taking a template in `toJSON`?

src/zedwallet++/AddressBook.h
rixombea: removed

src/zedwallet++/AddressBook.h
rixombea: removed

PR #794, opened 2019-04-26
WalletBackend Scan Height Allowance

brandonlehmann:
Allow the daemon/cache to return blocks prior to the one requested as long as the response contains the block we're looking for

### DIFF
--- a/src/WalletBackend/WalletSynchronizer.cpp
+++ b/src/WalletBackend/WalletSynchronizer.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2018, The TurtleCoin Developers
+// Copyright (c) 2018-2019, The TurtleCoin Developers
 //
 // Please see the included LICENSE file for more information.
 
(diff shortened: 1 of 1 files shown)


src/WalletBackend/WalletSynchronizer.cpp
```diff
@@ -208,19 +208,30 @@ std::vector<WalletTypes::WalletBlockInfo> WalletSynchronizer::downloadBlocks()
...
         /* Only check if a timestamp isn't given */
         if (m_startTimestamp == 0)
         {
-            /* The height we expect to get back from the daemon */
-            if (actualHeight != m_startHeight)
+            bool startHeightFoundInRange = false;
+
+            /* Loop through the blocks we got back and make sure that
+               we were given data for the start block we were looking for */
+            for (const auto &block : blocks)
```
zpalmtree: We could instead do

```cpp
const auto it = std::find_if(blocks.begin(), blocks.end(), [m_startHeight](const auto &block) {
    return block.blockHeight == m_startHeight
});

/* If we weren't given a block with the startHeight we were
   looking for then we don't need to store this data */
if (it == blocks.end())
{
    // etc
}
```

PR #796, opened 2019-04-28
few changes

rixombea:
- replace `boost::{trim, split, filesystem}` with common utility
- replace `boost::variant` w/ `std::variant` (~~ci **FAILED** for osx+clang~~ built with workaround/shim)

possible replacement: #797


zpalmtree:
nice

brandonlehmann:
.tip 1000 @rixombea

rixombea:
:fire:

src/Common/VariantShim.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#if defined(__clang__) && defined(__APPLE__)
+    #include "mpark/variant.hpp"
```
zpalmtree: Was the issue that clang apple just doesn't have std::variant? We might be able to use `__has_include` instead of hardcoding this. See FileSystemShim for inspiration.

rixombea: with `__has_include` it failed, looks like variant header is there, but it was barking on std::visit calls + std::get won't accept variant.

zpalmtree:
While it is quite a small dependency.. I'm not that enthusiastic at adding it just for clang on apple. Thoughts @brandonlehmann ?

brandonlehmann:
TBH I think it's robbing Peter to pay Paul. Pulling out Boost is nice but not if it's replacing it with another crutch. I'm honestly surprised we can't just pull in the header(s) necessary for Apple and link them from the platform directory.

rixombea:
what about using `mpark::variant` on all platform? at least until we can fully utilized stdlib?
It's basically same implementation as libc++ `std::variant` with added support for c++11/14 (also from the same guy who implement `std::variant` in llvm libc++)

And speaking of adding extra crutch, can't see any difference than when we're pulling cxxopts to replace `boost::program_options`.

brandonlehmann:
> what about using `mpark::variant` on all platform? at least until we can fully utilized stdlib?
> It's basically same implementation as libc++ `std::variant` with added support for c++11/14 (also from the same guy who implement `std::variant` in llvm libc++)
> 
> And speaking of adding extra crutch, can't see any difference than when we're pulling cxxopts to replace `boost::program_options`.

🔥 🚑

You win this round sir.

Personally, I'd give it a pass it if it was consistent across the platforms.

rixombea:
:cry:  :musical_note: I surrender, I surrender, I'm giving up the...... :musical_note:  :cry:

PR #805, opened 2019-05-05
zedwallet++: ask for spend key first during restore

thinkpol2:
Small cosmetic change that's been bothering me. When creating a wallet, zedwallet++ displays the view key and then the spend key. When restoring a wallet from keys, it asks for the spend key and then the view key. This changes it to request view key and then spend key.

### DIFF
--- a/src/zedwallet++/CommandImplementations.cpp
+++ b/src/zedwallet++/CommandImplementations.cpp
@@ -65,23 +65,21 @@ void printPrivateKeys(const std::shared_ptr<WalletBackend> walletBackend)
 
     const auto [error, mnemonicSeed] = walletBackend->getMnemonicSeed();
 
+    /* If this isn't a view only wallet, print out the spend key and mnemonic if available */
+    if (!walletBackend->isViewWallet())
+    {
+        std::cout << SuccessMsg("\nPrivate spend key:\n")
+                  << SuccessMsg(privateSpendKey) << "\n";
+
+        if (!error)
+        {
+            std::cout << SuccessMsg("\nMnemonic seed:\n")
+                      << SuccessMsg(mnemonicSeed) << "\n";
+        }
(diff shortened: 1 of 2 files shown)


thinkpol2:
I feel the same, but in looking at the code I see why @zpalmtree swapped it:

https://github.com/turtlecoin/turtlecoin/blob/1889a3d8767ccdaa92c60558f1e8eb35ae159176/src/zedwallet%2B%2B/CommandImplementations.cpp#L72

brandonlehmann:
> I feel the same, but in looking at the code I see why @zpalmtree swapped it:
> 
> https://github.com/turtlecoin/turtlecoin/blob/1889a3d8767ccdaa92c60558f1e8eb35ae159176/src/zedwallet%2B%2B/CommandImplementations.cpp#L72

That's not a good reason IMO :)

```c
void printPrivateKeys(const std::shared_ptr<WalletBackend> walletBackend)
{
    const auto [privateSpendKey, privateViewKey] = walletBackend->getPrimaryAddressPrivateKeys();

    const auto [error, mnemonicSeed] = walletBackend->getMnemonicSeed();

    /* If this isn't a view only wallet, print out the spend key and mnemonic if available */
    if (!walletBackend->isViewWallet())
        {
            std::cout << SuccessMsg("\nPrivate spend key:\n")
                  << SuccessMsg(privateSpendKey) << "\n";

        if (!error)
        {
            std::cout << SuccessMsg("\nMnemonic seed:\n")
                      << SuccessMsg(mnemonicSeed) << "\n";
        }
    }

    std::cout << SuccessMsg("Private view key:\n")
              << SuccessMsg(privateViewKey) << "\n";
}
```

brandonlehmann:
an you know... apply proper formatting because i'm horrible at that :)

PR #812, opened 2019-05-10
Fix cmake not defaulting to release build

zpalmtree:
Also fixes a misuse of assert() in cryptotest which made testing confusing :angry:


wrkzdev:
bravo!

PR #816, opened 2019-05-13
WalletBackend block cache + prefetch

zpalmtree:
Prefetches blocks to increase syncing speed. Sync speed comparison:

**zedwallet, remote node**: 28719 blocks in 5 mins, **96 blocks a second**
**zedwallet, local node**: 39411 blocks in 5 mins, **131 blocks a second**

**zedwallet-beta, remote node**: 18653 blocks in 5 mins, **62 blocks a second**
**zedwallet-beta with block prefetch**: 41249 blocks in 5 mins, **137 blocks a second**
**zedwallet-beta, with block prefetch, local node**: 87923 blocks in 4:42 mins, **312 blocks a second** - ran out of blocks to sync lol

I have not tested *that* intensively. Might be some bugs I missed. Did not crash in my tests, retains sync progress, and can send/receive transactions successfully. Possibly some race conditions/dead locks.

### DIFF
--- a/src/WalletBackend/BlockDownloader.h
+++ b/src/WalletBackend/BlockDownloader.h
@@ -23,6 +23,8 @@
 class BlockDownloader
 {
     public:
+        /* Destructor */
+        ~BlockDownloader();
 
         /////////////////////////////
         /* Public member functions */
@@ -33,19 +35,35 @@ class BlockDownloader
            available. May be empty (this is the norm when synced.) */
         std::vector<WalletTypes::WalletBlockInfo> fetchBlocks(const size_t blockCount) const;
 
+        /* Drops the oldest block from the internal queue */
+        void dropBlock();
+
+        /* Start block downloading process */
+        void start();
+
+        /* Stop block downloading process */
+        void stop();
+
     private:
 
         //////////////////////////////
         /* Private member functions */
         //////////////////////////////
 
-        uint64_t getHeight() const;
+        /* Synchronizes pre-fetching blocks */
+        void downloader();
 
+        /* Determines if we should prefetch more blocks */
+        bool shouldFetchMoreBlocks() const;
+
+        /* Gets checkpoints of stored (not processed) blocks */
         std::vector<Crypto::Hash> getStoredBlockCheckpoints() const;
 
+        /* Gets checkpoints of stored, processed, and infrequent checkpoints */
         std::vector<Crypto::Hash> getBlockCheckpoints() const;
 
-        void downloadBlocks();
+        /* Downloads a set of blocks, if needed */
+        bool downloadBlocks();
 
         //////////////////////////////
         /* Private member variables */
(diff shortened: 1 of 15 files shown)


src/WalletBackend/BlockDownloader.h
```diff
@@ -0,0 +1,130 @@
...
+        /* Sync progress */
+        SynchronizationStatus m_synchronizationStatus;
+
+        std::shared_ptr<SubWallets> m_subWallets;
+
+        /* For synchronizing block downloading */
+        std::mutex m_mutex;
+
+        /* Are we ready to go attempt to retrieve more data */
+        std::atomic<bool> m_goFish = true;
```
brandonlehmann: cute... but maybe a more descriptive variable name is beneficial

src/WalletBackend/SynchronizationStatus.cpp
```diff
@@ -33,6 +33,7 @@ void SynchronizationStatus::storeBlockHash(
                    << m_lastKnownBlockHeight + 1 << ", Received: "
                    << height << ".\nPossibly malicious daemon. Terminating.";
 
+            /* TODO: Convert to log message */
```
brandonlehmann: Please convert to log message :)

zpalmtree: I didn't want to convert it till being sure that my code works -- this is good at catching bugs

src/WalletBackend/BlockDownloader.h
zpalmtree: I think mine was more descriptive honestly.. but renamed.

brandonlehmann: Should it be consume instead of consumed?

zpalmtree: Well.. it's really a way of telling the block downloader that it should go fetch more data - but that name is already given to the condition variable. We fire it sometimes when we're stopping to unlock the thread, or if there are no blocks available. Hence go fish

brandonlehmann: yeah, i got the logic behind goFish :)

`m_consumeData` tells me I should "consume more data"
`m_consumedData` tells me that I have already "consumed more data"

Since you're saying the variable says that I *should* go consume more data -- then it isn't grammatically correct to use past tense, correct?

brandonlehmann: Now that I look again, i see it's flipped to `true` on `stop()` which seems contrary to what I just said and how I read your last message  :(

zpalmtree: It's used to cause the `std::condition_variable` which waits for data to have been consumed, before then attempting to download more blocks, to unlock, and cause the block download attempt to proceed.

However, if we are stopping, then this would wait forever, unless we flip it to true. Because of this, right after unlocking, we check if we're stopping, and if so, break the loop.

brandonlehmann: 👌

zpalmtree:
Is this fine to merge now, or?

PR #819, opened 2019-05-14
Altering default Target Architecture in Build Process

brandonlehmann:
* Sets the default target architecture to `native` instead of `default`
* Alters CI builds to specify a target architecture of `default`
* Removes tons of RocksDB fluff from the project CMakeLists.txt file and has RocksDB look at the `ARCH` flag instead to determine if it needs to run the feature detection(s)
* Updates the README.md to reflect information regarding the difference between `native` and `default` builds

This PR helps prep the code base for Argon2 optimizations coming in the Chukwa branch that benefit highly from `native` builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,12 @@ If you would like to compile yourself, read on.
 
 ### How To Compile
 
+#### Build Optimization
+
+The CMake build system will, by default, create optimized *native* builds for your particuar system type when you build the software. Using this method, the binaries created provide a better experience and all together faster performance.
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -14,6 +14,12 @@ If you would like to compile yourself, read on.
 
 ### How To Compile
 
+#### Build Optimization
+
+The CMake build system will, by default, create optimized *native* builds for your particuar system type when you build the software. Using this method, the binaries created provide a better experience and all together faster performance.
```
zpalmtree:
```suggestion
The CMake build system will, by default, create optimized *native* builds for your particular system type when you build the software. Using this method, the binaries created provide a better experience and all together faster performance.
```

PR #820, opened 2019-05-16
Split slow-hash.c into platform specific files

brandonlehmann:


zpalmtree:
Have you tested this compiles+runs fine on arm? Can you clarify what results in the portable version being used over the optimized version(s)?

brandonlehmann:
> Have you tested this compiles+runs fine on arm? Can you clarify what results in the portable version being used over the optimized version(s)?

Runs on my ARM vm as well as rpi and libre boards. Others would have to test elsewhere.

As for what gets the portable code:

x86 gets: `!defined NO_AES && (defined(__x86_64__) || (defined(_MSC_VER) && defined(_WIN64)))`
* Meaning, `NO_AES` is **not** defined **and** platform is (x86_64 **or** it's MS on a 64-bit machine)

arm gets: `!defined NO_AES && (defined(__arm__) || defined(__aarch64__))`
* Meaning: `NO_AES` is **not** defined **and** the platform is (ARM **or** aarch64)

everything else gets: `!(!defined NO_AES && (defined(__arm__) || defined(__aarch64__))) && !(!defined NO_AES && (defined(__x86_64__) || (defined(_MSC_VER) && defined(_WIN64))))`
* Meaning: If `NO_AES` is set you'll land here regardless of platform
* Also meaning: If you aren't on x86_64 **or** win64 **or** arm **or** aarch64 you'll land here

zpalmtree:
Thank you.

PR #821, opened 2019-05-17
Multithreaded wallet syncing

zpalmtree:
Multi-threads the wallet syncing progress. Threads are configurable via command line option.

* ~~I have encountered a couple of rare crashes with `std::bad_alloc`. Have not managed to track down or have a good method to reproduce, hence marked WIP. If anyone is bored and wants to do some performance testing or debugging and could figure out how to reproduce it, that would be grand.~~

* ~~I don't have any performance numbers right now.~~ It was not much faster than with block prefetch when I tested, due to being bottlenecked by the daemon. Planning on looking for some performance improvements there, but do not expect to find many. ~~I also plan on doing a full sync from zero comparison, but the daemon is not playing ball right now.~~

* This should help lower powered devices which are not bottlenecked on the daemon so much (for example, arm devices and phones) to sync with more speed by utilizing their threads more effectively.

| Program            | Node Type         | Block Pre Fetch      | Time To Sync From Zero          | Blocks Per Second | Threads |
|:------------------:|:-----------------:|:--------------------:|:-------------------------------:|:-----------------:|:--------|
| zedwallet-beta     | Local             | Yes                  | 57:40 minutes                   | 440               | 12      |
| zedwallet-beta     | Local             | Yes                  | 108:36 minutes                  | 232               | 1       |
| C++ on Phone     | Cache            | Yes                  | 280:44 minutes                 | 89                 | 1        |
| C++ on Phone     | Cache            | Yes                  | 275:43 minutes                 | 91                 | 8        |
| zedwallet             | Local             | Unknown          | 85:00 minutes                  | 295               | Unknown       |

### DIFF
--- a/src/WalletApi/ParseArguments.cpp
+++ b/src/WalletApi/ParseArguments.cpp
@@ -30,10 +30,9 @@ Config parseArguments(int argc, char **argv)
 
     options.add_options("Core")
         ("h,help", "Display this help message", cxxopts::value<bool>(help)->implicit_value("true"))
-
-        ("v,version", "Output software version information", cxxopts::value<bool>(version)->default_value("false")->implicit_value("true"))
         ("log-level", "Specify log level", cxxopts::value<int>(logLevel)->default_value(std::to_string(config.logLevel)), "#")
-        ("threads", "Specify number of wallet sync threads", cxxopts::value<unsigned int>(threads)->default_value(std::to_string(std::max(1u, std::thread::hardware_concurrency()))), "#");
+        ("threads", "Specify number of wallet sync threads", cxxopts::value<unsigned int>(threads)->default_value(std::to_string(std::max(1u, std::thread::hardware_concurrency()))), "#")
+        ("v,version", "Output software version information", cxxopts::value<bool>(version)->default_value("false")->implicit_value("true"));
 
     options.add_options("Network")
         ("p,port", "The port to listen on for http requests",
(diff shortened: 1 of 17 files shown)


zpalmtree:
The build now passes. I believe I have also fixed the crashes.

zpalmtree:
Working on performance numbers now for a few different configurations and devices...

src/WalletApi/ParseArguments.cpp
```diff
@@ -24,11 +26,14 @@ Config parseArguments(int argc, char **argv)
...
 
+    unsigned int threads;
+
     options.add_options("Core")
         ("h,help", "Display this help message", cxxopts::value<bool>(help)->implicit_value("true"))
 
         ("v,version", "Output software version information", cxxopts::value<bool>(version)->default_value("false")->implicit_value("true"))
-        ("log-level", "Specify log level", cxxopts::value<int>(logLevel)->default_value(std::to_string(config.logLevel)), "#");
+        ("log-level", "Specify log level", cxxopts::value<int>(logLevel)->default_value(std::to_string(config.logLevel)), "#")
+        ("threads", "Specify number of wallet sync threads", cxxopts::value<unsigned int>(threads)->default_value(std::to_string(std::max(1u, std::thread::hardware_concurrency()))), "#");
```
brandonlehmann: While you're in here, can you re-arrange these to be alphabetical? I think version just needs moved down.

zpalmtree: Done.

PR #825, opened 2019-05-27
Add support for skipCoinbaseTransactions in WalletBackend

zpalmtree:
If the skipCoinbaseTransactions flag is given, the daemon will not return blocks which have no non coinbase transactions in, and will also not return coinbase transactions in the non empty blocks returned.

This doesn't effect sync speed for a local daemon, but should hopefully improve sync speed for remote daemons. Will work on testing this out.

Skipping coinbase transactions is *enabled* by default, unless the `--scan-coinbase-transactions` flag is given. Discussion on whether this should be enabled by default is invited. If we proceed, some documentation may need updating.

I also added a `config/Config.h` header which contains a global config, which is read/writable from everywhere. This makes setting configuration options easier, without having to pass the results through multiple constructors. Currently only the `skipCoinbaseTransactions` is stored here, but it could be valuable moving most of the configuration here. It can combine values that can be altered at runtime, for example, sync threads and coinbase scanning, and values which should remain constant, such as address prefix, in one place.


SoreGums:
> This doesn't effect sync speed for a local daemon, but should hopefully improve sync speed for remote daemons. Will work on testing this out.

if the wallet doesn't need to scan coinbase transactions wouldn't that mean there is no time spent doing that task (getting the info, transfering it over http, the wallet doing something with it)

so maybe this statement is misdirecting, as the changes are for the wallet and daemon comms, aka RPC interface and nothing to do with the daemon? or maybe it is written like this

`Hopefully this speeds up syncing a wallet from 0, as there are less transactions to process/transmit between the daemon and wallet`

zpalmtree:
Write it how you like, it doesn't make the sync any faster with a local daemon :man_shrugging: 

We're bottlenecked on DB access instead of block processing speed with a local daemon, and because we're still reading the same amount of blocks (then discarding empty ones), it doesn't increase that aspect of it. It's *possible* we could optimize the DB query to not include empty ones - not sure if that's something we can do with a kv Db.

This statement: `there are fewer transactions to process/transmit between the daemon and wallet` is correct, though.

SoreGums:
OK, clear - the optimisation is purely in the amount of data going to the wallet then.

so if the local daemon is the bottleneck it'll still be the bottleneck 👍

LeoCornelius:
Any Main reason this has not been merged? or are you just waiting to fit it into the next release/ not mess up your work flow

brandonlehmann:
> Any Main reason this has not been merged? or are you just waiting to fit it into the next release/ not mess up your work flow

Reviewing & testing code takes time and some of us have other obligations.

LeoCornelius:
> > Any Main reason this has not been merged? or are you just waiting to fit it into the next release/ not mess up your work flow
> 
> Reviewing & testing code takes time and some of us have other obligations.

fair 
enough. sorry my comment cam across as ``why have you not done this yet!`` but i was just checking if the code had a issue ;)

PR #828, opened 2019-06-02
Fix Portable Slow Hash

brandonlehmann:
While working on the [turtlecoin-crypto](https://github.com/turtlecoin/turtlecoin-crypto) package, I discovered that the portable slow-hash method was not returning valid hash results. As we seldom build on architectures that aren't x86 or arm (or that don't support AES), this has gone unnoticed for some time.

The easiest way to test for the need for this fix is the following:

### Current Development Branch

To force the use of the portable code, run the following:

```bash
cmake -DNO_AES=1
make -j
./src/cryptotest
```

You will see that the hash test **fails**

Rebuild with AES

```bash
cmake -DNO_AES=0
make -j
./src/cryptotest
```

The hash tests now **pass**

### This PR

Do the same steps above with this PR pulled in and you'll see that the portable slow-hash now passes the hash tests as expected.

**Note:** I've also moved `cn_slow_hash_state` into slow-hash-common.h as it's the same in all of the implementations.

### DIFF
--- a/src/crypto/slow-hash-arm.c
+++ b/src/crypto/slow-hash-arm.c
@@ -41,18 +41,6 @@ STATIC INLINE void xor64(uint64_t *a, const uint64_t b)
     *a ^= b;
 }
 
-  #pragma pack(push, 1)
-union cn_slow_hash_state
-{
-    union hash_state hs;
-    struct
-    {
-        uint8_t k[64];
-        uint8_t init[INIT_SIZE_BYTE];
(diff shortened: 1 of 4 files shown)


zpalmtree:
Opened https://github.com/brandonlehmann/turtlecoin/pull/6 to address some compiler warnings, other than that, it works as described.

brandonlehmann:
Merged those in.

PR #829, opened 2019-06-02
Increased defaults for DB r/w and # of threads

LeoStehlik:
Increased DB read & write buffers to 1024 MB, max open files to 500 and 10 DB threads. This is to allow nodes to cope with high volumes of transactions. Helps nodes to stay synced with the network, and / or ensures node to be able to resync as fast as possible. Tested with 66k+ transactions in the pool.

### DIFF
--- a/src/config/CryptoNoteConfig.h
+++ b/src/config/CryptoNoteConfig.h
@@ -260,10 +260,10 @@ const uint64_t P2P_DEFAULT_INVOKE_TIMEOUT                    = 60 * 2 * 1000; //
 const size_t   P2P_DEFAULT_HANDSHAKE_INVOKE_TIMEOUT          = 5000;          // 5 seconds
 const char     P2P_STAT_TRUSTED_PUB_KEY[]                    = "";
 
-const uint64_t DATABASE_WRITE_BUFFER_MB_DEFAULT_SIZE         = 256;
-const uint64_t DATABASE_READ_BUFFER_MB_DEFAULT_SIZE          = 10;
-const uint32_t DATABASE_DEFAULT_MAX_OPEN_FILES               = 100;
-const uint16_t DATABASE_DEFAULT_BACKGROUND_THREADS_COUNT     = 2;
+const uint64_t DATABASE_WRITE_BUFFER_MB_DEFAULT_SIZE         = 1024;          // 1 GB 
+const uint64_t DATABASE_READ_BUFFER_MB_DEFAULT_SIZE          = 1024;          // 1 GB
+const uint32_t DATABASE_DEFAULT_MAX_OPEN_FILES               = 500;           // 500 files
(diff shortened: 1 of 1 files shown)


bobbieltd:
It will consume more RAM. Improvement is disputable.

LeoStehlik:
Yes, it does consume more RAM, that's what RAM is for. Improvement confirmed repeatedly on the DEGO network by throwing at the daemon 66k+ transactions. All nodes that applied at least those or much higher db cache settings were able to stay on top of the chain.

bobbieltd:
Agree, RAM is for running software but not just only TurtleCoind. How many people do that have high end hardware (well they need to use RAM for other softwares too) ?
My opinion can be wrong due to my incomplete knowledge of tech.
Taking account only your write buffer 1Gb (not mention 1Gb read). Turtlecoin default max write number = 6 (        cfOpts.max_write_buffer_number = 6;https://github.com/turtlecoin/turtlecoin/blob/56173096d0bed7f62faa2e62ebc3a79ac44c7e88/src/CryptoNoteCore/MainChainStorageRocksdb.cpp#L49)
Rocksdb documentation (Options::max_write_buffer_number, which is maximum number of write buffers that are built up in memory.). Turtlecoin can build up to 6Gb if there is I/O bottleneck). It is too much.
I already tried to tweak these params in the past without any seen improvement.
Other question : Is rocksdb database true bottleneck when node has high number if txes ? CPU rise or Disk I/O rise ?

brandonlehmann:
These are not hard set requirements. These are the defaults. You can change them at runtime.

bobbieltd:
Another observation : many newly launched coins have very small blockchain data size comparing with current Turtlecoin blockchain.
If ratio of Rocksdb buffer sizes / blockchain size is important, small new coins should handle thousands of txes.

brandonlehmann:
> Another observation : many newly launched coins have very small blockchain data size comparing with current Turtlecoin blockchain.
> If ratio of Rocksdb buffer sizes / blockchain size is important, small new coins should handle thousands of txes.

Don’t mean to be that guy... but this is the TurtleCoin repo. I’m not sure what point you’re trying to make here.

bobbieltd:
It is quite clear if you follow your Turtlecoin forks’ networks.

bobbieltd:
Well, I made my points (perhaps, wrong or bad reasoning). I have no more things to say. Thanks TRTL team and many contributors.

brandonlehmann:
> It is quite clear if you follow your Turtlecoin forks’ networks.

In case you missed it... this is the TurtleCoin repo. We aren’t responsible for forks. We are; however, responsible for TurtleCoin. Personally, I don’t follow the forks and have no plans to do so. 

If you’re having a problem with a forks network, perhaps you should open an issue on their repo or talk to their maintainers.

PR #836, opened 2019-06-20
Fix tx extra public key parsing rare bug

zpalmtree:
If the payment ID ends with `01`, the public key will be incorrectly parsed, causing transactions to not be scanned correctly. Assuming payment ID's are uniformly distributed, this effects one in every 256 transactions sent with a payment ID.

For example, https://explorer.turtlecoin.lol/transaction.html?hash=13f86f5871526403dd18cc435e0788f254e7e2fd7137c65becf1c09a98a22992 will have the public key parsed as  `01ca1d931715391de63748b4640a2a75faedd53e9da24aee989ca56236ed8921` instead of `ca1d931715391de63748b4640a2a75faedd53e9da24aee989ca56236ed892193`

The bug is in /getwalletsyncdata - so you will need to update your daemon and resync if you are effected by this bug.

The fix takes into account payment ID's and public keys when parsing the extra. I'm not sure if we need to handle any other tags to make it foolproof. We don't need to handle extra padding - this could only cause an issue if the padding contained a `0x02`, indicating a payment ID - but padding must be all zeros.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -137,7 +137,7 @@ endif()
 
 # Add the dependencies we need
 target_link_libraries(Common __filesystem)
-target_link_libraries(CryptoNoteCore Common Logging Crypto P2P Rpc Http Serialization System sqlite3 ${Boost_LIBRARIES})
+target_link_libraries(CryptoNoteCore Utilities Common Logging Crypto P2P Rpc Http Serialization System sqlite3 ${Boost_LIBRARIES})
 target_link_libraries(cryptotest Crypto Common)
 target_link_libraries(Errors Crypto SubWallets Utilities)
 target_link_libraries(Logging Common)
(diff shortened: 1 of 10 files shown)


bobbieltd:
My 1 cent comment (merged mining tag should not happen but parsing it just in case is good idea) : 
- Format : Merged mining tag is designed like extra nonce tag as one byte tag (0x03) + one byte size (n) + n bytes
- Number : one byte tag (0x03) + one byte size (0x21) + one byte depth + 32 byte merkleRoot
In this loki block (https://lokiblocks.com/tx/8cdb37e41730048510377acbf596acafb5e320f29cc2eba0594537f258a156f2), merged mining tag is 0321000c4e3ad1b71c13e576c4a33a8957bda69059cad6687e79b0228120f2b1676bd9

zpalmtree:
Thanks for format clarification, I'll update to handle it.

PR #837, opened 2019-06-23
Upgrade wallet format

zpalmtree:
* Adds a flag, `--upgrade-wallet-format` to `turtle-service` which upgrades a wallet file `x.wallet` to `new-x.wallet`
* Adds transparent upgrading to `WalletBackend` which places the old file `x.wallet` in `old-version-backup-x.wallet` with the upgraded wallet being placed on top of `x.wallet` and opened like a normal wallet.
Passwords are retained across wallet files.

Testing:

#### 1)
* Create a wallet with `zedwallet`
* Note down the address and sync progress
* Open the wallet with `zedwallet-beta`. It should open, the address should be the same, and sync progress should jump to the creation timestamp block height`

#### 2)
* Import a wallet with `zedwallet`, with a scan height
* Open the wallet with `zedwallet-beta`. It should open, and sync progress should continue from the same scan height (or amount after you left the wallet running)

#### 3)
* Find a `zedwallet` wallet with transactions or send some transactions to a `zedwallet` wallet and then open it in `zedwallet-beta`
* Balance should be the same
* Transactions should be listed identically
* Sending a transaction for some or all of the balance should succeed

#### 4)
* Upgrade a wallet with `zedwallet-beta`
* Try to open the `old-version-backup.wallet` with `zedwallet`
* It should succeed.


brandonlehmann:
Is the first feature (on the legacy app) really required if WalletBackend (which feeds wallet-api and zedwallet-bets) upgrades the wallet anyways?

zpalmtree:
Not really..

zpalmtree:
Actually, it could be useful if someone wants to upgrade a wallet to use in other backends, for example, the javascript backend.

zpalmtree:
Finally passing tests

brandonlehmann:
> Actually, it could be useful if someone wants to upgrade a wallet to use in other backends, for example, the javascript backend.

Why wouldn't they open the wallet with `zedwallet-beta` or `wallet-api` and allow the conversion to occur and then close it? If I'm being honest, I'm a little hesitant to add feature(s) to code that is being deprecated.

zpalmtree:
> > Actually, it could be useful if someone wants to upgrade a wallet to use in other backends, for example, the javascript backend.
> 
> Why wouldn't they open the wallet with `zedwallet-beta` or `wallet-api` and allow the conversion to occur and then close it? If I'm being honest, I'm a little hesitant to add feature(s) to code that is being deprecated.

Sure, it would probably be a better idea to open via `wallet-api`, however if you're wanting to do it programmatically, might be easier to just run the program and exit, rather than starting an api, opening wallet, then closing api. :man_shrugging: 

I'm not that tied to it so if you don't see any value I'll remove it.

brandonlehmann:
> > > Actually, it could be useful if someone wants to upgrade a wallet to use in other backends, for example, the javascript backend.
> > 
> > 
> > Why wouldn't they open the wallet with `zedwallet-beta` or `wallet-api` and allow the conversion to occur and then close it? If I'm being honest, I'm a little hesitant to add feature(s) to code that is being deprecated.
> 
> Sure, it would probably be a better idea to open via `wallet-api`, however if you're wanting to do it programmatically, might be easier to just run the program and exit, rather than starting an api, opening wallet, then closing api. 🤷‍♂
> 
> I'm not that tied to it so if you don't see any value I'll remove it.

If you upgrade the wallet with `turtle-service` via `--upgrade-wallet-format` -- can `turtle-service` still use it? If not, then it doesn't make any sense to have it as an option IMO given that you're going to have to open it with `wallet-api` anyways.

zpalmtree:
> > > > Actually, it could be useful if someone wants to upgrade a wallet to use in other backends, for example, the javascript backend.
> > > 
> > > 
> > > Why wouldn't they open the wallet with `zedwallet-beta` or `wallet-api` and allow the conversion to occur and then close it? If I'm being honest, I'm a little hesitant to add feature(s) to code that is being deprecated.
> > 
> > 
> > Sure, it would probably be a better idea to open via `wallet-api`, however if you're wanting to do it programmatically, might be easier to just run the program and exit, rather than starting an api, opening wallet, then closing api. man_shrugging
> > I'm not that tied to it so if you don't see any value I'll remove it.
> 
> If you upgrade the wallet with `turtle-service` via `--upgrade-wallet-format` -- can `turtle-service` still use it? If not, then it doesn't make any sense to have it as an option IMO given that you're going to have to open it with `wallet-api` anyways.
`--upgrade-wallet-format` doesn't touch the original file, and makes a new WalletBackend compatible file.
The transparent upgrading replaces the original file, and moves the old file to a new name.

So yes, if you run `--upgrade-wallet-format x.wallet` you can still open `x.wallet` in `turtle-service`

brandonlehmann:
> > > > > Actually, it could be useful if someone wants to upgrade a wallet to use in other backends, for example, the javascript backend.
> > > > 
> > > > 
> > > > Why wouldn't they open the wallet with `zedwallet-beta` or `wallet-api` and allow the conversion to occur and then close it? If I'm being honest, I'm a little hesitant to add feature(s) to code that is being deprecated.
> > > 
> > > 
> > > Sure, it would probably be a better idea to open via `wallet-api`, however if you're wanting to do it programmatically, might be easier to just run the program and exit, rather than starting an api, opening wallet, then closing api. man_shrugging
> > > I'm not that tied to it so if you don't see any value I'll remove it.
> > 
> > 
> > If you upgrade the wallet with `turtle-service` via `--upgrade-wallet-format` -- can `turtle-service` still use it? If not, then it doesn't make any sense to have it as an option IMO given that you're going to have to open it with `wallet-api` anyways.
> > `--upgrade-wallet-format` doesn't touch the original file, and makes a new WalletBackend compatible file.
> > The transparent upgrading replaces the original file, and moves the old file to a new name.
> 
> So yes, if you run `--upgrade-wallet-format x.wallet` you can still open `x.wallet` in `turtle-service`

That's not 100% what I meant :) If you upgrade the wallet via this means, and it creates `new-x.wallet` why would you continue to use `x.wallet`? If `turtle-service` can upgrade the file but can't use the file `new-x.wallet` it seems pointless IMO to even let `turtle-service` know about the new format.

If you're not married to it, I'd leave the legacy binaries alone as:

* It adds enhancements to legacy (deprecated tools) that are non-bugfix
* It's not really an upgrade path for the wallet format in this way as the tool can't app can't use the file it creates
* It works differently than the upgrade via `wallet-api` or `zedwallet-beta` (they replace `x.wallet` and save the original off to the side versus saving the new off to the side and leaving `x.wallet` alone)

zpalmtree:
Removed

src/WalletBackend/WalletBackend.cpp
```diff
@@ -377,6 +382,55 @@ std::tuple<Error, std::shared_ptr<WalletBackend>> WalletBackend::createWallet(
...
+        /* Currency contains our coin parameters, such as decimal places, supply */
+        const CryptoNote::Currency currency 
+            = CryptoNote::CurrencyBuilder(logManager).currency();
+
+        System::Dispatcher localDispatcher;
+        System::Dispatcher *dispatcher = &localDispatcher;
+
+        /* Our connection to turtlecoind */
+        std::unique_ptr<CryptoNote::INode> node(
+            new CryptoNote::NodeRpcProxy(daemonHost, daemonPort, 10, logManager)
```
brandonlehmann: Will this error out if you open `wallet-api` or `zedwallet-beta` against a SSL or cache implementation and prevent the upgrade from processing?

zpalmtree: Seems to work fine. I don't think the node is getting inited, as there was no delay when opening with a cache.

brandonlehmann: Gotcha. I didn't think so but wanted to make sure I wasn't cray cray.

brandonlehmann:
Travis completed. Looks like the webhook is hung.

PR #845, opened 2019-07-02
Decrease miner dependencies

zpalmtree:
Should make miner compiling a lot faster.


zpalmtree:
> If you’re going to pull the check hash and block hash code into utilities, it needs to be deduplicated from CachedBlock and other places as well.

It's used a *ton* of places, I think replacing it all would take a significant amount of time for little gain. Furthermore, it is providing a different set of needs than CachedBlock - CachedBlock, hence the name, caches the results of the calculations which I have pulled out here. In some places, especially `getBlockLongHash` we are probably requiring those calculations to be cached for performance. Here, however, the operations are only needed once.

I would rather not have the code that determines which hashing algorithm to use in two places... I'll think on the best way to handle that, I think I have a good idea for it.

brandonlehmann:
> I would rather not have the code that determines which hashing algorithm to use in two places... I'll think on the best way to handle that, I think I have a good idea for it.

That's my point. This duplicates code and requires that it be maintained in two places. If you've go an idea of how to avoid that, that'll solve my issue :)

zpalmtree:
60ceeed

zpalmtree:
Wait a sec, must have done a typo, it's not syncing.

zpalmtree:
fixed

PR #846, opened 2019-07-08
Remove Legacy Unused Code

brandonlehmann:
* Remove testnet references that have not been supported for some time.
* Remove premine references that have **never** been used during the **entire** life of the project.

### DIFF
--- a/src/Daemon/Daemon.cpp
+++ b/src/Daemon/Daemon.cpp
@@ -50,41 +50,14 @@ using namespace CryptoNote;
 using namespace Logging;
 using namespace DaemonConfig;
 
-void print_genesis_tx_hex(const std::vector<std::string> rewardAddresses, const bool blockExplorerMode, std::shared_ptr<LoggerManager> logManager)
+void print_genesis_tx_hex(const bool blockExplorerMode, std::shared_ptr<LoggerManager> logManager)
 {
-  std::vector<CryptoNote::AccountPublicAddress> rewardTargets;
-
   CryptoNote::CurrencyBuilder currencyBuilder(logManager);
   currencyBuilder.isBlockexplorer(blockExplorerMode);
 
   CryptoNote::Currency currency = currencyBuilder.currency();
 
-  for (const auto& rewardAddress : rewardAddresses)
-  {
-    CryptoNote::AccountPublicAddress address;
-    if (!currency.parseAccountAddressString(rewardAddress, address))
-    {
-      std::cout << "Failed to parse genesis reward address: " << rewardAddress << std::endl;
-      return;
-    }
-    rewardTargets.emplace_back(std::move(address));
-  }
-
-  CryptoNote::Transaction transaction;
-
-  if (rewardTargets.empty())
-  {
-    if (CryptoNote::parameters::GENESIS_BLOCK_REWARD > 0)
-    {
-      std::cout << "Error: Genesis Block Reward Addresses are not defined" << std::endl;
-      return;
-    }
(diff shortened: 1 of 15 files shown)


src/Daemon/Daemon.cpp
```diff
@@ -50,41 +50,16 @@ using namespace CryptoNote;
...
-    CryptoNote::AccountPublicAddress address;
-    if (!currency.parseAccountAddressString(rewardAddress, address))
-    {
-      std::cout << "Failed to parse genesis reward address: " << rewardAddress << std::endl;
-      return;
-    }
-    rewardTargets.emplace_back(std::move(address));
-  }
-
   CryptoNote::Transaction transaction;
```
zpalmtree: Can just do
```cpp
const CryptoNote::Transaction transaction = CryptoNote::CurrencyBuilder(logManager).generateGenesisTransaction();
```
or
```cpp
const auto transaction = CryptoNote::CurrencyBuilder(logManager).generateGenesisTransaction();
```
if that's your thing, now we don't need to conditionally assign it

src/P2p/NetNode.cpp
```diff
@@ -380,12 +380,8 @@ std::string print_peerlist_to_string(const std::list<PeerlistEntry>& pl) {
   //-----------------------------------------------------------------------------------
 
   bool NodeServer::init(const NetNodeConfig& config) {
-    if (!config.getTestnet()) {
-      for (auto seed : CryptoNote::SEED_NODES) {
-        append_net_address(m_seed_nodes, seed);
-      }
-    } else {
-      m_network_id.data[0] += 1;
+    for (auto seed : CryptoNote::SEED_NODES) {
```
zpalmtree:
```suggestion
    for (const auto &seed : CryptoNote::SEED_NODES) {
```

src/P2p/P2pNodeConfig.cpp
```diff
@@ -1,19 +1,7 @@
 // Copyright (c) 2012-2017, The CryptoNote developers, The Bytecoin developers
+// Copyright (c) 2018, The TurtleCoin Developers
```
zpalmtree:
```suggestion
// Copyright (c) 2018-2019, The TurtleCoin Developers
```

src/Rpc/CoreRpcServerCommandsDefinitions.h
```diff
@@ -358,7 +357,6 @@ struct COMMAND_RPC_GET_INFO {
       KV_MEMBER(minor_version)
       KV_MEMBER(start_time)
       KV_MEMBER(synced)
-      KV_MEMBER(testnet)
```
zpalmtree: Technically we're changing the API here, no? Really doubt anyone is using it, but worth noting. What are your thoughts?

brandonlehmann: Given that the `--testnet` flag hasn't existed and thus this was never set to true, any front-end utilities will see this as `false` when checking for it anyways as I don't recall ever coming across anything the explicitly checked for a an actual type `false`. It's minor, and we can not it in the release notes but I don't think it's going to hurt anything.

src/Daemon/Daemon.cpp
brandonlehmann: Updated. I missed that while ripping and cleaning a bit.

src/Rpc/CoreRpcServerCommandsDefinitions.h
zpalmtree: Ok, cool

PR #848, opened 2019-07-16
Fix issue with miner block serialization

zpalmtree:
See https://github.com/turtlecoin/turtlecoin/commit/3bf8f7259dda0953d53f8780ec68251cea4449e4#r34313195

@DaveLong Can you see if this fixes your issue? :smiley:

### DIFF
--- a/src/Miner/BlockUtilities.cpp
+++ b/src/Miner/BlockUtilities.cpp
@@ -69,7 +69,7 @@ Crypto::Hash getBlockHash(const CryptoNote::BlockTemplate &block)
 
     if (block.majorVersion >= CryptoNote::BLOCK_MAJOR_VERSION_2)
     {
-        const auto& parentBlock = getParentBlockBinaryArray(block, false);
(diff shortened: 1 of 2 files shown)


davehlong:
Almost, just need to change this line:

https://github.com/turtlecoin/turtlecoin/blob/0bf72928fbe7510a2703ee5f548bc56333c57809/src/Miner/BlockUtilities.cpp#L72

to `const auto& parentBlock = getParentBlockHashingBinaryArray(block, false);`

else the wrong hashes are reported in the miner for found blocks. Although the blocks are successfully submitted.

zpalmtree:
> Almost, just need to change this line:
> 
> https://github.com/turtlecoin/turtlecoin/blob/0bf72928fbe7510a2703ee5f548bc56333c57809/src/Miner/BlockUtilities.cpp#L72
> 
> to `const auto& parentBlock = getParentBlockHashingBinaryArray(block, false);`
> 
> else the wrong hashes are reported in the miner for found blocks. Although the blocks are successfully submitted.

Ah, good catch. Thanks again!

PR #851, opened 2019-07-28
Hard Fork to Chukwa (Argon2id) at Block 1,800,000

brandonlehmann:
- Argon2 Implementation from: https://github.com/WOnder93/argon2
  - Updated CPU feature detection
- Chukwa - Argon2id Properties: 512KB, 1 parallelism, 3 iterations.
- Hard Fork @ block 1,800,000
- Difficulty reset at block 1,800,000
  - Does not alter the difficulty calculation itself; however, it causes a massive drop in difficulty on the first block and slowly re-engages to 100% of the difficulty value over `DIFFICULTY_BLOCKS_COUNT_V3` (currently `61` blocks)
- Bumps the P2P Version to `6` and sets the minimum supported peer version to `5` (current as of v0.14.6)


src/cryptonotecore/Currency.cpp
```diff
@@ -750,6 +760,7 @@ namespace CryptoNote
             case BLOCK_MAJOR_VERSION_3:
             case BLOCK_MAJOR_VERSION_4:
             case BLOCK_MAJOR_VERSION_5:
+            case BLOCK_MAJOR_VERSION_6:
```
zpalmtree: Maybe we should change this to
```cpp
switch (block.getBlock().majorVersion)
{
    case BLOCK_MAJOR_VERSION_1:
    {
        return checkProofOfWorkV1(block, currentDiffic);
    }
    default:
    {
        return checkProofOfWorkV2(block, currentDiffic);
    }
}
```
?

src/cryptonotecore/Difficulty.cpp
```diff
@@ -7,6 +7,30 @@
 #include <algorithm>
 #include <config/CryptoNoteConfig.h>
 
+/* Helper method that helps us to calculate a difficulty reset with a ramp
+   towards returning full control to the normal DA */
+inline float calculateDifficultyResetMultiplier(
```
zpalmtree: Any particular reason to make this inline?

src/cryptonotecore/Difficulty.cpp
```diff
@@ -7,6 +7,30 @@
...
+    uint64_t currentDifficultyResetDepth = blockIndex - resetHeight;
+    float currentDifficultyResetDepthPercent = currentDifficultyResetDepth / resetWindow;
+
+    /* Then we figure out the inverse of the multiplier specified (the difficulty reduction)
+      that is being applied */
+    float currentDifficultyReduction = 1 - resetMultiplier;
+
+    /* Now we can figure out what our adjusted multiplier is by multiplying our reset depth percentage
+      by the overall difficulty reduction and add the result to the base multiplier */
+    float currrentDifficultyMultiplier =
```
zpalmtree:
```suggestion
    float currentDifficultyMultiplier =
```

src/cryptonotecore/Difficulty.cpp
```diff
@@ -7,6 +7,30 @@
...
+        (currentDifficultyResetDepthPercent * currentDifficultyReduction) + resetMultiplier;
+
+    return currrentDifficultyMultiplier;
```
zpalmtree:
```suggestion
    return currentDifficultyMultiplier;
```

src/cryptonotecore/Difficulty.cpp
```diff
@@ -136,3 +160,27 @@ uint64_t nextDifficultyV3(std::vector<uint64_t> timestamps, std::vector<uint64_t
...
+    {
+        /* Calculate our adjusted difficulty multiplier */
+        float currrentDifficultyMultiplier = calculateDifficultyResetMultiplier(
+            blockIndex,
+            CryptoNote::parameters::DIFFICULTY_RESET_HEIGHT_V1,
+            CryptoNote::parameters::DIFFICULTY_RESET_WINDOW_V1,
+            CryptoNote::parameters::DIFFICULTY_RESET_MULTIPLIER_V1);
+
+        /* We use that updated difficulty multiplier to calculate our adjusted difficulty */
+        return (uint64_t)nextDifficulty * currrentDifficultyMultiplier;
```
zpalmtree:
```suggestion
        return static_cast<uint64_t>(nextDifficulty * currentDifficultyMultiplier);
```

src/cryptonotecore/Difficulty.cpp
```diff
@@ -136,3 +160,27 @@ uint64_t nextDifficultyV3(std::vector<uint64_t> timestamps, std::vector<uint64_t
...
+{
+    /* This code supports the difficulty reset as activated by the specified height as required.
+       We allow the normal difficulty calculation to run unimpeded and then multiply the calculated
+       difficulty by the multiplier to arrive at our adjusted difficulty for the reset window specified */
+    if (blockIndex >= CryptoNote::parameters::DIFFICULTY_RESET_HEIGHT_V1
+        && blockIndex < CryptoNote::parameters::DIFFICULTY_RESET_HEIGHT_V1
+                            + CryptoNote::parameters::DIFFICULTY_RESET_WINDOW_V1)
+    {
+        /* Calculate our adjusted difficulty multiplier */
+        float currrentDifficultyMultiplier = calculateDifficultyResetMultiplier(
```
zpalmtree:
```suggestion
        float currentDifficultyMultiplier = calculateDifficultyResetMultiplier(
```

src/cryptonotecore/Difficulty.cpp
brandonlehmann: Is there any particular reason not to?

src/cryptonotecore/Difficulty.cpp
brandonlehmann: Pffft, you caught me

src/cryptonotecore/Difficulty.cpp
zpalmtree: Because the compiler is smarter than you

brandonlehmann: Then it shouldn't matter if it's inline or not.

src/cryptonotecore/Currency.cpp
brandonlehmann: I'll allow it.

src/cryptonotecore/Difficulty.cpp
brandonlehmann: whoops

src/cryptonotecore/Difficulty.cpp
brandonlehmann: Fixed

src/cryptonotecore/Difficulty.cpp
brandonlehmann: fixed

src/cryptonotecore/Difficulty.cpp
zpalmtree: Marking something inline doesn't effect optimization because the compiler ignores it, but can slow down compilation, I *believe*. It should only be done when you're declaring functions in headers to fix linking errors.

brandonlehmann: > Marking something inline doesn't effect optimization because the compiler ignores it, but can slow down compilation, I _believe_. It should only be done when you're declaring functions in headers to fix linking errors.

I'll change it to a private static, not because you said to (because the reason you gave that inline should be used sounds like popsicle sticks and duct tape), but because I want to.

brandonlehmann:
Resolved.

PR #855, opened 2019-07-29
wallet-api no-console option

brandonlehmann:
Adds `--no-console` option to wallet-api to allow for sending the process to the background without third-party utilities.


SoreGums:
cool looks good

PR #856, opened 2019-07-30
Fixing GCC+Clang compile warnings

zpalmtree:
Code now compiles with zero warnings on clang 8 + gcc 9, on Linux.


brandonlehmann:
CI test failed so I’m calling your bluff.

src/cryptonotecore/Currency.cpp
```diff
@@ -289,8 +288,8 @@ namespace CryptoNote
         uint64_t summaryAmounts = 0;
         for (size_t no = 0; no < outAmounts.size(); no++)
         {
-            Crypto::KeyDerivation derivation = boost::value_initialized<Crypto::KeyDerivation>();
```
brandonlehmann: There's one more of these in there that I think can come easily out, mind grabbing it? (L#57)

While you're in there... the trim on L#475, if that can be removed I believe will remove the boost requirement from this file.

zpalmtree: Done

brandonlehmann: Thanks sweaty.

PR #857, opened 2019-07-30
Backend tweaks

zpalmtree:
* Fix potential issue with forked input handling
* Convert consistency crashes to log warnings

### DIFF
--- a/src/subwallets/SubWallets.cpp
+++ b/src/subwallets/SubWallets.cpp
@@ -9,6 +9,7 @@
 #include <config/CryptoNoteConfig.h>
 #include <ctime>
 #include <mutex>
+#include <logger/Logger.h>
 #include <random>
 #include <utilities/Addresses.h>
 #include <utilities/Utilities.h>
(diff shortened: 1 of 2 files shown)


src/subwallets/SubWallets.cpp
```diff
@@ -343,9 +351,16 @@ void SubWallets::addTransaction(const WalletTypes::Transaction tx)
     {
         std::stringstream stream;
 
-        stream << "Transaction " << tx.hash << " was added to the wallet twice! Terminating.";
+        stream << "Transaction " << tx.hash << " was erronously "
```
brandonlehmann: `erroneously`. Also on L#311.

That whole line reads really weird to me for some reason.

`Transaction <hash> was erroneously attempted to be added to the wallet twice.`

`Transaction <hash> already exists in wallet. Ignoring.`

zpalmtree: Fixed

PR #858, opened 2019-08-01
Fix a couple of incorrect return codes

zpalmtree:
All other transaction types return 201, so updating to that.

I don't see any info why setNodeInfo() should return 202 as the request has been accepted and processed, so updating to 200.

Will fix conflicts in a bit, fork is out of date t_t


zpalmtree:
Conflicts fixed.

PR #860, opened 2019-08-07
Correcting Transaction Validation Error Message

brandonlehmann:


brandonlehmann:
Proceeding with merge as the commit tree won’t contain the Travis fix.

PR #863, opened 2019-08-08
Bubble the error message from isTransactionValidForPool up to the node RPC call

### LINKED ISSUE #861: Return better error messages via the node /sendrawtransaction method
The `/sendrawtransaction` RPC call to the daemon needs to return the full error message from `TransactionValidationErrors.h` when an error is encountered. The current responses of `OK` or `Failed` do not provide enough information for developers to easily discern what the actual issue was so that they can respond accordingly. This will require bubbling up the error from `Core::isTransactionValidForPool`

The full error message should be in **addition** to the currently provided `status` property to avoid breaking changes.

brandonlehmann:
Resolves #861

### DIFF
--- a/src/cryptonotecore/Core.cpp
+++ b/src/cryptonotecore/Core.cpp
@@ -456,7 +456,8 @@ namespace CryptoNote
             auto transactions = alt->getRawTransactions(alt->getTransactionHashes());
             for (auto &transaction : transactions)
             {
-                if (addTransactionToPool(std::move(transaction)))
+                const auto [success, error] = addTransactionToPool(std::move(transaction));
+                if (success)
                 {
                     // TODO: send notification
                 }
@@ -1372,7 +1373,8 @@ namespace CryptoNote
             auto tx = pool.getTransaction(hash);
             pool.removeTransaction(hash);
 
-            if (!addTransactionToPool(std::move(tx)))
+            const auto [success, error] = addTransactionToPool(std::move(tx));
+            if (!success)
             {
                 notifyObservers(makeDelTransactionMessage({hash}, Messages::DeleteTransaction::Reason::NotActual));
             }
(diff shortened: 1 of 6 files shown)


src/cryptonotecore/Core.cpp
```diff
@@ -1650,30 +1655,30 @@ namespace CryptoNote
...
+    std::tuple<bool, std::string> Core::isTransactionValidForPool(
         const CachedTransaction &cachedTransaction,
         TransactionValidatorState &validatorState)
     {
         auto [success, err] = Mixins::validate({cachedTransaction}, getTopBlockIndex());
 
         if (!success)
         {
-            return false;
+            return {false, "Transaction does not contain the proper number of sing signatures"};
```
zpalmtree:
```suggestion
            return {false, "Transaction does not contain the proper number of ring signatures"};
```

src/cryptonotecore/Core.cpp
```diff
@@ -1693,7 +1698,7 @@ namespace CryptoNote
                                      << " is not valid. Reason: transaction is too big ("
                                      << cachedTransaction.getTransactionBinaryArray().size()
                                      << "). Maximum allowed size is " << maxTransactionSize;
-            return false;
+            return {false, "Transaction is too large"};
```
zpalmtree: Might want to clarify that large is in terms of KB? Users have previously been confused by messages referring to size, thinking it was referring to the amount they were sending.. which it *kinda* was, but not directly.

src/cryptonotecore/Core.cpp
```diff
@@ -1706,10 +1711,10 @@ namespace CryptoNote
         {
             logger(Logging::WARNING) << "Transaction " << cachedTransaction.getTransactionHash()
                                      << " is not valid. Reason: fee is too small and it's not a fusion transaction";
-            return false;
+            return {false, "Transaction fee is too small for the transaction"};
```
zpalmtree:
```suggestion
            return {false, "Transaction fee is too small"};
```
Maybe :man_shrugging:

src/cryptonotecore/Core.cpp
brandonlehmann: Thank you, but I prefer it my way.

src/cryptonotecore/Core.cpp
brandonlehmann: `Transaction size (bytes) is too large` ?

zpalmtree: That works

brandonlehmann:
All cleaned up.

PR #865, opened 2019-08-19
this is annoying 

crappyrules:
Bound is the correct past tense of bind. Binded is an incorrect conjugation of the same verb.


zpalmtree:
Recommend squash on merge

PR #873, opened 2019-09-02
Proposal for zstd instead of lz4

wrkzdev:
* Replace lz4 with zstd
* Upgrade rocksdb to 6.4
* Some tests: https://github.com/wrkzcoin/wrkzcoin/issues/4

Please review.


zpalmtree:
Code changes look good to me. Do we want to remove the LZ4 folder if we're not using it anymore?

brandonlehmann:
> Code changes look good to me. Do we want to remove the LZ4 folder if we're not using it anymore?

Only if we make a note to make it clear in the release notes that LZ4 got the axe.

zpalmtree:
> > Code changes look good to me. Do we want to remove the LZ4 folder if we're not using it anymore?
> 
> Only if we make a note to make it clear in the release notes that LZ4 got the axe.

LZ4 is getting the axe either way if we merge this

brandonlehmann:
> > > Code changes look good to me. Do we want to remove the LZ4 folder if we're not using it anymore?
> > 
> > 
> > Only if we make a note to make it clear in the release notes that LZ4 got the axe.
> 
> LZ4 is getting the axe either way if we merge this

True, nuke the folder.

zpalmtree:
Next question, do we want to enable compression by default?

brandonlehmann:
> Next question, do we want to enable compression by default?

Not with the current read/write buffer/background threads defaults IMO. Compaction to lower levels as well as the compression overhead increases the memory requirements during those processes. We're already demanding 2GB of memory for *standard* operation and 10 threads -- which is more than many node operators have available.

I'm all for enabling it by default but the default read/write buffers need to come down to 256MB (still an increase from the originals) and drop the background threads down to `4`.

That being said, @wrkzdev what is the impact of enabling compression on an already synced node? How long does the daemon take to open/close the DB the first time? Node operators have to plan for this in their maintenance periods if enabled by default or enabling manually.

zpalmtree:
I think @LeoStehlik might have some insight on the buffer changes as well.

wrkzdev:
I haven't tried to enable compression to an existing synced data yet. I may suggest to sync from zero. IMO, shall be faster. RocksDB has so many parameters, we can also check the params where it hasn't been set in RocksDBWrapper (https://github.com/turtlecoin/turtlecoin/blob/development/src/cryptonotecore/RocksDBWrapper.cpp#L191), in `data-dir/DB/OPTIONS-XXXXX`

Closing/Opening, I haven't noticed any lag with TRTL daemon.

LeoStehlik:
My five cents, the buffer settings are highly individual system from system. Our goal should be to find defaults that bring operational stability and availability for most common users, however, educating our userbase (including service operators) that adjusting those values according to their system significantly above or below the defaults fine-tuned for their system/platform can help them. 

That said, there are two different settings people need to take into consideration. First, a different buffer settings helps while synchronizing blockchain. Second, there are different settings for day to day operations. 

As for myself, for normal operations, I never go down below 512 read/write buffers and 8 threads on my lowest-performing system architectures, barely below 2048/2048 and 16 threads on xxl VPS systems, and much higher on dedicated server hardware that run production services. This includes scenarios with zstd compression enabled. I found that performance with compression enabled and small read/write buffers while syncing suffers, however in normal ops lower values are ok. 

That said, for a user who wants to run a local daemon to occasionally synchronize the wallet at home and does not care a] how long blockchain sync will take b] if his/her node goes off the chain in case of high load of transactions in the tx pool, the 256/256/4 settings proposed by IBMCD are reasonable, however, service operators beware - you’ll need to increase these defaults.

brandonlehmann:
Copied from #dev_core

**iburnmycd**: The write buffer is what blooms and amplifies though. I’m okay with asymmetrical but no way can the default be 1gb write.  I still maintain that 256mb is the maximum default write for the average user. If service ops want to bump it they can via the cli args but most users don’t read past “start TurtleCoind”.

I know where your recommendations are coming from and don’t dispute that they help considerably for another chain but what’s required for the gander isn’t always the same for the goose

**LeoCuvee**: I’m not mad about it, all I want is to speak up - not all I say need to be agreed with, and it is fine. Side line - when I had my trtl testnet up and tried to losely replicate “heavy tx load” situation, the code behaves similar to that of other chain (on Linux). Pushing the defaults up was more of “having the network being able to cope with” rather than being reduced to a few working nodes. I look at those settings as risk mitigation - instead of being reduced to a 4-6 working nodes network at the time of very high # of tx, the increased defaults allow us stay up with about 50% of all nodes in the network. And agree, this may not pose such a high risk for trtl chain - but hey Inwould not be a good Turtle if I was quiet

**iburnmycd**: Fair enough and appreciated from my standpoint at least. @Olé Cuvée I have concerns with setting the bar so high for users to run them locally. I consider such requirements a barrier to entry that has to be approached with caution. That being said, it's important to remember that the node doesn't go down it merely appears to be down in that it's slow to respond to RPC requests and can delay some syncing. For a general user running a node locally, it's very likely a minor issue in the delay there. For a service op (public node & etc) I'm all for listing recommended settings and hardware requirements

In short, I'd feel much better about telling two dozen users (service ops) to turn up the settings than try to explain to 1k+ users that they need to turn down the settings

**LeoCuvee**: Makes sense, I consider this well debated and argued - I think we both understand where we come from, and it's mostly about "we don't need to force all users to use a big hammer, rather advice a subset of users that for certain scenarios a sledgehammer is the best tool. Bottom line ... as long as we don't set the default back to where it was, it still brings a significant improvement

brandonlehmann:
I'm testing this PR now. So far, I've confirmed that enabling compression on an already synced DB does **not** result in the compression being applied retroactively which was one of my concerns (that it would take time opening/closing to do the compression). I'm now running a resync from 0 with compression enabled.

brandonlehmann:
Initial results of the `DB` folder compression (block height 1,824,806):

Uncompressed: 59,888,484,048 bytes
Compressed: 40,355,804,718 bytes
Reduction: 19,532,679,330 bytes (~32.6% reduction)

Not too shabby. I'm doing another resync with the local cache (blocks.bin) replaced by rocksdb now.

brandonlehmann:
Results of the `blocks.bin` replacement via RocksDB:

Uncompressed: 22,022,820,875 bytes
Compressed: 23,945,815,369 bytes
Increase: 1,922,994,494 bytes (~8.7% increase)

Based on the testing, my recommendations:
* Remove `--rocksdb` option for the local cache
* Remove `--sqlite` option for the local cache
* Set `--db-enable-compression` on by default
* Reduce the **default** parameters for the DB options
  * `128MB` read cache
  * `256MB` write cache
  * `125` max open files
  * `4` db thread count
* Provide a notice, via the release notes, that it is recommended that **Service Operators** operate the daemon with higher than default values. Example:
  * `1024MB` read cache
  * `1024MB` write cache
  * `500` max open files
  * `10` db thread count

SoreGums:
hmm with this patch merged - the options files in the DB folder show 

`compression_per_level=kNoCompression:kNoCompression:kNoCompression:kNoCompression:kNoCompression:kNoCompression:kNoCompression:kNoCompression:kNoCompression:kNoCompression`

wonder why the config option didn't get picked up from the config file

`"db-enable-compression": true`

brandonlehmann:
> hmm with this patch merged - the options files in the DB folder show
> 
> `compression_per_level=kNoCompression:kNoCompression:kNoCompression:kNoCompression:kNoCompression:kNoCompression:kNoCompression:kNoCompression:kNoCompression:kNoCompression`
> 
> wonder why the config option didn't get picked up from the config file
> 
> `"db-enable-compression": true`

Did you sync from scratch? Also, do a load and dump config and verify it’s in the dump.

SoreGums:
well well well... it's `false`

image

Oh well will go again with the recommended settings above on my Win10 2GB 2x Core and see what happens.

It was from scratch, at least now this go around it won't be waiting on the network to send packets.

brandonlehmann:
Phew. At least my 7gb number held true for someone other than myself using the recommended. I’m not crazy :)

SoreGums:
well I went up to 2048, 20x threads and 1,000 files open :P

SoreGums:
probably need to setup a bunch of wallet-api's doing stuff to push that RAM number

SoreGums:
right on this new VM, config option is not picked up, only the `--db-enable-compression true` cli arg is used - wonder if the other options are being used....

PR #876, opened 2019-09-05
Add wallet upgrader binary

zpalmtree:
This adds a binary that upgrades the wallet, in a standalone binary rather than part of zedwallet-beta. It's a bit more streamlined for the upgrade process.

I haven't yet added it to the travis / appveyor. If you like the concept, I'll do that.


SoreGums:
Yes please. The Russian person that was in help was difficult to assist them... If this were a thing probably would have been simpler to get them sorted.

This upgrades from turtle-service style wallets too?

ExtraHash:
yes, this updates from the old to new format

SoreGums:
For some reason when opening their wallet file in zedwallet-beta it couldn't be opened. I'm guessing something funky with their password and turtle-service... If it worked fine in the Gui that was using turtle-service then something over there was being stupid....

ExtraHash:
turtle-service format shouldn't be any different than the zedwallet format, they're the same, so that would be odd

zpalmtree:
Could have been very old version 2 format instead of current zedwallet/turtle-service version 6 format, which we dropped support of around 0.8.4 or something.

zpalmtree:
Let me add to travis/appveyor before merging

brandonlehmann:
Aye

zpalmtree:
Done. Just needed to update travis, appveyor just globs all *.exe.

PR #877, opened 2019-09-06
Improved Wallet Upgrade Routine

brandonlehmann:
* Replaced backup mechanism with a simple backup operation (copy the file) before touching anything
  * If we can't back up, we exit immediately because moving forward, although unlikely, could result in data loss
* Wallet must be shutdown before attempting to write the new wallet format to the wallet file


ExtraHash:
does this now accept absolute paths as inputs as well?

brandonlehmann:
> does this now accept absolute paths as inputs as well?

No, this doesn't fix that.

See #878.

ExtraHash:
Doesn't #876 need to be merged first? it's still open

brandonlehmann:
> Doesn't #876 need to be merged first? it's still open

This is non-dependent on other PRs. Hence the reason they are separated.

PR #878, opened 2019-09-06
Less Boost, More Standard Library

brandonlehmann:
* Remove `boost::filesystem` in favor of `std::filesystem` in `FileMappedVector.h`

### DIFF
--- a/src/common/FileMappedVector.h
+++ b/src/common/FileMappedVector.h
@@ -8,8 +8,8 @@
 #include "common/ScopeExit.h"
 #include "system/MemoryMappedFile.h"
 
-#include <boost/filesystem.hpp>
 #include <cassert>
+#include <common/FileSystemShim.h>
 #include <cstdint>
 #include <string>
 
@@ -446,21 +446,24 @@ namespace Common
 
         const uint64_t initialCapacity = 10;
 
-        boost::filesystem::path filePath = path;
-        boost::filesystem::path bakPath = path + ".bak";
+        fs::path filePath = path;
+
+        fs::path bakPath = path + ".bak";
+
         bool fileExists;
-        if (boost::filesystem::exists(filePath))
+
+        if (fs::exists(filePath))
         {
-            if (boost::filesystem::exists(bakPath))
+            if (fs::exists(bakPath))
             {
-                boost::filesystem::remove(bakPath);
+                fs::remove(bakPath);
             }
 
             fileExists = true;
         }
-        else if (boost::filesystem::exists(bakPath))
+        else if (fs::exists(bakPath))
         {
-            boost::filesystem::rename(bakPath, filePath);
+            fs::rename(bakPath, filePath);
             fileExists = true;
         }
         else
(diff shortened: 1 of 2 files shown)


src/common/FileMappedVector.h
```diff
@@ -940,17 +943,26 @@ namespace Common
...
+        /* Define the path to the backup file */
+        fs::path bakPath = m_path + ".bak";
+
+        /* Create a new temporary filename and path. Due to variations in
+           where different platforms create the temporary path to, we are
+           only going to use the filename of the unique path and prepend
+           the backup file's parent_path (aka. directory path) to come up
+           with the full temporary file path */
+        std::string uniqueName = std::tmpnam(nullptr);
+        fs::path uniquePath = uniqueName;
```
zpalmtree: Why are we converting to a `fs::path` here, then calling `.filename().string()` on it below - Couldn't we just do
```cpp
fs::path tmpPath = bakPath.parent_path().string() + uniqueName;
```

It doesn't look like we're using it elsewhere.

src/common/FileMappedVector.h
```diff
@@ -940,17 +943,26 @@ namespace Common
...
+        fs::path tmpPath = bakPath.parent_path().string() + uniquePath.filename().string();
```
zpalmtree: Might we want to add a `-`, `_`, `.`, etc here? If our path is `foo` then the result file may be something like `foo.bakfileDjwifs` intead of `foo.bak_fileDjwifs`

src/common/FileMappedVector.h
brandonlehmann: `tmpnam()` returns a `char*`, not a `path` and we just need the filename. Using just `uniqueName` results in a mess and possible write/permission errors not to mention it creates it in the environment temp directory.

src/common/FileMappedVector.h
brandonlehmann: What? `foo.bak` would not be part of `tmpPath` due to the use of `parent_path()` which does not include the filename (`foo.bak`) thus the path would be `/some/path/fileDjwifs`

src/common/FileMappedVector.h
zpalmtree: I'm not sure I follow. I understand it returns a `char *`, and then assigning to a `std::string` copies it. However, it doesn't appear to return a filepath to me, just a temporary name, going by the examples given. For this reason, I believe `uniquePath.filename().string()` is going to result in the same thing as `uniqueName`

src/common/FileMappedVector.h
zpalmtree: Glanced over the `parent_path()`, ok.

src/common/FileMappedVector.h
zpalmtree: Eh, nevermind, I read the docs wrong again.

brandonlehmann: > For this reason, I believe `uniquePath.filename().string()` is going to result in the same thing as `uniqueName`

It doesn't; that's the misconception. It returns an absolute path to the temporary file on *some platforms*

PR #879, opened 2019-09-07
Default build type set back to 'Release'

brandonlehmann:
Repeats #812 that appears to have been lost in another PR. `CMAKE_BUILD_TYPE` is one of those special vars that requires a bit of special handling based upon information available on the CMake mailing lists.


brandonlehmann:
> Can we use lower case if/else to fit with the rest of the file?

Yeah, I'm blind as a bat.

PR #885, opened 2019-09-12
Fix RawBlock Serialization

brandonlehmann:
Fixes an issue whereby a RawBlock is serialized in the RPC call '/getblocks' that results in the transactions property only containing ONE transaction in the block. This creates the transactions property as an array and loads the transaction information (blob + size) into properties of an object that is added to the array of transactions.

In my code tracing, I was not able to find RawBlock serialization used like this anywhere else in the code other than the RPC call. I have maintained synchronization with the network after these changes, the DB storage all still works, and everything appears operational; however, additional sets of eyes on this and testing would be much appreciated to ensure that I did not miss something.

### DIFF
--- a/src/rpc/CoreRpcServerCommandsDefinitions.h
+++ b/src/rpc/CoreRpcServerCommandsDefinitions.h
@@ -68,9 +68,12 @@ namespace CryptoNote
                 block_ids; //*first 10 blocks id goes sequential, next goes in pow(2,n) offset, like 2, 4, 8, 16, 32, 64
                            // and so on, and the last one is always genesis block */
 
+            uint32_t blockCount;
+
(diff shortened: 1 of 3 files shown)


brandonlehmann:
I think I may want to add a `blockCount` parameter to this.

brandonlehmann:
> I think I may want to add a `blockCount` parameter to this.

Added

brandonlehmann:
Been running this approximately 48 hours and no issues with sync or otherwise.

zpalmtree:
Is this ready to merge, or are you doing more testing or considering modifying it further?

brandonlehmann:
I’m good with it. Was waiting to see if anyone else tried it but #yolo.

PR #890, opened 2019-09-29
Resolves #889

### LINKED ISSUE #889: p2p-external-port out of range
Seems there is some missed check on possibile allowable ports used for deamons nat settings.
Is possibile to set p2p-external-port on out of range port, example :3419769.
The port (  3419769 ) is out of 0-65535.

Currently checking the on-line peers list, some peers have port out-of-range.

brandonlehmann:
Might as well check all of the port values to be safe.

### DIFF
--- a/src/daemon/Daemon.cpp
+++ b/src/daemon/Daemon.cpp
@@ -195,6 +195,24 @@ int main(int argc, char *argv[])
         }
     }
 
+    if (config.p2pPort < 0 || config.p2pPort > 65535)
+    {
+        std::cout << "P2P Port must be between 0 and 65,535" << std::endl;
+        exit(1);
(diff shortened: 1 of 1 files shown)


src/daemon/Daemon.cpp
```diff
@@ -195,6 +195,24 @@ int main(int argc, char *argv[])
         }
     }
 
+    if (config.p2pPort < 0 || config.p2pPort > 65535)
```
zpalmtree: 0 should not be allowed surely?

brandonlehmann: Technically, the RFC says that `0` is legit. Whether or not it’s usable in practice is a different story :)

PR #895, opened 2019-10-04
Check if file is a directory before attempting to open it

### LINKED ISSUE #894: zedwallet-beta segfaults if you attempt to open a directory
### System Information

**Operating System**:  Arch Linux
**OS Version**:  latest
**CPU Make, Model, & Speed**:  intel i7 7700k
**Amount of RAM**:  16gb
**Hard Drive Type**:  ssd
**Hard Drive Size**:  2tb
**Volume Type**:  ext4

### TurtleCoin Build Information

**Version:**  v0.18.2.1292
**Binary Source:**  self-compiled from AUR
**If self compiled, please provide all the build commands you used**
```
yay -S turtlecoin-git
```

### Issue Report

**Binary Reported:**  zedwallet-beta

#### Expected Behavior

zedwallet should report if the wallet file entered is actually a directory instead of segfaulting

#### Current Behavior

If you happen to enter a directory name as the wallet filename on the open comand, you get an uncaught error and zedwallet-beta segfaults

```
1	open                     Open a wallet already on your system
 2	create                   Create a new wallet
 3	seed_restore             Restore a wallet using a seed phrase of words
 4	key_restore              Restore a wallet using a view and spend key
 5	view_wallet              Import a view only wallet
 6	exit                     Exit the program

What would you like to do?: 1
What is the name of the wallet you want to open?: test
Enter password: *
Unexpected error: basic_filebuf::underflow error reading the file: Is a directory
Please report this error, and what you were doing to cause it.
Hit enter to exit: 
Segmentation fault (core dumped)
```

##### Screen Shots

image

#### Steps to Reproduce

1. open zedwallet-beta
2. press 1
3. enter a directory name in the current folder
4. enter a password

#### Possible Solution

catch this error and notify the user instead of segfaulting

ExtraHash:
resolves #894

### DIFF
--- a/src/zedwallet++/Open.cpp
+++ b/src/zedwallet++/Open.cpp
@@ -358,17 +358,12 @@ std::string getExistingWalletFileName(const ZedConfig &config)
             {
                 std::cout << WarningMsg("\nWallet name can't be blank! Try again.\n\n");
             }
-            /* Check and make sure it's not a directory name that's been entered */
-            else if (fs::is_directory(walletName))
-            {
-                std::cout << WarningMsg("\nYou've entered a directory, not a wallet file. Try again.\n\n");
-            }
             /* Allow people to enter wallet name with or without file extension */
-            else if (fs::exists(walletName))
+            else if (fs::exists(walletName) && !fs::is_directory(walletName))
             {
                 return walletName;
             }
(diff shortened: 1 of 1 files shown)


ExtraHash:
OK, i believe this should work in all cases now

PR #899, opened 2019-10-06
Enforce max output sizes for both client + node

zpalmtree:
Offers separate limits for network and client, client limits ensures users do not create outputs that are unspendable, and network limit ensures outputs that are impossible to spend are not created.

### DIFF
--- a/src/config/CryptoNoteConfig.h
+++ b/src/config/CryptoNoteConfig.h
@@ -165,7 +165,7 @@ namespace CryptoNote
         /* 30 trillion atomic, or 300 billion TRTL */
         /* This is enforced on the daemon side. An output > 300 billion causes
          * an invalid block. */
-        const uint64_t MAX_OUTPUT_SIZE_NODE   = 300'000'000'000'00;
+        const uint64_t MAX_OUTPUT_SIZE_NODE   = 250'000'000'000'00;
 
         /* 100 billion atomic, or 1 billion TRTL */
         /* This is enforced on the client side. An output > 1 billion will not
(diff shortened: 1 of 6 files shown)


src/config/CryptoNoteConfig.h
```diff
@@ -162,6 +162,18 @@ namespace CryptoNote
 
         const uint64_t MAX_EXTRA_SIZE_V2_HEIGHT = 1300000;
 
+        /* 30 trillion atomic, or 300 billion TRTL */
+        /* This is enforced on the daemon side. An output > 300 billion causes
+         * an invalid block. */
+        const uint64_t MAX_OUTPUT_SIZE_NODE   = 300'000'000'000'00;
```
brandonlehmann: Actually, > 250'000'000'000'00 will cause the same problem won't it?

zpalmtree: `250'000'000'000'00` is an output of `200'000'000'000'00` and an output of `50'000'000'000'00` ;)

brandonlehmann: how do you explain those outputs that aren't pretty amounts in the chain then mister? :)

zpalmtree: touché

ExtraHash:
What happens to the one already created 1 billion output, would this render it unspendable?

zpalmtree:
> What happens to the one already created 1 billion output, would this render it unspendable?
Can you do math?

brandonlehmann:
> What happens to the one already created 1 billion output, would this render it unspendable?

1bil < 100bil

zpalmtree:
Removing WIP, have tested both client side and node side and it functions as I expected.

src/config/CryptoNoteConfig.h
brandonlehmann: I'll ask again -- what prevents someone from making a `non-pretty` output that exceeds 250bil?

zpalmtree: Fixed

PR #900, opened 2019-10-06
Improve print_pool_sh output

zpalmtree:
Example: 
image

### DIFF
--- a/src/daemon/DaemonCommandsHandler.cpp
+++ b/src/daemon/DaemonCommandsHandler.cpp
@@ -17,6 +17,7 @@
 #include <serialization/SerializationTools.h>
 #include <utilities/ColouredMsg.h>
 #include <utilities/FormatTools.h>
+#include <utilities/Utilities.h>
 
 namespace
 {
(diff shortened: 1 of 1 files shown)


0des:
Could this output be more condensed if color is used to convey without linebreaks whether a tx is fusion?

zpalmtree:
> Could this output be more condensed if color is used to convey without linebreaks whether a tx is fusion?

I was thinking perhaps doing `Hash: xxx, Fusion: Yes/No` on one line, but I think expecting people to know whether red/green means fusion or not is a bit presumptious, or do you have some other method in mind?

0des:
An idea would be, first line after `pool state` can say:
Transactions in green are marked as fusion
Transactions in blue are normal fee-paid transactions

Then omit the word `hash:` and print each hash on sequential lines without breaks in the colors mentioned above to denote state of the isFusion check

Further, if the pool is larger than a useful amount of transactions to print (as was the case earlier this evening), an abbreviated list could be printed along with an ellipses (well-understood international context clue) and prompt to display or output a full non-truncated list.

zpalmtree:
> An idea would be, first line after `pool state` can say:
> Transactions in green are marked as fusion
> Transactions in blue are normal fee-paid transactions
> 
> Then omit the word `hash:` and print each hash on sequential lines without breaks in the colors mentioned above to denote state of the isFusion check
> 
> Further, if the pool is larger than a useful amount of transactions to print (as was the case earlier this evening), an abbreviated list could be printed along with an ellipses (well-understood international context clue) and prompt to display or output a full non-truncated list.

The main reason I decided to do this PR is as you mentioned, the list being indigestible, and so adding a summary so I can easily see how many transactions there are in the pool.

If you did add a key, probably would be a good idea to also print it at the summary at the bottom, so you don't have to scroll up a ton to see it.

I put the fusion output on the same line, and removed the line breaks, which makes the output a lot more condensed. Does this look more agreeable?

zpalmtree:
image

Whoops, hit enter too fast

0des:
Indeed, this is more condensed.

PR #902, opened 2019-10-07
Rework status command

zpalmtree:
image


0des:
Is this expanded to two columns horizontally when the terminal dimensions allow for it?

zpalmtree:
> Is this expanded to two columns horizontally when the terminal dimensions allow for it?

No, I'm not aware of a cross platform way to query terminal info - and frankly, I think it probably looks better kept as one column?

0des:
Thank you.

wrkzdev:
I like this one :)

wrkzdev:
Can we have tx pool in `status` also?

zpalmtree:
> Can we have tx pool in `status` also?

Just the number of transactions in the pool?

zpalmtree:
Added the number of transactions in the pool to the table.

PR #904, opened 2019-10-07
Transaction Pool Validation Improvements on addBlock()

brandonlehmann:
Replaced the actualizePoolTransactions & actualizePoolTransactionsLite that is used when a new block is added to the TOP of the chain with another method that takes the pre-computed transaction validator state from a new block then uses that state to perform the following:

- Retrieves the full list of transaction hashes from the current pool
- Computes the new maximum transaction size (bytes)
- Loops through the transactions in the pool to identify any that are now invalid with the addition of the new block. The transaction is removed from the transaction pool if one of the following conditions are met:
  * If the transaction mixins (ring size) is now invalid
  * If the transaction size (bytes) is now too large
  * If the transaction contains outputs that were just spent

It skips the following checks that are redundant as the transaction is already in the pool and the checks have already been completed.

- Does not make sure that tx_extra is less than the configured limit
  - Note: the transaction itself doesn't change just because a block is added
- Does not verify that the correct network transaction fee has been included
  - Note: the transaction itself doesn't change just because a block is added
- Does not verify that the offsets (global indexes) exist
  - Note: unneeded as we're only triggering on adding block to the top not a reorg
- Does not verify that the outputs being spent are unlocked
  - Note: the transaction itself doesn't change just because a block is added
- Does not verify the ring signature(s) are correct
  - Note: the transaction itself doesn't change just because a block is added & unneeded as we're only triggering on adding block to the top not a reorg

I've tested this locally for a bit and I'm maintaining sync. the addBlock routine is running much faster as we aren't getting held up waiting for every transaction in the mempool to *deep* validate again.

### DIFF
--- a/src/cryptonotecore/Core.cpp
+++ b/src/cryptonotecore/Core.cpp
@@ -1244,13 +1244,10 @@ namespace CryptoNote
 
                     updateBlockMedianSize();
 
-                    /* We've used these transactions, remove them from the pool if they are there */
-                    for (const auto tx : transactions)
-                    {
-                        transactionPool->removeTransaction(tx.getTransactionHash());
-                    }
-
-                    actualizePoolTransactionsLite(validatorState);
+                    /* Take the current block spent key images and run them
+                       against the pool to remove any transactions that may
+                       be in the pool that would now be considered invalid */
+                    checkAndRemoveInvalidPoolTransactions(validatorState);
 
                     ret = error::AddBlockErrorCode::ADDED_TO_MAIN;
                     logger(Logging::DEBUGGING) << "Block " << blockStr << " added to main chain.";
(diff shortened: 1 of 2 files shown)


src/cryptonotecore/Core.cpp
```diff
@@ -1378,44 +1373,53 @@ namespace CryptoNote
...
-    }
+        const auto maxTransactionSize = getMaximumTransactionAllowedSize(blockMedianSize, currency);
 
-    void Core::actualizePoolTransactionsLite(const TransactionValidatorState &validatorState)
-    {
-        auto &pool = *transactionPool;
-        auto hashes = pool.getTransactionHashes();
+        for (const auto poolTxHash : poolHashes)
+        {
+            const auto poolTx = pool.getTransaction(poolTxHash);
```
zpalmtree: Is it possible that the hash we acquired from `getTransactionHashes()` could no longer be present at this point?

brandonlehmann: I *think* you're asking if this has a chance of throwing, and it's a valid concern; however, I have two reasons why I don't believe it's a concern:

1. This is the exact same pattern for how the transaction is retrieved that was removed from `actualizePoolTransactions` and `actualizePoolTransactionsLite` -- didn't a pose an issue there
2. The daemon is very single-threaded and thus the chances of anything removing it from the pool before this blocking code clears is slim to nil.

If this were doing *deep* validations -- considering the time some take, I'd consider it a major concern; however, if we didn't run into the issue when it was slow, I highly doubt we'll run into it with it being a lot faster.

zpalmtree: Alright.

brandonlehmann: If you want, I can throw a `checkIfTransactionPresent` in there :)

PR #909, opened 2019-10-08
Moving /queryblocksdetailed behind a separate cli flag

brandonlehmann:


DroppingThePacketsHard:
Why? Does this have any purpose?

brandonlehmann:
> Why? Does this have any purpose?

Wouldn’t have submitted it if it didn’t have a purpose :)

DroppingThePacketsHard:
I don't like to have to type more than I need 😄

0des:
You also don't seem to like to read. This commit looks similar to other work we've seen recently. Hmm.

brandonlehmann:
> I don't like to have to type more than I need 😄

Use config files.

DroppingThePacketsHard:
> Use config files.
Aight, will do that 👍

PR #911, opened 2019-10-14
CORS policy: Request header field content-type is not allowed - preflight checks

kryptonchain:
- As discussed on discord with The man on the Clapham omnibus AKA Zpalmtree when the wallet backend is purely used in JS/browser environment some browsers are doing preflight/OPTIONS checks to the daemon.

These checks are failing in the jsonMethod function..

This is a fix similar to: https://github.com/turtlecoin/turtlecoin/blob/development/src/walletapi/ApiDispatcher.cpp#L1412
code.


zpalmtree:
> Should we also add `Access-Control-Allow-Methods` as well to be safe? I think `POST, GET, OPTIONS` is enough.

To be safe we need to add all the stuff that wallet-api adds, aye.. on the other hand, I've already implemented this perfectly in the RPC rewrite so feels like maybe a waste of time striving for perfection for code that's gonna get removed quite soon?

brandonlehmann:
> soon

Til then... ? :)

ExtraHash:
As  a temporary solution, I've used something like apache to reverse proxy the rpc port and add the headers myself

kryptonchain:
Yes I've also tried the reversed proxy solution, however it breaks the concept that everyone should be able to run a node easily.

My idea was with the js app that I can pick the node to connect from a list of nodes according maybe geo proximity or ping.
So anyone could run a node and then the js app can connect the user to the closest one for syncing or signing a transaction. 
But it requires the ability being able to easily launch a full node. If it requires tweaking apache with reverse proxy probably requires bit more advanced user then just someone who wants to support the network for example. 

In theory I could even work out the peers who are connected with - - enable-blockexplorer - - enable-cors option from /peers and then auto add them to the list of nodes maybe based on some condition like they not have some crazy fee set for example

kryptonchain:
Thanks @brandonlehmann for cleaning it up and merging + @zpalmtree for the tip!

PR #919, opened 2019-10-23
Catch & Re-Throw Deserialization Issues of MainChainStorage

brandonlehmann:
Provides a helpful message when data cannot be properly de-serialized from MainChainStorage

### DIFF
--- a/src/cryptonotecore/MainChainStorage.cpp
+++ b/src/cryptonotecore/MainChainStorage.cpp
@@ -7,6 +7,7 @@
 
 #include "common/CryptoNoteTools.h"
 #include "common/FileSystemShim.h"
+#include <sstream>
 
 namespace CryptoNote
 {
@@ -52,7 +53,26 @@ namespace CryptoNote
                 + " is out of range. Blocks count: " + std::to_string(storage.size()));
         }
 
-        return storage[index];
+        try
+        {
+            return storage[index];
+        }
+        catch (std::exception &)
+        {
+            /* Intercept the exception here and display a friendly help message */
+            std::stringstream errorMessage;
+
+            errorMessage << "Local blockchain cache corruption detected." << std::endl
+                         << "Block with index " << std::to_string(index)
+                         << " could not be deserialized from the blockchain cache."
+                         << std::endl << std::endl
+                         << "Please try to repair this issue by starting the node with the option: "
+                         << "--rewind-to-height " << std::to_string(index - 1) << std::endl
+                         << "If the above does not repair the issue, "
+                         << "please launch the node with the option: --resync" << std::endl;
+
+            throw std::runtime_error(errorMessage.str());
+        }
     }
 
     uint32_t MainChainStorage::getBlockCount() const


src/cryptonotecore/MainChainStorage.cpp
```diff
@@ -52,7 +53,30 @@ namespace CryptoNote
...
         }
 
-        return storage[index];
+        RawBlock result;
+
+        try
+        {
+            result = storage[index];
+        }
+        catch (...)
```
zpalmtree: Can this be `std::exception &` instead of `(...)`? Or do we envisage catching exceptions which don't inherit from `std::exception`, like boost exceptions? If that's the case, I'd prefer to see an extra catch clause for those.

src/cryptonotecore/MainChainStorage.cpp
```diff
@@ -52,7 +53,30 @@ namespace CryptoNote
                 + " is out of range. Blocks count: " + std::to_string(storage.size()));
         }
 
-        return storage[index];
+        RawBlock result;
+
+        try
+        {
+            result = storage[index];
```
zpalmtree: We could maybe just do `return storage[index]` here if we're not using the variable anywhere else.

brandonlehmann: Fair.

src/cryptonotecore/MainChainStorage.cpp
brandonlehmann: This catches about 3 or 4 different possible errors. I understand where you’re coming from but the result is the same regardless for what caused it.

zpalmtree: It should catch every normal exception, given they all inherit from `std::exception`. Unless you have reason to believe they're throwing `std::string` or boost exceptions, this will catch all the possible errors.

src/cryptonotecore/MainChainStorage.cpp
brandonlehmann: Tweaked.

src/cryptonotecore/MainChainStorage.cpp
brandonlehmann: Okay, okay. Let's see how this plays out.

src/cryptonotecore/MainChainStorage.cpp
```diff
@@ -52,7 +53,28 @@ namespace CryptoNote
...
+                         << std::endl << std::endl
+                         << "Please try to repair this issue by starting the node with the option: "
+                         << "--rewind-to-height " << std::to_string(index - 1) << std::endl
+                         << "If the above does not repair the issue, "
+                         << "please launch the node with the option: --resync" << std::endl;
+
+            throw std::runtime_error(errorMessage.str());
+        }
+
+        return RawBlock();
```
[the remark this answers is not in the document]
brandonlehmann: I read it on the internet.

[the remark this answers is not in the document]
brandonlehmann: Done

brandonlehmann:
GH actions Windows edition can get lost.

PR #931, opened 2019-11-13
Allow Arbitrary Data in Transaction Extra Nonce via Wallet-API

brandonlehmann:
This allows for the addition of arbitrary hexadecimal data to be added to a transaction nonce via wallet-api.

The additional body parameter `extra` has been added to support this.

Once supplied, the data is encoded behind a special nonce tag (`127`) and appended to the nonce such that:

<tag><varint encoded length of data><data>

Example: https://explorer.turtlecoin.lol/transaction.html?hash=5591fc3bb7384d304341066b214635687c615c860b17024d8cdb593e23aae017

Note: This does not remove the 1,024 byte limit for the TX_EXTRA so that limit still applies.

This matches how I handled this in [turtlecoin-utils](https://github.com/turtlecoin/turtlecoin-utils)

### DIFF
--- a/src/utilities/ParseExtra.cpp
+++ b/src/utilities/ParseExtra.cpp
@@ -6,6 +6,7 @@
 #include <utilities/ParseExtra.h>
 /////////////////////////////////
 
+#include <common/Varint.h>
 #include <config/Constants.h>
 
 namespace Utilities
@@ -28,18 +29,26 @@ namespace Utilities
         return parsed.mergedMiningTag;
     }
 
+    std::vector<uint8_t> getExtraDataFromExtra(const std::vector<uint8_t> &extra)
+    {
+        const ParsedExtra parsed = parseExtra(extra);
+        return parsed.extraData;
+    }
+
     ParsedExtra parseExtra(const std::vector<uint8_t> &extra)
     {
         ParsedExtra parsed {Constants::NULL_PUBLIC_KEY, std::string(), {0, Constants::NULL_HASH}};
 
         bool seenPubKey = false;
+        bool seenNonce = false;
+        bool seenExtraData = false;
         bool seenPaymentID = false;
         bool seenMergedMiningTag = false;
 
         for (auto it = extra.begin(); it != extra.end(); it++)
         {
             /* Nothing else to parse. */
-            if (seenPubKey && seenPaymentID && seenMergedMiningTag)
+            if (seenPubKey && seenPaymentID && seenMergedMiningTag && seenExtraData)
             {
                 break;
             }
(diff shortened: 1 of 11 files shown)


src/walletbackend/Transfer.cpp
```diff
@@ -1011,24 +1019,51 @@ namespace SendTransaction
...
+        if (!extraData.empty())
+        {
+            /* Indicate this is arbitrary data */
+            extraNonce.push_back(Constants::TX_EXTRA_ARBITRARY_DATA_IDENTIFIER);
+
+            /* Determine the length of the data and varint encode it */
+            std::vector<uint8_t> extraDataSize = Tools::uintToVarintVector(extraData.size());
+
+            /* Write the length of the data out to extra */
+            std::copy(std::begin(extraDataSize), std::end(extraDataSize), std::back_inserter(extraNonce));
```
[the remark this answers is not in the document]
brandonlehmann: Probably.

src/walletbackend/WalletBackend.cpp
```diff
@@ -779,10 +780,11 @@ std::tuple<Error, Crypto::Hash> WalletBackend::sendFusionTransactionBasic()
 std::tuple<Error, Crypto::Hash> WalletBackend::sendFusionTransactionAdvanced(
     const uint64_t mixin,
     const std::vector<std::string> subWalletsToTakeFrom,
-    const std::string destination)
+    const std::string destination,
+    const std::vector<uint8_t> extraData)
```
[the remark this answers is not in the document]
brandonlehmann: They are advanced :)

src/walletbackend/Transfer.cpp
```diff
@@ -1011,24 +1019,51 @@ namespace SendTransaction
...
+
+        if (!extraNonce.empty())
+        {
             /* Indicate this is the extra nonce */
             extra.push_back(Constants::TX_EXTRA_NONCE_IDENTIFIER);
 
-            /* Add the length of the extra nonce (PID tag + PID length == 33) */
-            extra.push_back(1 + 32);
+            /* Determine the length of the nonce data and varint encode it */
+            std::vector<uint8_t> extraNonceSize = Tools::uintToVarintVector(extraNonce.size());
```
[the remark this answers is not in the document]
brandonlehmann: Plausible as on second read through that’s a pretty rigid read. I’ll update ParseExtra for more flexibility.

src/walletbackend/Transfer.cpp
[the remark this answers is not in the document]
brandonlehmann: Yes. Done.

src/walletbackend/Transfer.cpp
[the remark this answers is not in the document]
brandonlehmann: Reworked ParseData a bit. I believe this handles this question.

src/walletbackend/Transfer.cpp
```diff
@@ -26,15 +27,18 @@ namespace SendTransaction
            as the static constructors were used */
         const std::string defaultAddress = subWallets->getPrimaryAddress();
 
-        return sendFusionTransactionAdvanced(defaultMixin, {}, defaultAddress, daemon, subWallets);
+        const std::vector<uint8_t> extraData;
+
+        return sendFusionTransactionAdvanced(defaultMixin, {}, defaultAddress, daemon, subWallets, extraData);
```
zpalmtree: If we want to be terser here, we could just do 
```suggestion
        return sendFusionTransactionAdvanced(defaultMixin, {}, defaultAddress, daemon, subWallets, {});
```

Might complain if we're taking a reference, though.

src/utilities/ParseExtra.cpp
```diff
@@ -69,36 +78,86 @@ namespace Utilities
...
+            /* Found nonce information and need to decode it.
+            /* Nonce is a sub-tagged field and thus we need to work through
+               the data to determine what fields are in here */
+            if (c == Constants::TX_EXTRA_NONCE_IDENTIFIER && elementsRemaining > 1 && !seenNonce)
+            {
+                /* Get the length of the following data in the field */
+                size_t nonceSize = 0;
 
-            /* Payment ID looks like this (payment ID is stored in extra nonce)
+                const auto readNonceSize = Tools::read_varint(it + 1, extra.end(), nonceSize);
```
zpalmtree: Am I correct in believing that `readNonceSize` is the size of the varint holding the `nonceSize`, while `nonceSize` is the size of the following data?

src/utilities/ParseExtra.cpp
```diff
@@ -69,36 +78,86 @@ namespace Utilities
...
+                        if (s == Constants::TX_EXTRA_ARBITRARY_DATA_IDENTIFIER && nElementsRemaining > 1 && !seenExtraData)
+                        {
+                            /* Read out the size of the data */
+                            size_t dataSize = 0;
+
+                            const auto readDataSize = Tools::read_varint(is + 1, nonceData.end(), dataSize);
+
+                            /* If there are enough bytes left to read based upon the size above then
+                               read out the data */
+                            if (nElementsRemaining >= 1 + dataSize)
```
zpalmtree: Shouldn't this be 
```suggestion
                            if (nElementsRemaining >= 1 + dataSize + readDataSize)
```
to include the length of the varint?

src/utilities/ParseExtra.cpp
brandonlehmann: Yes. The first is the number or bytes consumed in doing so. The second is the actual value.

src/utilities/ParseExtra.cpp
brandonlehmann: Yes, fixed locally, will push as others are addressed.

brandonlehmann: Pushed

src/walletbackend/Transfer.cpp
brandonlehmann: Worked for both the sendFusionTransactionBasic and sendTransactionBasic. Updated.

brandonlehmann:
> This parsing code looks pretty good to me, but have you tested it with slightly too short/too long lengths? It's very easy to get off by one errors in code like this.

I've ran it through as many tests as I possibly can of long/short values and while it does not throw an error -- improperly constructed fields will lead to unexpected behavior.

Example: If you put the public key behind the nonce field, and you say the nonce field is 25-bytes long but only supply 24-bytes, the public key will **not** be read as the identifier `01` will be considered part of the nonce field. Little things like that. It's pretty much the same issue as if someone were to construct extra with garbage data at the front and the parser would find `01` and the next 32-bytes are considered the public key -- although the *real* public key is farther back in the array. Garbage in, garbage out.

All that being said, I've moved this to a WIP as I'd like to extend wallet-api to be able to return this information when getting transaction information; however, I haven't started on that just yet.

brandonlehmann:
Removed WIP tag. Storing and retrieving the data in wallets, as an option, will be addressed in another PR.

PR #933, opened 2019-11-16
Daemon+WalletBackend timestamp adjustments

### LINKED ISSUE #704: Daemon+WalletBackend timestamp adjustments
The current /getwalletsyncdata rounds a timestamp to midnight. Depending on what time of the day you start a fresh wallet, you may have no blocks to grab (we need to roll back a bit more than we currently do with the timestamp adjustment), or too many (since it's rounding to midnight which is quite far away).

We should instead just respect the timestamp given, and have a better adjust so we always sync about 100 or so blocks to be safe.

Might require a bit of work, since the timestamp fetching routines I believe hit the two DB caches, which aren't the most fun to work with.

brandonlehmann:
Resolves #704

### DIFF
--- a/src/cryptonotecore/DatabaseBlockchainCache.cpp
+++ b/src/cryptonotecore/DatabaseBlockchainCache.cpp
@@ -200,16 +200,6 @@ namespace CryptoNote
             return requestExtendedTransactionInfos(transactionHashes, database, result);
         }
 
-        uint64_t roundToMidnight(uint64_t timestamp)
-        {
-            if (timestamp > static_cast<uint64_t>(std::numeric_limits<time_t>::max()))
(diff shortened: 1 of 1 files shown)


zpalmtree:
Imo, we should not round the timestamp. We already round it here - https://github.com/turtlecoin/turtlecoin/blob/development/src/utilities/Utilities.cpp#L170 and can adjust it more if needed.

brandonlehmann:
> Imo, we should not round the timestamp. We already round it here - https://github.com/turtlecoin/turtlecoin/blob/development/src/utilities/Utilities.cpp#L170 and can adjust it more if needed.

Well okay then. Removed entirely.

PR #934, opened 2019-11-16
Remove legacy Zedwallet and promote Zedwallet++ out of beta

brandonlehmann:


src/serialization/SerializationTools.h
```diff
@@ -18,7 +18,7 @@
 #include <serialization/BinaryOutputStreamSerializer.h>
 #include <serialization/CryptoNoteSerialization.h>
 #include <vector>
-#include <zedwallet/Types.h>
+#include <ZedwalletTypes.h>
```
zpalmtree: Can we not just remove the serialization stuff for zedwallet, or is it used elsewhere?

brandonlehmann: I did, then the rabbit hole went deeper.

brandonlehmann: I figure it can be totally removed once we do the same to turtle-service some day.

zpalmtree: mehhhh

brandonlehmann: If you’d like, I’ll get my helmet out.

zpalmtree: Nah, I can handle it in a follow up PR, I probably know where it's used better

Issue #842, opened 2019-06-27
Key images are regenerated on transaction creation [WB]


zpalmtree:
Despite generating the key images when we receive a transaction, we also do this when making the transaction, which is unnecessary and slows down the process.

PR #937, opened 2019-11-18
Don't regenerate key images where not necessary

zpalmtree:
Resolves: #842


brandonlehmann:
CI webook was stuck but was completed.

PR #903, opened 2019-10-07
Full RPC rewrite with threading

zpalmtree:
A full RPC rewrite. Runs the RPC on a separate thread.

TODO: Synchronize data races, testing

Resolves: #603

PR #942, opened 2019-11-22
Fix a bug that prevents blockchain to be loaded if the blockindexes.bin corrupts

ngeojiajun:
blockindexes.bin hold the indexes for each block in the main storage. The file began with 8 bytes with type of uint64_t to represent the count of the entries. After that, for each entry 4 bytes used to store it. The original code will fail the loading if to read fails. But if the indexes file was corrupt due to the incomplete write then the file should be recoverable by updating the count of entries to the correct value.

In this fix, if the EOF comes before the indexes fully loaded, instead of failing it the code will automatically fix it by simply terminate the loop and allow the deamon still working.

### DIFF
--- a/src/cryptonotecore/SwappedVector.h
+++ b/src/cryptonotecore/SwappedVector.h
@@ -9,10 +9,10 @@
 #include "common/StdOutputStream.h"
 #include "serialization/BinaryInputStreamSerializer.h"
 #include "serialization/BinaryOutputStreamSerializer.h"
+#include "logger/Logger.h"
 #include <cstddef>
 #include <cstdint>
 #include <fstream>
-#include <cstdio>
 #include <iomanip>
 #include <iostream>
 #include <list>
@@ -234,7 +234,7 @@ template<class T> SwappedVector<T>::~SwappedVector()
 template<class T>
 bool SwappedVector<T>::open(const std::string &itemFileName, const std::string &indexFileName, size_t poolSize)
 {
-    if (poolSize == 0)
+	if (poolSize == 0)
     {
         return false;
     }
@@ -263,8 +263,8 @@ bool SwappedVector<T>::open(const std::string &itemFileName, const std::string &
 					return false;
 				}
 				else {
-					i--;
-					std::cout<<"Detected EOF before all indexes begin logged. Adjusting index counts to "<<i<<std::endl;
+					Logger::logger.log("Blockchain indexes file appears to be corrupted. Attempting automatic recovery by rewinding to " + std::to_string(i),
+						Logger::WARNING, { Logger::FILESYSTEM});
 					break;
 				}
             }
@@ -272,7 +272,7 @@ bool SwappedVector<T>::open(const std::string &itemFileName, const std::string &
             offsets.emplace_back(itemsFileSize);
             itemsFileSize += itemSize;
         }
-
+		
         m_offsets.swap(offsets);
         m_itemsFileSize = itemsFileSize;
     }
--- a/src/cryptonotecore/SwappedVector.h
+++ b/src/cryptonotecore/SwappedVector.h
@@ -265,6 +265,9 @@ bool SwappedVector<T>::open(const std::string &itemFileName, const std::string &
 				else {
 					Logger::logger.log("Blockchain indexes file appears to be corrupted. Attempting automatic recovery by rewinding to " + std::to_string(i),
 						Logger::WARNING, {Logger::FILESYSTEM,Logger::DATABASE});
+					//Reopen the indexes file to force changes to be commited
+					m_indexesFile.close();
+					m_indexesFile.open(indexFileName, std::ios::in | std::ios::out | std::ios::binary);
 					break;
 				}
             }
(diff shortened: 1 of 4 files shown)


src/cryptonotecore/SwappedVector.h
```diff
@@ -256,9 +256,17 @@ bool SwappedVector<T>::open(const std::string &itemFileName, const std::string &
...
             m_indexesFile.read(reinterpret_cast<char *>(&itemSize), sizeof itemSize);
+
             if (!m_indexesFile)
             {
-                return false;
+				if (!m_indexesFile.eof()) { //fail it only if the other IO occured
+					return false;
+				}
+				else {
+					i--;
```
zpalmtree: I'm not sure what we're trying to achieve with this line. `i` seems to only be used in the loop, and just for tracking the number of iterations. But, we break right after, so adjusting `i` seems to only be needed to make the message printed lower correct. Could we not just use `i - 1` instead there?

src/cryptonotecore/SwappedVector.h
```diff
@@ -256,9 +256,17 @@ bool SwappedVector<T>::open(const std::string &itemFileName, const std::string &
...
+					std::cout<<"Detected EOF before all indexes begin logged. Adjusting index counts to "<<i<<std::endl;
```
zpalmtree: Generally, we try and not use `std::cout` to log directly, instead we prefer to use a `Logger` instance.
You could add `#include <logger/Logger.h>` to the preamble, then instead do something like
```cpp
Logger::logger.log(
    "Blockchain indexes file appears to be corrupted. Attempting automatic recovery by rewinding indexes to " + std::to_string(i - 1),
    Logger::WARNING,
    { LogCategory::DATABASE }
);
```
I tried to modify the message to be a bit more understandable to neophytes.
I also added a new `LogCategory`, `DATABASE` here - you can add categories here: https://github.com/turtlecoin/turtlecoin/blob/46efa51534b2f094b98e6082bb0dc3654e7e8568/src/logger/Logger.h#L22-L29
Feel free to modify to add multiple `LogCategory`'s if you think something else will be more informative - a message can be logged with multiple categories.

src/cryptonotecore/SwappedVector.h
ngeojiajun: Yes, should be ok

src/cryptonotecore/SwappedVector.h
ngeojiajun: Ok, i will reflect the changes later.

src/cryptonotecore/SwappedVector.h
ngeojiajun: after some testing the "i" shouldn't be adjusted

src/cryptonotecore/SwappedVector.h
ngeojiajun: I attempted it but got no output in neither log file nor console

zpalmtree: By default, the logger is disabled. Right now, it's not hooked up in the daemon executable, but I handled the hooking up in this pull request: https://github.com/turtlecoin/turtlecoin/pull/903/files#diff-38f2ab37cf3375846c77f4fd2aff8e50R241

I figured since that one is close to go, there's no point you duplicating the effort and adding merge conflicts.

ngeojiajun: Ok, just wait for that PR to be merged for the logging code to work right?

zpalmtree: Yep. If you want to test it out, you can just add that logging stuff to `Daemon.cpp`. I think we can proceed with merging this if it's ready as the RPC stuff will get merged before next release I think.

src/cryptonotecore/SwappedVector.h
```diff
@@ -256,15 +256,26 @@ bool SwappedVector<T>::open(const std::string &itemFileName, const std::string &
...
             {
-                return false;
+				if (!m_indexesFile.eof()) { //fail it only if the other IO occured
+					return false;
+				}
+				else {
+					Logger::logger.log("Blockchain indexes file appears to be corrupted. Attempting automatic recovery by rewinding to " + std::to_string(i),
+						Logger::WARNING, {Logger::FILESYSTEM,Logger::DATABASE});
+					//Reopen the indexes file to force changes to be commited
+					m_indexesFile.close();
```
zpalmtree: Can you expand on what we're attempting to achieve by closing and reopening the file? It doesn't appear to have been modified in this loop, so I'm not sure how this causes changes to be commited. As far as I can tell, the fix works by causing the `itemsFileSize` and `offsets` to be updated with the truncated values that are actually on disk.

ngeojiajun: Reading after EOF will result an error flag to be set on it. If file IO was attempted after that, it will fail. In another words, the deamon started refusing to add block as the indexes file couldn't be updated.

ngeojiajun: Reopening the file will reset all the flags associated with that stream

zpalmtree: Perhaps we could achieve the same thing by using `.clear()`? A bit more obvious what we're doing, I feel. https://stackoverflow.com/a/7681612/8737306

```cpp
m_indexesFile.clear();
m_indexesFile.seekg(0, std::ios::beg);
```

ngeojiajun: OK, i will update it later

PR #948, opened 2019-11-26
Allow syncing with /getrawblocks instead of /getwalletsyncdata

zpalmtree:
Since wallet syncing is generally limited on the speed of fetching data, we can offload some of the daemon's load by letting the client parse the raw blocks, instead. The effect of this will be most apparent when one daemon is serving many different users wallets.
This can also potentially lower the amount of data transferred.

* Should we make it possible to disable this feature using a command line option?

Note: To test this PR, you must be running a daemon with #903 applied.

### DIFF
--- a/src/cryptonotecore/Core.h
+++ b/src/cryptonotecore/Core.h
@@ -208,6 +208,10 @@ namespace CryptoNote
 
         virtual uint64_t get_current_blockchain_height() const;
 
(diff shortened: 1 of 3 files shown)


brandonlehmann:
Non-optional if you ask me. My base tests before suggesting this is orders of magnitude faster. Trying to work this into other processes as well.

brandonlehmann:
Holding as it has a dependency of #903

PR #959, opened 2019-12-01
Prompt user to add .wallet extension if not present in wallet-upgrader

### LINKED ISSUE #929: Possibly force wallet-upgrader to use .wallet extension
Proton only looks for wallets with a `.wallet` extension. Possibly this is something that should be fixed on their end, or perhaps we should force a `.wallet` extension when upgrading wallets.

ExtraHash:
Resolves #929

### DIFF
--- a/src/walletupgrader/Main.cpp
+++ b/src/walletupgrader/Main.cpp
@@ -10,6 +10,7 @@
 #include <cxxopts.hpp>
 #include <utilities/PasswordContainer.h>
 #include <utilities/ColouredMsg.h>
+#include <utilities/Input.h>
 #include <walletbackend/WalletBackend.h>
 
 int main(int argc, char **argv)
@@ -91,6 +92,29 @@ int main(int argc, char **argv)
                 /* Allow people to enter wallet name with or without file extension */
                 else if (fs::exists(walletName))
                 {
+                    /* We don't want to prompt the user about the rename if the wallet and
+                     * password are given, otherwise it will break non-interactive upgrade
+                     */
+                    if (!walletGiven && !passGiven)
+                    {
+                        bool appendExtension = Utilities::confirm("Wallet does not end in the .wallet extension. This may break compatability with some wallets. Do you want to add the .wallet extension?");
+                        if (appendExtension)
+                        {
+                            int renameSuccess = std::rename(walletName.c_str(), walletFileName.c_str());
+                            if (renameSuccess != 0)
+                            {
+                                std::cout << WarningMsg("Failed to rename file. Do we have permissions to write files in this folder? Exiting.\n");
+                                return 1;
+                            }
+                            filename = walletFileName;
+                            break;
+                        } 
+                        else
+                        {
+                            filename = walletName;
+                            break;
+                        }
+                    }
                     filename = walletName;
                     break;
                 }


src/walletupgrader/Main.cpp
```diff
@@ -91,7 +92,18 @@ int main(int argc, char **argv)
                 /* Allow people to enter wallet name with or without file extension */
                 else if (fs::exists(walletName))
                 {
-                    filename = walletName;
+                    bool appendExtension = Utilities::confirm("Wallet does not end in the .wallet extension. This may break compatability with some wallets. Do you want to add the .wallet extension?", true);
+                    while (true)
```
zpalmtree: We don't need a while loop here.

src/walletupgrader/Main.cpp
```diff
@@ -91,7 +92,18 @@ int main(int argc, char **argv)
...
+                    {
+                        if (appendExtension) {
```
zpalmtree: Please use allman style braces.
For example:
```cpp
if (appendExtension)
{
    // do something
}
else
{
    // do something
}

File: src/walletupgrader/Main.cpp [hunk 3]
```diff
@@ -91,7 +92,18 @@ int main(int argc, char **argv)
                 /* Allow people to enter wallet name with or without file extension */
                 else if (fs::exists(walletName))
                 {
-                    filename = walletName;
+                    bool appendExtension = Utilities::confirm("Wallet does not end in the .wallet extension. This may break compatability with some wallets. Do you want to add the .wallet extension?", true);
```
zpalmtree: We should check that `walletGiven` and `passGiven` aren't both true here, otherwise it will break non interactive upgrading.

src/walletupgrader/Main.cpp
```diff
@@ -91,7 +92,18 @@ int main(int argc, char **argv)
...
+                    while (true)
+                    {
+                        if (appendExtension) {
+                            std::rename(walletName.c_str(), walletFileName.c_str());
```
zpalmtree: We should check the return value of `std::rename` to ensure it has succeeded before continuing.

src/walletupgrader/Main.cpp
zpalmtree: `confirm` defaults to `true`, so the second argument isn't needed.

ExtraHash:
OK, should be good now

src/walletupgrader/Main.cpp
```diff
@@ -91,6 +92,24 @@ int main(int argc, char **argv)
                 /* Allow people to enter wallet name with or without file extension */
                 else if (fs::exists(walletName))
                 {
+                    if (!walletGiven && !passGiven)
+                    {
+                        bool appendExtension = Utilities::confirm("Wallet does not end in the .wallet extension. This may break compatability with some wallets. Do you want to add the .wallet extension?");
+                        if (appendExtension)
+                        {
+                            int renameSuccess = std::rename(walletName.c_str(), walletFileName.c_str());
+                            if (!renameSuccess) {
```
zpalmtree: The braces are wrong here again. Also, can we please write a message to the user here before we quit so they know what's going on? Something like..
"Failed to rename file. Do we have permissions to write files in this folder?"

src/walletupgrader/Main.cpp
```diff
@@ -91,6 +92,24 @@ int main(int argc, char **argv)
                 /* Allow people to enter wallet name with or without file extension */
                 else if (fs::exists(walletName))
                 {
+                    if (!walletGiven && !passGiven)
```
zpalmtree: Could we add a comment for future readers so they know why we're doing this bit?

ExtraHash:
:ok_hand:

src/walletupgrader/Main.cpp
```diff
@@ -91,6 +92,29 @@ int main(int argc, char **argv)
...
+                     */
+                    if (!walletGiven && !passGiven)
+                    {
+                        bool appendExtension = Utilities::confirm("Wallet does not end in the .wallet extension. This may break compatability with some wallets. Do you want to add the .wallet extension?");
+                        if (appendExtension)
+                        {
+                            int renameSuccess = std::rename(walletName.c_str(), walletFileName.c_str());
+                            if (renameSuccess != 0)
+                            {
+                                std::cout << WarningMsg("Failed to rename file. Do we have permissions to write files in this folder? Try again.\n");
```
zpalmtree: "Try again" seems misleading, as they will almost certainly get the same error - unless they figured out they needed to remount their filesystem as writable or something - and if they're smart enough to know how to do that, they can probably figure out they need to re-run the program after doing so :smiley:

src/walletupgrader/Main.cpp
```diff
@@ -91,6 +92,29 @@ int main(int argc, char **argv)
...
+                     */
+                    if (!walletGiven && !passGiven)
+                    {
+                        bool appendExtension = Utilities::confirm("Wallet does not end in the .wallet extension. This may break compatability with some wallets. Do you want to add the .wallet extension?");
+                        if (appendExtension)
+                        {
+                            int renameSuccess = std::rename(walletName.c_str(), walletFileName.c_str());
+                            if (renameSuccess != 0)
+                            {
+                                std::cout << WarningMsg("Failed to rename file. Do we have permissions to write files in this folder? Ping @zpalmtree at the main discord for help.\n");
```
zpalmtree: this is just silly.

zpalmtree:
Please squash on merge if I'm not the one who merges it.

ExtraHash:
squashed

PR #965, opened 2019-12-06
Threaded input validation

zpalmtree:
Adds a threadpool which is then utilized to validate the inputs in a transaction in parallel.

Needs a bit more testing. I was seeing some segfaults on my node which I think may be related, running in debug mode to try and track them down now.

This PR is most helpful when
* Syncing from zero *without checkpoints*
* Syncing top blocks which are outside the checkpoints range

I saw my CPU usage go from 8% (1 core out of 12 fully pegged) to between 30-50% usage while syncing without checkpoints.

I think this same idea may be applicable in other parts of the syncing process.

### DIFF
--- a/src/cryptonotecore/Core.cpp
+++ b/src/cryptonotecore/Core.cpp
@@ -1145,7 +1145,7 @@ namespace CryptoNote
         {
             uint64_t fee = 0;
             auto transactionValidationResult =
-                validateTransaction(transaction, validatorState, cache, fee, previousBlockIndex, false);
+                validateTransaction(transaction, validatorState, cache, m_transactionValidationThreadPool, fee, previousBlockIndex, false);
             if (transactionValidationResult)
             {
                 logger(Logging::DEBUGGING) << "Failed to validate transaction " << transaction.getTransactionHash()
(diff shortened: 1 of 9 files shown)


brandonlehmann:
Initial thoughts are that we should extend the option for the number of threads to the daemon CLI arguments as part of this process.

ngeojiajun:
I tried this locally but it stucks at
>2019-Dec-08 07:32:15.414456 DEBUG   Request to add block 2083017 (54a6e068169886397a29270e56bc6ebf6f7a6857326db30f89531e0e36dd6e66)

without any cpu usage and disk activity

zpalmtree:
> I tried this locally but it stucks at
> 
> > 2019-Dec-08 07:32:15.414456 DEBUG   Request to add block 2083017 (54a6e068169886397a29270e56bc6ebf6f7a6857326db30f89531e0e36dd6e66)
> 
> without any cpu usage and disk activity

Is this reproducible?
Only thing I could imagine to cause that is the threads are never waking up to process the job.

ngeojiajun:
I done no specific action and just replace the turtlecoind.exe with one that applied this PR then that happens.
Here the full command line:
>Turtlecoind --db-max-open-files 1200 --data-dir <***>  --db-write-buffer-size 1200 --db-enable-compression --log-level 3

zpalmtree:
> Initial thoughts are that we should extend the option for the number of threads to the daemon CLI arguments as part of this process.

Done

> I done no specific action and just replace the turtlecoind.exe with one that applied this PR then that happens.

I tried running with a single thread and could not reproduce this. Also couldn't see anything that would cause a deadlock :/

ngeojiajun:
> > Initial thoughts are that we should extend the option for the number of threads to the daemon CLI arguments as part of this process.
> 
> Done
> 
> > I done no specific action and just replace the turtlecoind.exe with one that applied this PR then that happens.
> 
> I tried running with a single thread and could not reproduce this. Also couldn't see anything that would cause a deadlock :/

i will retry after rebuilt the file from scratch

zpalmtree:
Bug mentioned by @ngeojiajun should now be fixed by 5344bff70d2b1c407d2f7b6340d6b8258eaaf1f7.

ngeojiajun:
local test completed. the block were verified and added correctly

PR #967, opened 2019-12-15
Fee per byte

zpalmtree:
#### TODO:
* ~Input ordering for WalletBackend - Suggested method is picking from increasing base10 buckets in order, then wrapping back again to the smallest bucket till amount is fulfilled. This should allow a decent diversity of inputs, while allowing large transactions to be fulfilled and transaction estimates to be relatively accurate.~ Done

* Testing for edge cases

* Testing prepared/delayed transactions

* Updating both `wallet-api` and `turtle-service` documentation to mention `feePerByte` parameter, making `fee` parameter optional for `turtle-service`, updating return type to include the fee of the sent transaction.

* ~Optional: Allow fractional fee per byte to make rounder numbers for fees - since currently they are based on the amount of 256 byte chunks, they don't make for very sexy results~ Done

~Note: Currently fee per byte validation is enforced both for sent transactions and pool transactions, instantly. This could potentially be changed, but was done this way both for simpler code, and the suggestion that this may be desired.~ Fake news

By the way - the turtle-service code is kinda shitty. Could probably be simplified, but fuck it, right?

### DIFF
--- a/src/utilities/Utilities.cpp
+++ b/src/utilities/Utilities.cpp
@@ -216,6 +216,80 @@ namespace Utilities
         return true;
     }
 
+    uint64_t getTransactionFee(
+        const size_t transactionSize,
+        const uint64_t height,
+        const uint64_t feePerByte)
+    {
+        const double numChunks = std::ceil(transactionSize / 256.0);
+        const uint64_t multiplier = feePerByte * 256;
+
+        return static_cast<uint64_t>(numChunks) * multiplier;
+    }
+
+    uint64_t getMinimumTransactionFee(
+        const size_t transactionSize,
+        const uint64_t height)
+    {
+        return getTransactionFee(
+            transactionSize,
+            height,
+            CryptoNote::parameters::MINIMUM_FEE_PER_BYTE_V1
+        );
+    }
+
+    size_t estimateTransactionSize(
+        const uint64_t mixin,
+        const size_t numInputs,
+        const size_t numOutputs,
+        const bool havePaymentID,
+        const size_t extraDataSize)
+    {
+        const size_t KEY_IMAGE_SIZE = sizeof(Crypto::KeyImage);
+        const size_t OUTPUT_KEY_SIZE = sizeof(decltype(CryptoNote::KeyOutput::key));
+        const size_t AMOUNT_SIZE = sizeof(uint64_t) + 2; // varint
+        const size_t GLOBAL_INDEXES_VECTOR_SIZE_SIZE = sizeof(uint8_t); // varint
+        const size_t GLOBAL_INDEXES_INITIAL_VALUE_SIZE = sizeof(uint32_t); // varint
+        const size_t GLOBAL_INDEXES_DIFFERENCE_SIZE = sizeof(uint32_t); // varint
+        const size_t SIGNATURE_SIZE = sizeof(Crypto::Signature);
+        const size_t EXTRA_TAG_SIZE = sizeof(uint8_t);
+        const size_t INPUT_TAG_SIZE = sizeof(uint8_t);
+        const size_t OUTPUT_TAG_SIZE = sizeof(uint8_t);
+        const size_t PUBLIC_KEY_SIZE = sizeof(Crypto::PublicKey);
+        const size_t TRANSACTION_VERSION_SIZE = sizeof(uint8_t);
+        const size_t TRANSACTION_UNLOCK_TIME_SIZE = sizeof(uint64_t);
+        const size_t EXTRA_DATA_SIZE = extraDataSize > 0 ? extraDataSize + 4 : 0;
+        const size_t PAYMENT_ID_SIZE = havePaymentID ? 34 : 0;
+
+        /* The size of the transaction preamble */
+        const size_t headerSize = TRANSACTION_VERSION_SIZE
+                                + TRANSACTION_UNLOCK_TIME_SIZE
+                                + EXTRA_TAG_SIZE
+                                + EXTRA_DATA_SIZE
+                                + PUBLIC_KEY_SIZE
+                                + PAYMENT_ID_SIZE;
+        
+        /* The size of each transaction input */
+        const size_t inputSize = INPUT_TAG_SIZE
+                               + AMOUNT_SIZE
+                               + KEY_IMAGE_SIZE
+                               + SIGNATURE_SIZE
+                               + GLOBAL_INDEXES_VECTOR_SIZE_SIZE
+                               + GLOBAL_INDEXES_INITIAL_VALUE_SIZE
+                               + mixin * (GLOBAL_INDEXES_DIFFERENCE_SIZE + SIGNATURE_SIZE);
+
+        const size_t inputsSize = inputSize * numInputs;
+
+        /* The size of each transaction output. */
+        const size_t outputSize = OUTPUT_TAG_SIZE
+                                + OUTPUT_KEY_SIZE
+                                + AMOUNT_SIZE;
+
+        const size_t outputsSize = outputSize * numOutputs;
+
+        return headerSize + inputsSize + outputsSize;
+    }
+
     size_t
         getApproximateMaximumInputCount(const size_t transactionSize, const size_t outputCount, const size_t mixinCount)
     {
--- a/src/utilities/Utilities.cpp
+++ b/src/utilities/Utilities.cpp
@@ -260,7 +260,7 @@ namespace Utilities
         const size_t OUTPUT_TAG_SIZE = sizeof(uint8_t);
         const size_t PUBLIC_KEY_SIZE = sizeof(Crypto::PublicKey);
         const size_t TRANSACTION_VERSION_SIZE = sizeof(uint8_t);
-        const size_t TRANSACTION_UNLOCK_TIME_SIZE = sizeof(uint64_t);
+        const size_t TRANSACTION_UNLOCK_TIME_SIZE = sizeof(uint64_t) + 2; // varint
         const size_t EXTRA_DATA_SIZE = extraDataSize > 0 ? extraDataSize + 4 : 0;
         const size_t PAYMENT_ID_SIZE = havePaymentID ? 34 : 0;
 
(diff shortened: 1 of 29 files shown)


brandonlehmann:
> * Optional: Allow fractional fee per byte to make rounder numbers for fees - since currently they are based on the amount of 256 byte chunks, they don't make for very sexy results

This sounds clunky. Could you please provide further detail on what you mean?

> Note: Currently fee per byte validation is enforced both for sent transactions and pool transactions, instantly. This could potentially be changed, but was done this way both for simpler code, and the suggestion that this may be desired.

I think this would be better set/enabled at the fork height; otherwise, it's difficult for users to clear the fee minimums (and figure them out) if pool operators were to run this code (once merged into development). As it causes changes in both wallets and the node, it has to be a coordinated effort between wallets, the network, and pools.

> By the way - the turtle-service code is kinda shitty. Could probably be simplified, but fuck it, right?

Out of scope. :)

zpalmtree:
> This sounds clunky. Could you please provide further detail on what you mean?

Currently, fee per byte is 2 atomic units per byte. We do the fee in chunks of 256 bytes. So, our fee will step up, 5.12 TRTL, 10.24 TRTL, 15.36 TRTL, 20.48 TRTL. If instead our fee per byte was `1.953125`... then our fees would be 5 TRTL, 10 TRTL, 15 TRTL, 20 TRTL. It's the little things :)

> I think this would be better set/enabled at the fork height; otherwise, it's difficult for users to clear the fee minimums (and figure them out) if pool operators were to run this code (once merged into development). As it causes changes in both wallets and the node, it has to be a coordinated effort between wallets, the network, and pools.

Fair enough.

brandonlehmann:
> > This sounds clunky. Could you please provide further detail on what you mean?
> 
> Currently, fee per byte is 2 atomic units per byte. We do the fee in chunks of 256 bytes. So, our fee will step up, 5.12 TRTL, 10.24 TRTL, 15.36 TRTL, 20.48 TRTL. If instead our fee per byte was `1.953125`... then our fees would be 5 TRTL, 10 TRTL, 15 TRTL, 20 TRTL. It's the little things :)

Is this as easy as changing it to a float and just doing `1.953125`? If so, I support that idea.

zpalmtree:
> Is this as easy as changing it to a float and just doing `1.953125`? If so, I support that idea.

Probably. Lets give it a shot

zpalmtree:
Implemented 
image

zpalmtree:
> Note: Currently fee per byte validation is enforced both for sent transactions and pool transactions, instantly. This could potentially be changed, but was done this way both for simpler code, and the suggestion that this may be desired.

Well, I don't know where I got that idea from, because it's actually only enforced daemon side when we hit the fork height. So, should be good on that front.

brandonlehmann:
> > Note: Currently fee per byte validation is enforced both for sent transactions and pool transactions, instantly. This could potentially be changed, but was done this way both for simpler code, and the suggestion that this may be desired.
> 
> Well, I don't know where I got that idea from, because it's actually only enforced daemon side when we hit the fork height. So, should be good on that front.

Do the wallets auto-switch from 0.10 minimum to calculating at fork height?

zpalmtree:
> > > Note: Currently fee per byte validation is enforced both for sent transactions and pool transactions, instantly. This could potentially be changed, but was done this way both for simpler code, and the suggestion that this may be desired.
> > 
> > 
> > Well, I don't know where I got that idea from, because it's actually only enforced daemon side when we hit the fork height. So, should be good on that front.
> 
> Do the wallets auto-switch from 0.10 minimum to calculating at fork height?

No, the wallets instantly start using fee per byte. Would be easy to implement at a height if desired, but doesn't really effect anything negatively unless you're really counting your turtles

src/utilities/Utilities.cpp
```diff
@@ -216,6 +216,83 @@ namespace Utilities
...
+        const size_t EXTRA_TAG_SIZE = sizeof(uint8_t);
+        const size_t INPUT_TAG_SIZE = sizeof(uint8_t);
+        const size_t OUTPUT_TAG_SIZE = sizeof(uint8_t);
+        const size_t PUBLIC_KEY_SIZE = sizeof(Crypto::PublicKey);
+        const size_t TRANSACTION_VERSION_SIZE = sizeof(uint8_t);
+        const size_t TRANSACTION_UNLOCK_TIME_SIZE = sizeof(uint64_t);
+        const size_t EXTRA_DATA_SIZE = extraDataSize > 0 ? extraDataSize + 4 : 0;
+        const size_t PAYMENT_ID_SIZE = havePaymentID ? 34 : 0;
+
+        /* The size of the transaction preamble */
```
brandonlehmann: This is me being picky, but it's actually a transaction `header` not `preamble`

src/utilities/Utilities.cpp
```diff
@@ -216,6 +216,83 @@ namespace Utilities
...
+        const size_t GLOBAL_INDEXES_VECTOR_SIZE_SIZE = sizeof(uint8_t); // varint
+        const size_t GLOBAL_INDEXES_INITIAL_VALUE_SIZE = sizeof(uint32_t); // varint
+        const size_t GLOBAL_INDEXES_DIFFERENCE_SIZE = sizeof(uint32_t); // varint
+        const size_t SIGNATURE_SIZE = sizeof(Crypto::Signature);
+        const size_t EXTRA_TAG_SIZE = sizeof(uint8_t);
+        const size_t INPUT_TAG_SIZE = sizeof(uint8_t);
+        const size_t OUTPUT_TAG_SIZE = sizeof(uint8_t);
+        const size_t PUBLIC_KEY_SIZE = sizeof(Crypto::PublicKey);
+        const size_t TRANSACTION_VERSION_SIZE = sizeof(uint8_t);
+        const size_t TRANSACTION_UNLOCK_TIME_SIZE = sizeof(uint64_t);
```
brandonlehmann: Since you're defining `AMOUNT_SIZE` as `sizeof(uint64_t) + 2` because it's a varint, this uint64_t should also be +2 as well? It's technically a varint field as well.

src/utilities/Utilities.cpp
```diff
@@ -216,6 +216,83 @@ namespace Utilities
...
+                                + PAYMENT_ID_SIZE;
+        
+        /* The size of each transaction input */
+        const size_t inputSize = INPUT_TAG_SIZE
+                               + AMOUNT_SIZE
+                               + KEY_IMAGE_SIZE
+                               + SIGNATURE_SIZE
+                               + GLOBAL_INDEXES_VECTOR_SIZE_SIZE
+                               + GLOBAL_INDEXES_INITIAL_VALUE_SIZE
+                               + mixin * (GLOBAL_INDEXES_DIFFERENCE_SIZE + SIGNATURE_SIZE);
```
brandonlehmann: `mixin * (GLOBAL_INDEXES_DIFFERENCE_SIZE + SIGNATURE_SIZE)` is this correct?

it should just be `mixin * SIGNATURE_SIZE` -- the payload doesn't have a vector header or any other overhead unless I'm missing something here.

src/walletbackend/Transfer.cpp
```diff
@@ -1150,9 +1449,32 @@ namespace SendTransaction
...
+                ? expectedFee.feePerByte
+                : CryptoNote::parameters::MINIMUM_FEE_PER_BYTE_V1;
+
+            const size_t txSize = toBinaryArray(tx).size();
+
+            const size_t calculatedFee = static_cast<uint64_t>(feePerByte * txSize);
+
+            /* Ensure fee is greater or equal to the fee per byte specified,
+             * and no more than two times the fee per byte specified. */
+            return actualFee >= calculatedFee && actualFee <= calculatedFee * 2;
```
brandonlehmann: Whoa! Where'd this come from? Are you saying that I can't pay MORE than 2x the calculated fee per byte to put my transaction at the top of the pool?! How dare you tell me how I can't spend my funds.

zpalmtree: I think it makes sense to implement a check to ensure the user isn't paying significiantly more than they requested to pay. This is certainly a little fuzzy, so feel free to suggest a better method if you have one.

src/utilities/Utilities.cpp
zpalmtree: I took this straight from some other transaction estimate related code, and didn't verify the numbers. I'll assume you're correct and update.

src/utilities/Utilities.cpp
zpalmtree: OK

src/walletbackend/Transfer.cpp
brandonlehmann: We have use cases (.trtl DNS) which dictates high network transaction fees to pay for services -- while I certainly understand your point, I'm not sure that I can get behind any limit.

src/utilities/Utilities.cpp
brandonlehmann: Open to verification for sure just not sure where the DIFFERENCE_SIZE came from

src/walletbackend/Transfer.cpp
zpalmtree: I'm not sure I follow. You can perfectly specify a large fee. You can either pay a `MinimumFee()`, a `FeePerByte()`, or a `FixedFee()`.

* If you specify a `MinimumFee()`, then we will ensure you pay between the min fee and the min fee x 2, so in this case we will pay between `1.953125` to `3.90625` atomic units per byte.

* If you specify a `FeePerByte()`, then we will ensure you pay between the specified fee per byte and the specified fee per byte x2, so same as `MinimumFee()`, just with a different value potentially.

* If you specify a `FixedFee()`, then we will verify that the fee is identical to the fee you specified. Transaction will be rejected if `< min fee`.

As an end user, you can choose a `MinimumFee()`, by not providing the `fee` or `feePerByte` param. You can specify a `FeePerByte()` by giving a `feePerByte` parameter with the desired fee per byte value. You can specify a `FixedFee()` by using the `fee` param.

brandonlehmann: Ah, gotcha. I misread the logic tree. My bad.

src/utilities/Utilities.cpp
zpalmtree: Fixed

src/utilities/Utilities.cpp
zpalmtree: Fixed

src/utilities/Utilities.cpp
zpalmtree: removed

PR #968, opened 2019-12-15
Rework output capping

zpalmtree:
Update output cap to a fixed `30`. May need some discussion. For example, if paying 5 pool users, with an amount of `12345.67`, that's 7 * 5 outputs = 35, and then some change will of course be needed.

### DIFF
--- a/src/config/CryptoNoteConfig.h
+++ b/src/config/CryptoNoteConfig.h
@@ -209,9 +209,9 @@ namespace CryptoNote
            to help curtail fusion transaction spam. */
         const size_t FUSION_TX_MAX_POOL_COUNT = 20;
 
-        const size_t NORMAL_TX_MAX_OUTPUT_RATIO_V1 = 10;
+        const size_t NORMAL_TX_MAX_OUTPUT_COUNT_V1 = 90;
 
-        const size_t NORMAL_TX_MAX_OUTPUT_RATIO_V1_HEIGHT = 2200000;
+        const size_t NORMAL_TX_MAX_OUTPUT_COUNT_V1_HEIGHT = 2200000;
 
         const uint32_t UPGRADE_HEIGHT_V2 = 1;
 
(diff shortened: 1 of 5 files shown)


brandonlehmann:
I'm not opposed to setting the fixed cap higher.  Coupled with #967 I think we can accept a maximum output count of maybe `50`.

wrkzdev:
I would suggest minimum `60`. Ideally, `90`.

zpalmtree:
> 
> 
> I would suggest minimum `60`. Ideally, `90`.

Can you expand on your thoughts behind these numbers?

wrkzdev:
> > I would suggest minimum `60`. Ideally, `90`.
> 
> Can you expand on your thoughts behind these numbers?

My opinion:

When `fee per byte` activates, it doesn't matter if outputs is `50`, `60`, or `90`. The bigger size tx, the higher fee user needs to spend.

Two tx'es of `30` outputs + another tx of `60` outputs **versus** one tx of `90` outputs (of similar amount from the same wallet) could be just a bit higher from one to another. I think by having less tx numbers in network tx pool can be better for sync'ing daemon also.

On the other hand, if one needs to optimize his wallet, wallet collect up to 94 or 97 inputs of small amount (let's say 0.01 TRTL) into one tx (as I had ever seen in tx pool before). That is to say, one fusion tx can eat all the 0.01 TRTL outputs created by one tx which has `90` outputs already. Unlike before which abuser could execute one tx and has more than 3,600 outputs and needs at least ~ 40 fusion tx'es to collect them into a batch of 0.97 TRTL.

If I were a pool operator, I would like my pool able to send to many miners at once rather than splitting them into several tx'es.

brandonlehmann:
> If I were a pool operator, I would like my pool able to send to many miners at once rather than splitting them into several tx'es.

The pools already split large batches of payments into many different transactions depending on the batch. I'm not sure this is applicable for this reason.


That being said, if we just want to set a higher limit, I'm not against that either as long as there's some logic behind the cap.

zpalmtree:
I can get behind the logic that it doesn't matter really how many outputs you have with fee per byte as long as it's capped at some reasonable value - you're going to be paying your way regardless.

Updated cap to `90`, and resolved conflicts.

brandonlehmann:
giphy

brandonlehmann:
giphy

wrkzdev:
image

@zpalmtree the best.

PR #969, opened 2019-12-17
Remove invalid transactions from pool when included in block

zpalmtree:
To ensure we don't get stuck in a loop where we cannot create blocks because we have a transaction in the pool that was previously valid, but is no longer valid, we should remove invalid transactions from the pool upon failing to add a block containing the tx.


brandonlehmann:
@zpalmtree conflicts :)

zpalmtree:
Conflicts resolved

PR #980, opened 2019-12-30
Fix Daemon `status` command when using threaded RPC

brandonlehmann:
Resolve issue whereby the `status` command in the daemon does not always return a result

This is due to binding to a non-interface IP (0.0.0.0) and the fact that we cannot make HTTP calls to 0.0.0.0 reliably and should specify an IP address to issue the RPC calls against. This commit simply replaces 0.0.0.0 with 127.0.0.1 (localhost) for purposes of routing the client RPC request. In addition, creates the http client in the method scope instead of at the class level as I did encounter a few race conditions whereby the commands handler was somehow initializing before the RpcServer instance was able to return the correct port(s).

### DIFF
--- a/src/daemon/DaemonCommandsHandler.cpp
+++ b/src/daemon/DaemonCommandsHandler.cpp
@@ -54,12 +54,13 @@ DaemonCommandsHandler::DaemonCommandsHandler(
     CryptoNote::Core &core,
     CryptoNote::NodeServer &srv,
     std::shared_ptr<Logging::LoggerManager> log,
-    RpcServer *prpc_server):
+    const std::string ip,
+    const uint32_t port):
     m_core(core),
     m_srv(srv),
     logger(log, "daemon"),
     m_logManager(log),
-    m_rpcServer(std::get<0>(prpc_server->getConnectionInfo()).c_str(), std::get<1>(prpc_server->getConnectionInfo()))
+    m_rpcServer(ip.c_str(), port)
 {
     m_consoleHandler.setHandler(
         "?",
--- a/src/daemon/Daemon.cpp
+++ b/src/daemon/Daemon.cpp
@@ -253,7 +253,7 @@ int main(int argc, char *argv[])
             {
                 oldLogLevel = Logging::DEBUGGING;
                 logColour = Logging::DEFAULT;
-            } 
+            }
             else if (level == Logger::INFO)
             {
                 oldLogLevel = Logging::INFO;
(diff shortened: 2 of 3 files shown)


src/daemon/DaemonCommandsHandler.cpp
```diff
@@ -348,7 +348,24 @@ bool DaemonCommandsHandler::print_pool_sh(const std::vector<std::string> &args)
 //--------------------------------------------------------------------------------
 bool DaemonCommandsHandler::status(const std::vector<std::string> &args)
 {
-    auto res = m_rpcServer.Get("/info");
+    /* Get the RPC IP address we are bound to */
+    auto ip = std::get<0>(m_rpcServer->getConnectionInfo());
```
zpalmtree: would just doing
```cpp
auto [ip, port] = m_rpcServer->getConnectionInfo();
```
not work?

src/daemon/DaemonCommandsHandler.cpp
```diff
@@ -348,7 +348,24 @@ bool DaemonCommandsHandler::print_pool_sh(const std::vector<std::string> &args)
...
+       localhost IP address to reach ourselves */
+    if (ip == "0.0.0.0") {
+        ip = "127.0.0.1";
+    }
+
+    /* Get the RPC port we are bound to */
+    auto port = std::get<1>(m_rpcServer->getConnectionInfo());
+
+    /* Set up our HTTP client */
+    auto l_rpcServer = httplib::Client(ip.c_str(), port);
```
zpalmtree: I understand your reasoning to setup a new client in the method scope from the description, but does this not potentially have a decent amount of overhead if we're making a new client for every `status` call?

brandonlehmann: Do you have a better idea of where to set it up to avoid a race condition? I suppose it could set it on first call and only first call, but I think the overhead is negligible.

src/daemon/DaemonCommandsHandler.cpp
brandonlehmann: Didn't try, didn't think to try. I'll give it a go and report.

src/daemon/DaemonCommandsHandler.cpp
zpalmtree: Perhaps we could just ensure we don't start the commands handler till the RPC is initialized?

e.g. move https://github.com/turtlecoin/turtlecoin/blob/development/src/daemon/Daemon.cpp#L445 to https://github.com/turtlecoin/turtlecoin/blob/development/src/daemon/Daemon.cpp#L464

src/daemon/DaemonCommandsHandler.cpp
brandonlehmann: Handled.

src/daemon/DaemonCommandsHandler.cpp
brandonlehmann: Sure. Handled.

PR #990, opened 2019-12-31
HTTP Client Retry on Socket Failure

brandonlehmann:
Implements a singular retry in `Client::send()` if the request failed, likely due to a socket timeout issue.


brandonlehmann:
Reworked mutex scope handling thanks to @zpalmtree

PR #1012, opened 2020-01-18
Multi-signature Primitives

brandonlehmann:
This PR performs the following updates/changes/additions:

* Moves `EllipticCurvePoint` and `EllipticCurveScalar` into `CryptoTypes.h` where they belong
* Permits all of the types in `CryptoTypes.h` to be initialized using a `std::string` in the constructor thereby making them even easier to work with.
* Adds `sc_mul()` and `sc_mulsub()` to `crypto-ops`
* Separates `generateRingSignatures()` into `prepareRingSignatures()` and `completeRingSignatures()` to allow for creating *partial ring signatures* that can be restored later via the new Multisig methods when multiple parties have different parts of the signature
  * `generateRingSignatures()` calls the `prepareRingSignatures()` followed by `completeRingSignatures()` automatically.
* Ports the multisig primitives over from https://github.com/turtlecoin/turtlecoin-crypto and changes the return values to tuples where necessary.
* Sets the `Crypto::Signature` type to initialize the `data` property to all `0`s for making sure that we can easily recognize incomplete ("prepared") signatures.
* Permits use of the following outside of the *friend* scope
  * `Crypto::scReduce32()`
  * `Crypto::hashToScalar()`
  * `Crypto::derivation_to_scalar`
* Extends `Crypto::derive_public_key()` and `Crypto::derive_secret_key()` to support passing the derivation scalar and thus avoiding re-performing the `derivation_to_scalar()` when not necessary.
* Adds the following test to the `cryptotest` binary:
  * `Crypto::crypto_ops::prepareRingSignatures()`
  * `Crypto::crypto_ops::generateRingSignatures()`
  * `Crypto::Multisig::calculate_multisig_private_keys()`
  * `Crypto::Multisig::calculate_shared_public_key()`
  * `Crypto::Multisig::calculate_shared_private_key()`
  * `Crypto::Multisig::restore_key_image()`
  * `Crypto::Multisig::generate_partial_signing_key()`
  * `Crypto::Multisig::restore_ring_signatures()`

For full working details of **how** the math behind multi-signature primitives work, please see https://hackmd.io/EC2_FT8hRKWBTPvRrEKcFQ (working document, more info to come, and a blog article soon to follow).

Implementing this into wallet(s) is next.


include/CryptoTypes.h
```diff
@@ -241,10 +360,37 @@ namespace Crypto
...
+        /* Initializes the class from a json string */
+        void fromString(const std::string &s)
+        {
+            if (!Common::podFromHex(s, data))
+            {
+                throw std::invalid_argument("Error parsing JSON keyimage, wrong length or not hex");
+            }
+        }
+
+        uint8_t data[64] = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
zpalmtree: I think you could rewrite this as `uint8_t data[64] = {}` or `uint8_t data[64] = {0};`

brandonlehmann: Updated to `{}` which indeed works fine.

PR #1029, opened 2020-02-06
Add optimize target parameter

zpalmtree:
Adds an `optimizeTarget` parameter to `/transactions/send/fusion/advanced`. Amounts larger than this target will not be fused. Also makes the destination parameter optional.

### DIFF
--- a/src/errors/Errors.cpp
+++ b/src/errors/Errors.cpp
@@ -316,6 +316,11 @@ std::string Error::getErrorMessage() const
                    "Please re-prepare and re-send the transaction, ensuring you "
                    "specify the correct transaction hash.";
         }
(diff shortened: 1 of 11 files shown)


rashedmyt:
Should this be a WIP PR, since the functionality isn't enabled

brandonlehmann:
No, it’s enabled. It’s an alternate code path that is not enabled that would fail current network transaction validation rules.  Check the discussion in #dev_general for details but there are ups and downs to disabling that network check to permit operation in that manner.

brandonlehmann:
For more clarity: what is enabled here is what is currently available via Wallet Green, which works. What is not enabled is an alternative.

rashedmyt:
Ah okay..

Issue #1030, opened 2020-02-08
/submitblock should not be allowed when daemon is not synced


zpalmtree:
Previously, this would give a 'Core is busy' error. Currently, we allow submitblock to continue, which may cause issues.

zpalmtree:
We should also apply this to /getblocktemplate

PR #1031, opened 2020-02-08
Prevent calling certain endpoints when unsynced

zpalmtree:
Resolves: #1030

### DIFF
--- a/src/rpc/RpcServer.cpp
+++ b/src/rpc/RpcServer.cpp
@@ -144,7 +144,7 @@ RpcServer::RpcServer(
             .Get("/peers", router(&RpcServer::peers, RpcMode::Default, bodyNotRequired, syncNotRequired))
 
             .Post("/json_rpc", jsonRpc)
-            .Post("/sendrawtransaction", router(&RpcServer::sendTransaction, RpcMode::Default, bodyRequired, syncNotRequired))
+            .Post("/sendrawtransaction", router(&RpcServer::sendTransaction, RpcMode::Default, bodyRequired, syncRequired))
             .Post("/getrandom_outs", router(&RpcServer::getRandomOuts, RpcMode::Default, bodyRequired, syncNotRequired))
             .Post("/getwalletsyncdata", router(&RpcServer::getWalletSyncData, RpcMode::Default, bodyRequired, syncNotRequired))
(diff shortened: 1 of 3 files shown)


src/rpc/RpcServer.cpp
```diff
@@ -134,22 +138,22 @@ RpcServer::RpcServer(
...
-            .Post("/getrandom_outs", router(&RpcServer::getRandomOuts, RpcMode::Default, bodyRequired))
-            .Post("/getwalletsyncdata", router(&RpcServer::getWalletSyncData, RpcMode::Default, bodyRequired))
-            .Post("/get_global_indexes_for_range", router(&RpcServer::getGlobalIndexes, RpcMode::Default, bodyRequired))
-            .Post("/queryblockslite", router(&RpcServer::queryBlocksLite, RpcMode::Default, bodyRequired))
-            .Post("/get_transactions_status", router(&RpcServer::getTransactionsStatus, RpcMode::Default, bodyRequired))
-            .Post("/get_pool_changes_lite", router(&RpcServer::getPoolChanges, RpcMode::Default, bodyRequired))
-            .Post("/queryblocksdetailed", router(&RpcServer::queryBlocksDetailed, RpcMode::AllMethodsEnabled, bodyRequired))
-            .Post("/get_o_indexes", router(&RpcServer::getGlobalIndexesDeprecated, RpcMode::Default, bodyRequired))
-            .Post("/getrawblocks", router(&RpcServer::getRawBlocks, RpcMode::Default, bodyRequired))
+            .Post("/sendrawtransaction", router(&RpcServer::sendTransaction, RpcMode::Default, bodyRequired, syncNotRequired))
```
brandonlehmann: I would argue that sync **is** required for sending a transaction via the node otherwise it *may* not properly validate the transaction inputs, return a response to the wallet saying it sent, only to have other nodes invalidate it.

LeoStehlik: true... seen that happening actually

LeoStehlik: other than that, this fix a must for pool operators, otherwise it can get very nasty ... (you probably sense me speaking from experience...)

zpalmtree: Updated

Issue #1022, opened 2020-01-27
RocksDB thread safety in validateTransactionInputsExpensive


zpalmtree:
There appears to be an issue where multi-threading in [`validateTransactionInputsExpensive`](https://github.com/turtlecoin/turtlecoin/blob/26b6eac05372c2afa38cfeeacc310590ea5d0426/src/cryptonotecore/ValidateTransaction.cpp#L477) causes crashes. I have seen a number of crash logs similar to the following:

```
#0  0x00007f3ff0670f2e in malloc () from /usr/lib/libc.so.6
#1  0x000056291bc3f9da in operator new(unsigned long) ()
#2  0x000056291b848602 in rocksdb::IterKey::EnlargeBuffer(unsigned long) ()
#3  0x000056291b874889 in rocksdb::IterKey::TrimAppend(unsigned long, char const*, unsigned long) ()
#4  0x000056291b95e1b8 in bool rocksdb::DataBlockIter::ParseNextDataKey<rocksdb::DecodeEntry>(char const*) ()
#5  0x000056291b95d065 in rocksdb::DataBlockIter::Seek(rocksdb::Slice const&) ()
#6  0x000056291b8db942 in rocksdb::BlockBasedTable::Get(rocksdb::ReadOptions const&, rocksdb::Slice const&, rocksdb::GetContext*, rocksdb::SliceTransform const*, bool) ()
#7  0x000056291b87361e in rocksdb::TableCache::Get(rocksdb::ReadOptions const&, rocksdb::InternalKeyComparator const&, rocksdb::FileMetaData const&, rocksdb::Slice const&, rocksdb::GetContext*, rocksdb::SliceTransform const*, rocksdb::HistogramImpl*, bool, int) ()
#8  0x000056291b87bfe2 in rocksdb::Version::Get(rocksdb::ReadOptions const&, rocksdb::LookupKey const&, rocksdb::PinnableSlice*, rocksdb::Status*, rocksdb::MergeContext*, unsigned long*, bool*, bool*, unsigned long*, rocksdb::ReadCallback*, bool*, bool) ()
[...]

PR #1032, opened 2020-02-08
Use a thread safe DB access method for threaded input is spent check

zpalmtree:
Possible fix to #1022 

Not very tested. Idea is that `Get()` is thread safe whilst `MultiGet` is assumed to not be, so if we just use repeated `Get()`'s, it should be fine to use.

### DIFF
--- a/include/IDataBase.h
+++ b/include/IDataBase.h
@@ -21,5 +21,7 @@ namespace CryptoNote
         virtual std::error_code write(IWriteBatch &batch) = 0;
 
         virtual std::error_code read(IReadBatch &batch) = 0;
+
+        virtual std::error_code readThreadSafe(IReadBatch &batch) = 0;
     };
 } // namespace CryptoNote
--- a/src/cryptonotecore/DatabaseBlockchainCache.cpp
+++ b/src/cryptonotecore/DatabaseBlockchainCache.cpp
@@ -1188,7 +1188,7 @@ namespace CryptoNote
     bool DatabaseBlockchainCache::checkIfSpent(const Crypto::KeyImage &keyImage, uint32_t blockIndex) const
     {
         auto batch = BlockchainReadBatch().requestBlockIndexBySpentKeyImage(keyImage);
-        auto res = database.read(batch);
+        auto res = database.readThreadSafe(batch);
         if (res)
         {
             logger(Logging::ERROR) << "checkIfSpent failed, request to database failed: " << res.message();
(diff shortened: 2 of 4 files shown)


brandonlehmann:
This may have a performance impact without understanding how RocksDB is performing the `MultiGet` call.

There are distinct differences in how DB systems *typically* handle a "multiget" in that it may look like (pseudo)

1) `SELECT * FROM tbl WHERE key = ? OR key = ??` which is typically faster than performing a union on via two calls

2) `SELECT * FROM tbl WHERE key = ? UNION SELECT * FROM tbl WHERE key = ??` which is typically faster than

3) `SELECT * FROM tbl WHERE key = ?` followed by `SELECT * FROM tbl WHERE key = ??` and then joined together in the package that asked for the data in the first place

Essentially, you've turned this into #3, which is, in most DB systems, the slowest possible path.

zpalmtree:
> This may have a performance impact without understanding how RocksDB is performing the `MultiGet` call.
> 
> There are distinct differences in how DB systems _typically_ handle a "multiget" in that it may look like (pseudo)
> 
>     1. `SELECT * FROM tbl WHERE key = ? OR key = ??` which is typically faster than performing a union on via two calls
> 
>     2. `SELECT * FROM tbl WHERE key = ? UNION SELECT * FROM tbl WHERE key = ??` which is typically faster than
> 
>     3. `SELECT * FROM tbl WHERE key = ?` followed by `SELECT * FROM tbl WHERE key = ??` and then joined together in the package that asked for the data in the first place
> 
> 
> Essentially, you've turned this into #3, which is, in most DB systems, the slowest possible path.

In this case I think it will have minimal or no performance impact. This alteration is only applied to the `checkIfSpent` function, which takes a single key, `keyImage`, rather than multiple keys. So we should just be doing a single `.Get()`.

brandonlehmann:
> > This may have a performance impact without understanding how RocksDB is performing the `MultiGet` call.
> > There are distinct differences in how DB systems _typically_ handle a "multiget" in that it may look like (pseudo)
> > ```
> > 1. `SELECT * FROM tbl WHERE key = ? OR key = ??` which is typically faster than performing a union on via two calls
> > 
> > 2. `SELECT * FROM tbl WHERE key = ? UNION SELECT * FROM tbl WHERE key = ??` which is typically faster than
> > 
> > 3. `SELECT * FROM tbl WHERE key = ?` followed by `SELECT * FROM tbl WHERE key = ??` and then joined together in the package that asked for the data in the first place
> > ```
> > 
> > 
> > Essentially, you've turned this into #3, which is, in most DB systems, the slowest possible path.
> 
> In this case I think it will have minimal or no performance impact. This alteration is only applied to the `checkIfSpent` function, which takes a single key, `keyImage`, rather than multiple keys. So we should just be doing a single `.Get()`.

Gotcha. Thanks for the clarity.

PR #1037, opened 2020-02-28
Better way to ensure daemon is fully synced

zpalmtree:
Sometimes, the p2p will report we are synced (due to syncing with someone elses unsynced node). If we check both the height and the p2p sync status, we should get a more accurate result.

Note: Not tested, daemon is syncing.


rashedmyt:
LGTM from @RocksteadyTC

PR #1056, opened 2020-04-29
WIP: Issue#1025 updating error returns with specific error numbers

JeremyDoubleday:


src/walletbackend/WalletBackend.cpp
```diff
@@ -187,15 +187,15 @@ std::tuple<Error, std::shared_ptr<WalletBackend>> WalletBackend::importWalletFro
...
+        return {Error(INVALID_WALLET_FILENAME, "The wallet filename is not valid. Error: " + std::string(strerror(errno))), nullptr};
     }
 
     /* Convert the mnemonic into a private spend key */
     auto [mnemonicError, privateSpendKey] = Mnemonics::MnemonicToPrivateKey(mnemonicSeed);
 
     if (mnemonicError)
     {
-        return {mnemonicError, nullptr};
+        return {Error(INVALID_MNEMONIC, std::string(strerror(errno))), nullptr};
```
zpalmtree, line 198: We should only be using `errno` where we're getting back an error from the operating system, which is mainly from calls when we're interacting with the file system. I'm not 100% sure if there's any documentation listing which ones these are, but you can easily verify that it is being set by providing invalid arguments, and then checking if you get an appropriate error.

E.g.

```cpp
if (!std::ifstream("non-existent-file.txt"))
{
    std::cout << "Failed to open file: " << strerror(errono) << std::endl;
}
```

This line and a few of the others will never set `errno` as they don't interact with the operating system.

src/walletbackend/WalletBackend.cpp
```diff
@@ -187,15 +187,15 @@ std::tuple<Error, std::shared_ptr<WalletBackend>> WalletBackend::importWalletFro
     /* Check the filename is valid */
     if (Error error = checkNewWalletFilename(filename); error != SUCCESS)
     {
-        return {error, nullptr};
+        return {Error(INVALID_WALLET_FILENAME, "The wallet filename is not valid. Error: " + std::string(strerror(errno))), nullptr};
```
zpalmtree, line 190: I think it would make a lot more sense to update `checkNewWalletFilename` to include the `errno` string, rather than updating at each usage of the function. This will ensure we are not accidentally returning `INVALID_WALLET_FILENAME` when we should be returning `WALLET_FILE_ALREADY_EXISTS`, as can currently occur with your changes.

brandonlehmann:
@JeremyDoubleday Following up to see if you planned to circle back on this one.

JeremyDoubleday:
> @brandonlehmann Probably not anytime soon honestly, unfortunately been working on other things for school instead.

brandonlehmann:
> > @brandonlehmann Probably not anytime soon honestly, unfortunately been working on other things for school instead.

Not a problem. Just wanted to make sure in case someone else wanted to step in and help. No use doubling efforts :)

brandonlehmann:
Manually merged into 1025 branch so that work can be completed.

PR #1072, opened 2020-06-20
Fix clang 10 build issues and warnings, including rocksDB upgrade

zpalmtree:


brandonlehmann:
What version of RocksDB is this (as I think we should include in the commit message on squash)?

zpalmtree:
> 
> 
> What version of RocksDB is this (as I think we should include in the commit message on squash)?

6.10.2 https://github.com/facebook/rocksdb/releases/tag/v6.10.2

zpalmtree:
Commit history is clean now just need build to pass

PR #1073, opened 2020-06-21
Require coinbase transactions to prove the destination address

zpalmtree:


brandonlehmann:
Targeting 1.0.0 for this if that’s okay :)

PR #1076, opened 2020-06-25
Remove nlohmann for rapidjson & daemon configuration reading enhancement

### LINKED ISSUE #714: Replace all usage of nlohmann::json with rapidjson

### LINKED ISSUE #1058: Raise a warning if the daemon can't parse a config file
I run my daemon with a config file to avoid a United Nations Headquarters situation (way too many flags), but in editing said file I made a typo that prevented it from being parsed.  As it stands, that process fails silently and the daemon launches with default settings.  I think a user should be notified if a config file can't be parsed (I only noticed because I customize the data directory, which is information printed on launch).  Something like this:
```
WARNING: There was a problem with the configuration file provided at '/path/to/config.json' - launching daemon with default settings.
```
Optionally, more detailed information could be given.  Different warnings for a nonexistent file or one that simply can't be parsed would be nice, and potentially, depending on difficulty of implementation, pointing to a specific line of the config that caused the parsing error.

brandonlehmann:
Resolves #714 
Resolves #1058 

Light testing performed.


RocksteadyTC:
YESSSSSS

include/WalletTypes.h
```diff
@@ -19,9 +20,36 @@ namespace WalletTypes
     struct KeyOutput
     {
         Crypto::PublicKey key;
+
         uint64_t amount;
-        /* Daemon doesn't supply this, blockchain cache api does. */
+
+        // TODO: Get rid of this
         std::optional<uint64_t> globalOutputIndex;
```
zpalmtree, line 27: I'm not sure if this data structure ends up getting written to file with the `globalOutputIndex` member present, or if it's just used to decode an endpoint. But if it does, and we only conditionally read this variable, then loading and saving a wallet would cause the data to be lost.

src/walletapi/ApiDispatcher.cpp
```diff
@@ -1232,41 +1525,68 @@ std::tuple<Error, uint16_t>
...
+    ApiDispatcher::getTransactions(const httplib::Request &req, httplib::Response &res, const rapidjson::Document &body) const
 {
-    nlohmann::json j {{"transactions", m_walletBackend->getTransactions()}};
+    rapidjson::StringBuffer sb;
 
-    publicKeysToAddresses(j);
+    rapidjson::Writer<rapidjson::StringBuffer> writer(sb);
 
-    res.set_content(j.dump(4) + "\n", "application/json");
+    writer.StartArray();
```
zpalmtree, line 1551: It looks like here and in a few other places we're writing the array as the top level of the JSON, but in the previous version, it has a `transactions` key with the array as the value. Is that intentional?

include/WalletTypes.h
brandonlehmann: I didn’t see it saved, but I need to go through it again and check.

src/walletapi/ApiDispatcher.cpp
brandonlehmann: Unintentionally lost during refactoring. Will correct.

brandonlehmann: Handled.

include/WalletTypes.h
brandonlehmann: Appears to be back filling it elsewhere for now. Will leave it be.

RocksteadyTC:
FLY MY LITTLE PRETTIES FLYYYYYYYY

RocksteadyTC:
fly

PR #1077, opened 2020-06-25
Add Karai Pointer and Hash tag to TX_EXTRA

brandonlehmann:
Depends on #1076  do not merge before that one.


brandonlehmann:
May make an alteration. Will comment when good.

brandonlehmann:
Going to revamp this.

PR #1082, opened 2020-07-23
Update gcc build instructions to match the matrix.yml

ExtraHash:
I have tested these instructions on a fresh VM with ubuntu latest LTS, and they work fine. These seem to be the instructions the CI tests use. The cmake that ships in the default ubuntu repositories is 3.16.3 right now, so needing to install CMake from pip is no longer an issue. I'm thinking the clang instructions could probably be cleaned up significantly as well as they also use the same exact build instructions in the matrix, with the exception of the cc and ccx variables and compilers installed obviously, but I don't use clang so I didn't want to mess with it.

### DIFF
--- a/COMPILE.md
+++ b/COMPILE.md
@@ -38,11 +38,8 @@ However, if you wish to create _portable_ binaries that can be shared between sy
 
 ##### Ubuntu with GCC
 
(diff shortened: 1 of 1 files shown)


brandonlehmann:
Changing base branch

Issue #893, opened 2019-10-02
Incorrectly resetting instead of rewinding on subwallet import


zpalmtree:
This line <https://github.com/turtlecoin/turtlecoin/blob/103b8efece92212e795f50477d6c51029dbdd5e0/src/walletbackend/WalletBackend.cpp#L848> and [maybe] the one above should be a `rewind`, not a `reset`. If, for example, we had a subwallet created at a height of 100,000, our wallet is synced to 500,000, and we import a subwallet with a scan height of 400,000, we will lose any data from 100,000 to 400,000, due to the reset.

We should instead rewind to the height of 400,000, removing any data after 400,000, but retaining the data before.

I don't think I have implemented rewind in the C++ backend yet, but you can port it from the js backend here - <https://github.com/turtlecoin/turtlecoin-wallet-backend-js/blob/f367a8e6d66afe5fa285d6acbd8da3a2d8593b7a/lib/WalletBackend.ts#L955>

ExtraHash:
Rewinding seems to still be very buggy even after we changed the method of balance calculation; the balance would be incorrect for some reason until a full rescan was performed, particularly with deep rewinds. Is there no way to only scan for the one extra viewkey rather than rewinding all of them?

brandonlehmann:
> Is there no way to only scan for the one extra viewkey rather than rewinding all of them?

I believe you mean scan for one extra *spend key* as the *view key* doesn't change in subwallets.

zpalmtree:
No, this would greatly complicate the code. This is also a completely unrelated repo

ExtraHash:
Correct, I did mean that. Also, it's not exactly unrelated if rewind is going to be ported out of the js backend

brandonlehmann:
Resolved in #1088

PR #1085, opened 2020-07-30
added rewind to height option

mrrovot:
I think this should work

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,7 +207,6 @@ if (NOT MSVC)
             set(CMAKE_CXX_COMPILER "/usr/local/opt/llvm@8/bin/clang++")
             set(CMAKE_EXE_LINKER_FLAGS "-L/usr/local/opt/llvm@8/lib -Wl,-rpath,/usr/local/opt/llvm@8/lib")
             set(CMAKE_SHARED_LINKER_FLAGS "-L/usr/local/opt/llvm@8/lib -Wl,-rpath,/usr/local/opt/llvm@8/lib")
-            #set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++ -I/usr/local/opt/llvm/include/c++/v1")
         endif ()
     endif ()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,7 +202,12 @@ if (NOT MSVC)
 
         elseif (CMAKE_C_COMPILER_ID STREQUAL "Clang")
             # Need to link against the llvm libc++ library, default is too old for std::filesystem
-            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++ -I/usr/local/opt/llvm/include/c++/v1")
+
+            set(CMAKE_C_COMPILER "/usr/local/opt/llvm@8/bin/clang")
+            set(CMAKE_CXX_COMPILER "/usr/local/opt/llvm@8/bin/clang++")
+            set(CMAKE_EXE_LINKER_FLAGS "-L/usr/local/opt/llvm@8/lib -Wl,-rpath,/usr/local/opt/llvm@8/lib")
+            set(CMAKE_SHARED_LINKER_FLAGS "-L/usr/local/opt/llvm@8/lib -Wl,-rpath,/usr/local/opt/llvm@8/lib")
+            #set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++ -I/usr/local/opt/llvm/include/c++/v1")
         endif ()
     endif ()
 
--- a/src/walletbackend/WalletSynchronizer.cpp
+++ b/src/walletbackend/WalletSynchronizer.cpp
@@ -740,6 +740,12 @@ void WalletSynchronizer::reset(uint64_t startHeight)
 void WalletSynchronizer::rewind(uint64_t startHeight)
 {
 
+    if (startHeight < m_startHeight)
+    {
+        m_startHeight = startHeight;
+        m_startTimestamp = 0;
+    }
+    
     /* Discard downloaded blocks and sync status */
     m_blockDownloader = BlockDownloader(m_daemon, m_subWallets, m_startHeight, m_startTimestamp);
 
--- a/src/walletbackend/WalletSynchronizer.cpp
+++ b/src/walletbackend/WalletSynchronizer.cpp
@@ -694,7 +694,7 @@ void WalletSynchronizer::stop()
     Logger::logger.log("Stopping sync process", Logger::DEBUG, {Logger::SYNC});
 
     /* Tell the threads to stop */
-    m_shouldStop = true; 
+    m_shouldStop = true;
 
     /* Tell the block downloader to stop and wait for it */
     m_blockDownloader.stop();
@@ -737,14 +737,14 @@ void WalletSynchronizer::reset(uint64_t startHeight)
 }
  
 
-void WalletSynchronizer::rewind(uint64_t rewindToHeight)
+void WalletSynchronizer::rewind(uint64_t startHeight)
 {
     /* Reset start height / timestamp */
-    m_rewindToHeight = rewindToHeight;
+    m_startHeight = startHeight;
     m_startTimestamp = 0;
 
     /* Discard downloaded blocks and sync status */
-    m_blockDownloader = BlockDownloader(m_daemon, m_subWallets, m_rewindToHeight, m_startTimestamp);
+    m_blockDownloader = BlockDownloader(m_daemon, m_subWallets, m_startHeight, m_startTimestamp);
 
     /* Need to call start in your calling code - We don't call it here so
        you can schedule the start correctly */
--- a/src/walletbackend/WalletSynchronizer.cpp
+++ b/src/walletbackend/WalletSynchronizer.cpp
@@ -694,7 +694,7 @@ void WalletSynchronizer::stop()
     Logger::logger.log("Stopping sync process", Logger::DEBUG, {Logger::SYNC});
 
     /* Tell the threads to stop */
-    m_shouldStop = true;
+    m_shouldStop = true; 
 
     /* Tell the block downloader to stop and wait for it */
     m_blockDownloader.stop();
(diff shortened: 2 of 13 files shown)


mrrovot:
will compile it tomorrow on my mac to fix the errors

src/walletbackend/WalletSynchronizer.cpp
```diff
@@ -694,7 +694,7 @@ void WalletSynchronizer::stop()
     Logger::logger.log("Stopping sync process", Logger::DEBUG, {Logger::SYNC});
 
     /* Tell the threads to stop */
-    m_shouldStop = true;
+    m_shouldStop = true;
```
brandonlehmann, line 697: Errant space :)

mrrovot:
Works on my machine

mrrovot:
I think this fixes this #893

brandonlehmann:
Set to target v1.0.0

src/walletbackend/WalletSynchronizer.cpp
```diff
@@ -735,6 +735,20 @@ void WalletSynchronizer::reset(uint64_t startHeight)
     /* Need to call start in your calling code - We don't call it here so
        you can schedule the start correctly */
 }
+ 
+
+void WalletSynchronizer::rewind(uint64_t startHeight)
+{
+    /* Reset start height / timestamp */
+    m_startHeight = startHeight;
+    m_startTimestamp = 0;
```
zpalmtree, line 744: I don't think we should change these two values. This is meant to be the creation height/timestamp of the wallet - this allows us to potentially do a 'reset' if desired - but from when the wallet was created, rather than 0, to avoid wasted sync time. Since we're just doing a rewind, we want to retain the original creation height/timestamp.

src/zedwallet++/CommandImplementations.cpp
```diff
@@ -295,6 +295,37 @@ void reset(const std::shared_ptr<WalletBackend> walletBackend)
     walletBackend->m_eventHandler->onTransaction.resume();
 }
 
+void rewind(const std::shared_ptr<WalletBackend> walletBackend)
+{
+    const uint64_t scanHeight = ZedUtilities::getScanHeight();
```
zpalmtree, line 300: The text that this helper function gives out doesn't really match what the `rewind` function is doing - It should probably be something like "What block height do you want to rewind your wallet to? All blocks after this height will be rescanned, if you suspect a transaction has been missed by the sync process".

We could either copy the function and alter it slightly, take a parameter for the prompt text, or use a bool parameter to flip output for rewind/reset, whatever works.

mrrovot: working on it

mrrovot: Added a `getRewindToHeight()` utility because `getScanHeight` was used in many other files so didn't want to mess it up

src/walletbackend/WalletSynchronizer.cpp
mrrovot: I am not sure I understand. Could I do this instead ? 
```   
 m_startTimestamp = getUint64FromJSON(j, "startTimestamp");
 m_startHeight = getUint64FromJSON(j, "startHeight");

```

zpalmtree: Just remove these two lines, the initialization of the blockdownloader below should be enough to reset the sync process for this class.

src/zedwallet++/Utilities.cpp
```diff
@@ -48,7 +48,53 @@ namespace ZedUtilities
...
+                      << InformationMsg("rewind your wallet to?") << "\n\n"
+                      << "All blocks after this height will be rescanned, "
+                      << "use this command if you suspect a transaction "
+                      << "has been missed by the sync process."
+                      << "\n\n"
+                      << "If you do not know the exact height, "
+                      << "err on the side of caution so transactions do not "
+                      << "get missed."
+                      << "\n\n"
+                      << InformationMsg("Hit enter for the sub-optimal default ") << InformationMsg("of zero: ");
```
zpalmtree, line 96: I think it might be best to make the default something like `current wallet sync height - 1000`. A rewind is usually suggested when a quite recent transaction has been missed, and it's rarely desired to do a deep resync. 

How about this:

"What block height do you want to rewind your wallet to?
All blocks after this height will be rescanned, use the command if you suspect a transaction has been missed by the sync process.
Hit enter for the default of `<current wallet sync height - 1000>` (1000 blocks ago)."

You'll probably need to add a new parameter to the function passing in the wallet sync height.

CMakeLists.txt
```diff
@@ -202,7 +202,12 @@ if (NOT MSVC)
 
         elseif (CMAKE_C_COMPILER_ID STREQUAL "Clang")
             # Need to link against the llvm libc++ library, default is too old for std::filesystem
-            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++ -I/usr/local/opt/llvm/include/c++/v1")
+
+            set(CMAKE_C_COMPILER "/usr/local/opt/llvm@8/bin/clang")
+            set(CMAKE_CXX_COMPILER "/usr/local/opt/llvm@8/bin/clang++")
+            set(CMAKE_EXE_LINKER_FLAGS "-L/usr/local/opt/llvm@8/lib -Wl,-rpath,/usr/local/opt/llvm@8/lib")
+            set(CMAKE_SHARED_LINKER_FLAGS "-L/usr/local/opt/llvm@8/lib -Wl,-rpath,/usr/local/opt/llvm@8/lib")
+            #set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++ -I/usr/local/opt/llvm/include/c++/v1")
```
zpalmtree, line 210: If these changes indeed fix the apple build, go ahead and remove the commented out line.

src/zedwallet++/Utilities.cpp
mrrovot: Z is there a difference if I get the walletBlockCount from `walletBackend->getStatus()` or `walletBackend->getSyncStatus` .. so far I got it working with `getStatus()` but not sure if there is a preference/difference

zpalmtree: No, they both return the same value for `walletBlockCount`, which is the value you want to be using, so either is fine.

mrrovot:
I think the rewind is still not working properly, I tried to rewind to block 2400000 or any other random block but always starts at ~2510000. know what can cause this?

zpalmtree:
> 
> 
> I think the rewind is still not working properly, I tried to rewind to block 2400000 or any other random block but always starts at ~2510000. know what can cause this?
> maybe the new scanHeight is not being saved to the walletBackend somewhere

Is that the height the wallet was created at?

mrrovot:
Yes I believe so

zpalmtree:
OK, I think we might need to re-add these lines, but conditionally: https://github.com/turtlecoin/turtlecoin/pull/1085/commits/4762e7497b884010a0bb9b4e5dd917281dd8d8d7


```
/* Reset start height / timestamp */
if (startHeight < m_startHeight)
{
    m_startHeight = startHeight;
    m_startTimestamp = 0;
}
```

If you rewind to a height before the wallet was created, then we need to update the start height.

mrrovot:
> OK, I think we might need to re-add these lines, but conditionally: [4762e74](https://github.com/turtlecoin/turtlecoin/commit/4762e7497b884010a0bb9b4e5dd917281dd8d8d7)
> 
> ```
> /* Reset start height / timestamp */
> if (startHeight < m_startHeight)
> {
>     m_startHeight = startHeight;
>     m_startTimestamp = 0;
> }
> ```
> 
> If you rewind to a height before the wallet was created, then we need to update the start height.

Yep, that made it work!

mrrovot:
On another note: maybe some commands could be added as flags when starting the wallet, for example: `--reset height`, but dunno just thinking..

cause if I want to reset it feels hacky to `connect without a node > hit continue > reset > reconnect with a node`

src/zedwallet++/Utilities.cpp
```diff
@@ -77,5 +77,60 @@ namespace ZedUtilities
...
+
+            std::cout << InformationMsg("What block height do you want to ")
+                      << InformationMsg("rewind your wallet to?") << "\n\n"
+                      << "All blocks after this height will be rescanned, "
+                      << "use this command if you suspect a transaction "
+                      << "has been missed by the sync process."
+                      << "\n\n"
+                      << "If you do not know the exact height, "
+                      << "err on the side of caution so transactions do not "
+                      << "get missed."
```
zpalmtree, line 100: Lets get rid of this text since you don't need to err on the side of caution when rewinding - ideally nothing has been missed.

```
                      << "\n\n"
                      << "If you do not know the exact height, "
                      << "err on the side of caution so transactions do not "
                      << "get missed."
```

src/zedwallet++/Utilities.cpp
```diff
@@ -77,5 +77,60 @@ namespace ZedUtilities
...
+                      << "\n\n"
+                      << InformationMsg("Hit enter for the default of <")
```
zpalmtree, line 102: I was unclear, I meant to use the `<>` to denote a variable, not to include it in the output.

src/zedwallet++/Utilities.cpp
```diff
@@ -77,5 +77,60 @@ namespace ZedUtilities
...
+                      << InformationMsg(status.walletBlockCount)
```
zpalmtree, line 103: This should be `status.walletBlockCount - 1000` here - but that won't work if the height is less than `1000` currently, and the same issue is present where you use it below.

Lets define a variable above, `const uint64_t defaultRewindHeight = status.walletBlockCount < 1000 ? 0 : status.walletBlockCount - 1000`, just after `status` is defined, then we can use `defaultRewindHeight` here, and below, where we `return status.walletBlockCount - 1000`

zpalmtree:
> 
> 
> On another note: maybe some commands could be added as flags when starting the wallet, for example: `--reset height`, but dunno just thinking..
> 
> cause if I want to reset it feels hacky to `connect without a node > hit continue > reset > reconnect with a node`
> 
> I guess it will work only if passing the wallet and password as flags with the reset

Absolutely, that would be a nice feature to add. I would suggest adding it as separate PR from this one, if you're interested in implementing it.

Two other things which would be easy to implement once this PR is merged also
* Rewinding wallet-api endpoint
* https://github.com/turtlecoin/turtlecoin/issues/893

mrrovot:
> Absolutely, that would be a nice feature to add. I would suggest adding it as separate PR from this one, if you're interested in implementing it.
> 
> Two other things which would be easy to implement once this PR is merged also
> 
> * Rewinding wallet-api endpoint
> * #893

NP, will try to do those after this one is done

mrrovot:
> > Two other things which would be easy to implement once this PR is merged also
> > 
> > * Rewinding wallet-api endpoint
> > * #893

fixed #893 
where do I find the "Rewinding wallet-api endpoint" issue?

zpalmtree:
> 
> 
> > > Two other things which would be easy to implement once this PR is merged also
> > > 
> > > * Rewinding wallet-api endpoint
> > > * #893
> 
> fixed #893
> where do I find the "Rewinding wallet-api endpoint" issue?

There's no issue created for it yet. Feel free to create one, or just make the PR.

mrrovot:
> There's no issue created for it yet. Feel free to create one, or just make the PR.

Opened PR #1089

PR #1088, opened 2020-08-01
Fix 893

### LINKED ISSUE #893: Incorrectly resetting instead of rewinding on subwallet import

mrrovot:
Resolves #893

### DIFF
--- a/src/walletbackend/WalletBackend.cpp
+++ b/src/walletbackend/WalletBackend.cpp
@@ -902,7 +902,7 @@ std::tuple<Error, std::string>
                 m_walletSynchronizer->reset(scanHeight);
 
                 /* Reset transactions, inputs, etc */
-                m_subWallets->reset(scanHeight);
+                m_subWallets->rewind(scanHeight);
             }
         }
 


brandonlehmann:
Rebased here as well and build is re-triggered. Should be better now :)

PR #1091, opened 2020-08-02
Prevent importing same wallet multiple times

### LINKED ISSUE #1024: Prevent importing same wallet multiple times
Currently, I believe we do not prevent a user from importing the same wallet multiple times, via WalletBackend or wallet-api. We should check if the wallet already exists and kick back an error if so. I would suggest doing this in the WalletBackend or SubWallets code paths.

brandonlehmann:
Resolves #1024


RocksteadyTC:
leroy jenkins?

brandonlehmann:
> leroy jenkins?

wait for the code owner :)

Issue #862, opened 2019-08-07
Use matches property in ApiDispatcher regex


zpalmtree:
Some calls in the [ApiDispatcher](https://github.com/turtlecoin/turtlecoin/blob/development/src/walletapi/ApiDispatcher.cpp) use a regex, for example, `getTransactionDetails`. They then extract the query parameters like so:

```cpp
std::tuple<Error, uint16_t> ApiDispatcher::getTransactionDetails(
    const httplib::Request &req,
    httplib::Response &res,
    const nlohmann::json &body) const
{
    std::string hashStr = req.path.substr(std::string("/transactions/hash/").size());
```

We could instead extract `hashStr` using the `matches` property on the `req` object, by adding a capture group to the hash regex. Once this is done, the code would be something like

```cpp
std::tuple<Error, uint16_t> ApiDispatcher::getTransactionDetails(
    const httplib::Request &req,
    httplib::Response &res,
    const nlohmann::json &body) const
{
    std::string hashStr = req.matches[0];
```

It is not much of a gain in readability here, but there are some methods which have multiple parameters needing to be extracted which would benefit.

brandonlehmann:
I think I just made this worse :)

brandonlehmann:
Handled in #1092

PR #1092, opened 2020-08-03
added matches property in ApiDispatcher regex

mrrovot:


mrrovot:
This solves issue #862

brandonlehmann:
Wanna try taking a stab at the same in the daemon?

mrrovot:
> Wanna try taking a stab at the same in the daemon?

Will try

Issue #793, opened 2019-04-26
Enhancement: Scan height range


derp-derp-derp:
It would be cool if zedwallet allowed you to enter not only the starting block to scan for transactions, but also the ending block... like a range.

_e.g._ I have a wallet, and I know I only used it from block x to block x, but the import still takes a long time per having to scan all of the blocks after the starting point, as opposed to stopping at a certain block.

zpalmtree:
Should it scan this range, and then resume scanning from the 'current' height, or just stop there and never scan more blocks?

derp-derp-derp:
Great question. Resuming scanning from the current height makes sense I think.

RocksteadyTC:
+1 resume

mrrovot:
In the future couldn't we add to this feature by:

1 - take a synced wallet
2 - export to JSON file all block heights where transactions happened
3 - have the `scanRange` function with an option to import a JSON file
4 - if you need to import the wallet from scratch, just have import it together with the JSON file and the new sync will be really quick

just an idea

brandonlehmann:
> In the future couldn't we add to this feature by:
> 
> 1 - take a synced wallet
> 2 - export to JSON file all block heights where transactions happened
> 3 - have the `scanRange` function with an option to import a JSON file
> 4 - if you need to import the wallet from scratch, just import it together with the JSON file and the new sync will be really quick
> 
> just an idea

If you only ask a node, of which a lot of uses use remote ones, for *interesting* blocks you’re telling the node operator exactly which funds are yours. ;)

zpalmtree:
Implemented in #1093

PR #1093, opened 2020-08-08
Adding Height range

mrrovot:
This tries to solve issue #793


src/walletbackend/WalletSynchronizer.h
```diff
@@ -52,6 +52,9 @@ class WalletSynchronizer
     /* Constructors */
     //////////////////
 
+
+    std::optional<uint64_t> endScanHeight;
```
zpalmtree, line 56: This should be a private member variable. We also prefix member variables with `m_` as convention, so please move it to the appropriate section below and rename it `m_endScanHeight`

src/walletbackend/WalletSynchronizer.cpp
```diff
@@ -663,7 +683,7 @@ void WalletSynchronizer::checkLockedTransactions()
 /* Launch the worker thread in the background. It's safest to do this in a
    seperate function, so everything in the constructor gets initialized,
    and if we do any inheritance, things don't go awry. */
-void WalletSynchronizer::start()
+void WalletSynchronizer::start(const bool stopSyncThread)
```
zpalmtree, line 686: this should be named `startSyncThread` in this function.

src/walletbackend/WalletSynchronizer.cpp
```diff
@@ -679,7 +699,9 @@ void WalletSynchronizer::start()
     m_blockProcessingQueue.start();
     m_processedBlocks.start();
 
-    m_syncThread = std::thread(&WalletSynchronizer::mainLoop, this);
+    if (!stopSyncThread) {
```
zpalmtree, line 702: Please use allman style braces, that is, with a new brace on a new line.

src/walletbackend/WalletSynchronizer.cpp
```diff
@@ -704,22 +726,27 @@ void WalletSynchronizer::stop()
     m_haveBlocksToProcess.notify_all();
     m_haveProcessedBlocksToHandle.notify_all();
 
-    m_blockProcessingQueue.clear();
-    m_processedBlocks.clear();
 
-    /* Wait for the block downloader thread to finish (if applicable) */
-    if (m_syncThread.joinable())
+    if (!stopSyncThread)
```
zpalmtree, line 730: This logic is inverted. We should be stopping the sync thread when `stopSyncThread` is `true`. Make sure `stopSyncThread` and `startSyncThread` for the `stop/start` functions default to true, to not alter the existing behaviour. Your existing code seems to be flipped around both with the default values and how you read from them, so it works as expected, but doesn't read correctly.

src/walletbackend/WalletSynchronizer.cpp
```diff
@@ -704,22 +726,27 @@ void WalletSynchronizer::stop()
...
     {
-        m_syncThread.join();
+        m_blockProcessingQueue.clear();
+        m_processedBlocks.clear();
```
zpalmtree, line 733: I don't think we should be conditionally clearing these data structures. I don't believe we'll want this to be populated when the sync thread resumes at the new height.

src/zedwallet++/Utilities.cpp
```diff
@@ -128,4 +128,84 @@ namespace ZedUtilities
...
+            } else {
+
+              endHeight = std::stoull(stringEndHeight);
+            }
+
+            try
+            {   
+                return { startHeight, endHeight };
+            }
+            catch (const std::out_of_range &)
```
zpalmtree, line 198: This try catch isn't in the right place. We need the `try` for the two `std::stoull`'s - if the user inputs a non number or an out of range argument, it will throw. The braces and spacing also needs a bit of love in this function.

src/walletbackend/WalletSynchronizer.cpp
```diff
@@ -165,19 +163,19 @@ void WalletSynchronizer::mainLoop()
 
 
         /* if endScanHeight is set, stop syncing at endScanHeight and start syncing from top of the chain */
-        if (endScanHeight != std::nullopt && getCurrentScanHeight() >= endScanHeight)
+        if (m_endScanHeight != std::nullopt && getCurrentScanHeight() >= m_endScanHeight)
```
zpalmtree, line 166: Should be able to simplify this to just

```suggestion
        if (m_endScanHeight && getCurrentScanHeight() >= m_endScanHeight)
```

src/zedwallet++/Utilities.cpp
```diff
@@ -4,208 +4,202 @@
...
-            catch (const std::invalid_argument &)
-            {
-                std::cout << WarningMsg("Failed to parse height - input is not ") << WarningMsg("a number!")
-                          << std::endl
-                          << std::endl;
-            }
-        }
+  }
+
+  uint64_t getScanHeight()
```
zpalmtree, line 42: Looks like your editor might have auto formatted this file? If you are not sure how formatting should look, there is a clang-format script here https://github.com/turtlecoin/turtlecoin/blob/development/scripts/format.js, with the associated rules here https://github.com/turtlecoin/turtlecoin/blob/development/.clang-format

src/zedwallet++/Utilities.cpp
```diff
@@ -4,208 +4,202 @@
...
+
+      /* Remove commas so user can enter height as e.g. 200,000 */
+      Utilities::removeCharFromString(stringStartHeight, ',');
+
+      if (stringStartHeight == "") 
+      {
+        stringStartHeight = "0";
+      }
+
+      std::string defaultEndHeight = std::to_string(std::stoull(stringStartHeight) + 1000);
```
zpalmtree: There's a `std::stoull` here which isn't wrapped in a try/catch. In addition, we've got one while loop for multiple inputs. If the first is bad, we'll still prompt them for a second, then fail on the second. It's much preferable for us to fail instantly on a bad input, so we should have a while loop per input. Make sure that entering a non int causes the input to be rejected and to reprompt for a valid one.

If we parse them one by one it also means you can use the parsed value here instead of having to parse the input with `std:stoull` twice.

src/zedwallet++/Utilities.cpp
mrrovot: Is there a way to undo what this script did? now it won't compile, it gives all type of errors like ```src/common/StringTools.cpp:84:45: error: expected expression
        for (uint64_t i = 0; i<text.size()> > 1; ++i)```

zpalmtree: just reset git staged changes `git checkout -f`

zpalmtree: And it seems to be because that file has fucked up formatting already and it breaks clang format, not sure why.

```
-        for (uint64_t i = 0; i<text.size()>> 1; ++i)
+        for (uint64_t i = 0; i<text.size()> > 1; ++i)
```

should be `for (uint64_t i = 0; i < text.size() >> 1; ++i)`

zpalmtree: You can of course use clang-format manually for a single file if desired to just fix the file you're working on. Code compiles after fixing the three instances in the bugged file.

mrrovot: When you run the script it changes many files in many directories inside /src so I believe many files have different format then the script.

Can I add a `--file path` args to the srcript so it can also format single files upon requests?

mrrovot: Just saw your above comment, will use clang-format manually in the future

zpalmtree: If you like. The script has been ran previously on the whole codebase, but it's not automatically ran on every PR, so naturally it gets out of date.

src/zedwallet++/Utilities.cpp
```diff
@@ -128,4 +124,99 @@ namespace ZedUtilities
...
+            {
+                stringEndHeight = std::to_string(startHeight + 1000);
+            }
+
+            try
+            {
+                return
+                {
+                   startHeight, 
+                   endHeight = std::stoull(stringEndHeight)
```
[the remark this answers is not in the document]
mrrovot: If I create `getHeight()` with a configurable message it would mean that `commandDispatcher.cpp` would become the place with the string to pass to` getHeight()`, is this the desirable outcome? 

Right now we have 3 cases where we would need `getHeight()`:  rewind, scanRange and reset, I could also create the 3 strings in the `getHeight()` and choose the correct one to show based on a value passed to the function and not bloat commandDispatcher with strings. 
Please let me know the preferred way

src/zedwallet++/CommandImplementations.cpp
```diff
@@ -326,6 +335,53 @@ void rewind(const std::shared_ptr<WalletBackend> walletBackend)
...
 
+void scanRange(const std::shared_ptr<WalletBackend> walletBackend)
+{
+
+    const std::string startHeightMsg = "What height would you like to begin scanning your wallet from?\n\nThis can greatly speed up the initial wallet scanning process.\n\nIf you do not know the exact height, err on the side of caution so transactions do not get missed.\n\nHit enter for the sub-optimal default of zero:" ;
+    const uint64_t startHeight = getHeight(startHeightMsg);
+
+
+    std::string defaultEndHeight;
+    defaultEndHeight = std::to_string(startHeight + 1000);
```
zpalmtree, line 346: No need to split up the declaration and assignment here. 
```
const std::string defaultEndHeight = std::to_string(startHeight + 1000);
```

src/zedwallet++/CommandImplementations.cpp
```diff
@@ -265,8 +265,11 @@ void status(const std::shared_ptr<WalletBackend> walletBackend)
 }
 
 void reset(const std::shared_ptr<WalletBackend> walletBackend)
-{
-    const uint64_t scanHeight = ZedUtilities::getScanHeight();
+{   
+
+    const std::string msg = "What height would you like to begin scanning your wallet from?\n\nThis can greatly speed up the initial wallet scanning process.\n\nIf you do not know the exact height, err on the side of caution so transactions do not get missed.\n\nHit enter for the sub-optimal default of zero: ";
```
zpalmtree, line 270: Can we split this message up so it's a reasonable line length? You can either concat strings with `+` or the compiler will automatically concat strings together, for example:

```
const std::string msg = "What height would you like to begin scanning your wallet from?\n\n"
                        "This can greatly speed up the initial wallet scanning process.\n\n"
```

Please apply the same treatment to the other long strings in this file. I try and not go over 120 chars per line.

mrrovot: What do you think about the `const std::string getHeightMsg` I added in `Open.h`? the same string was repetitive so I put it up in a variable but wasn't sure if to add it to the header file or to the top of the cpp file.

zpalmtree: It's not a bad idea but I'm not a huge fan of it being in either the header file *or* the cpp.. It doesn't really belong in CryptoNoteConfig or WalletConfig. Maybe could add an overloaded `getHeight()` which doesn't take any args, and passes the default message on to the existing `getHeight()`?

mrrovot: Added the overloaded function, looks/feels way better.

src/walletapi/ParseArguments.cpp
```diff
@@ -82,7 +83,8 @@ ApiConfig parseArguments(int argc, char **argv)
          "Specify the <password> to access the RPC server.",
          cxxopts::value<std::string>(config.rpcPassword),
          "<password>");
-
+    // clang-format off
```
zpalmtree, line 86: This should be
```suggestion
    // clang-format on
```

src/zedwallet++/CommandImplementations.cpp
```diff
@@ -326,6 +342,57 @@ void rewind(const std::shared_ptr<WalletBackend> walletBackend)
     walletBackend->m_eventHandler->onTransaction.resume();
 }
 
+void scanRange(const std::shared_ptr<WalletBackend> walletBackend)
+{
+    const std::string startHeightMsg = "What height would you like to begin scanning your wallet from?\n\n"
+                                       "This can greatly speed up the initial wallet scanning process.\n\n"
+                                       "If you do not know the exact height, err on the side of caution so "
+                                       "transactions do not get missed.\n\n"
+                                       "Hit enter for the sub-optimal default of zero:";
```
zpalmtree, line 351:
```suggestion
                                       "Hit enter for the sub-optimal default of zero: ";
```

src/zedwallet++/CommandImplementations.cpp
```diff
@@ -326,6 +342,57 @@ void rewind(const std::shared_ptr<WalletBackend> walletBackend)
...
+
+    const uint64_t startHeight = getHeight(startHeightMsg);
+
+    std::string defaultEndHeight = std::to_string(startHeight + 1000);
+
+    const std::string endHeightMsg = "What height would you like to end scanning your wallet from?\n\nHit "
+                                     "enter for the default of "
+                                     + defaultEndHeight + ": ";
+
+    uint64_t endHeight = getHeight(endHeightMsg);
```
zpalmtree, line 361: The default option isn't working here - it gives `0` rather than the expected value.

image

zpalmtree: Also, while we're here, `What height would you like to stop scanning your wallet at?` reads a little better

zpalmtree (approved):

zpalmtree:
@mrrovot Need to fix merge conflicts before merging

mrrovot:
> @mrrovot Need to fix merge conflicts before merging

done

PR #1113, opened 2021-03-11
Bubble Up Transaction Validation Details

brandonlehmann:
If a transaction fails validation, bubble the actual reason supplied in the validation routine up to the top so that we can do better troubleshooting later. The message, in some cases, then includes additional information that  may be helpful for further troubleshooting.


brandonlehmann:
Example: POST /transaction

image

zpalmtree (approved):

PR #1117, opened 2021-04-05
Update version of download link

cmxhost:


brandonlehmann:
This requires additional checks as from what I can recall versions >= 1.70 of boost introduced other issues; however, that was related to 1.72 iirc.

cmxhost:
Got it

PR #1126, opened 2021-06-17
Update CuveeCZ seed node IP address

LeoStehlik:
We've moved CZ server room location to a new place and switched ISP. Hence a new IP address.


RocksteadyTC:
congrats on the new location!

brandonlehmann (approved):

